PR #13, opened 2020-04-16
fix(rxjs-state): 🐞 build library in Angular Package Format for more generic support (fixes #3)

### LINKED ISSUE #3: Error: Class constructor RxState cannot be invoked without 'new' in stackblitz
For some reason it doesn't work on stackblitz.

https://stackblitz.com/edit/ngx-rx-state-demo?file=package.json

yjaaidi:
This will build `rxjs-state` with **ng-packagr** too _(just like `ngx-rx-state`)_ and produce a library in Angular Package Format. This should fix issues with Jest & Stackblitz.

Other improvement ideas:

1. We could harmonize the monorepo by using `ng-packagr` for all packages.
2. Package names should either start with a scope or not be prefixed with `@`. This means `@rxjs-etc` should be renamed to `rxjs-etc` or `@ngx-rx/rxjs-etc` with a preference for the latest.

### DIFF
--- a/angular.json
+++ b/angular.json
@@ -105,6 +105,18 @@
       "projectType": "library",
       "schematics": {},
       "architect": {
+        "build": {
+          "builder": "@nrwl/angular:package",
+          "options": {
+            "tsConfig": "libs/rxjs-state/tsconfig.lib.json",
+            "project": "libs/rxjs-state/ng-package.json"
+          },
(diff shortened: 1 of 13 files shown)


BioPhoton:
Introduced `@ngx-rx` scope everywhere.

Thanks!

PR #21, opened 2020-04-19
Update README.md

edbzn:
Fix `rxjs-state` library link.


edbzn:
Close in favor of #22

PR #23, opened 2020-04-20
feature(strategies) added a new very aggressive (brutal) strategy

hoebbelsB:
implemented a strategy called 'brutal' which `detaches` the component entirely from the `changeDetection`. The component will get `reattached` before each rendering cycle and after that being `detached` again.


BioPhoton:
Closed as already mergred

PR #36, opened 2020-04-25
chore: remove ngcc postinstall hook

LayZeeDK:
It's doesn't seem to be needed anymore. This has changed a lot in the Angular 9.x releases.

If it's needed again, it should be something like

```
"postinstall": "ngcc && ngcc --properties main"
```

The last part is for producing UMD bundles for Jest.


BioPhoton:
I see. Thanks!!

PR #37, opened 2020-04-25
refactor(template): remove unnecessary metadata from template module

LayZeeDK:
There's no need to add Angular modules to `imports` when we don't have `declarations`.

After this refactor, I don't see the need for the `MODULES` constant.


BioPhoton:
Thanks a lot @LayZeeDK

PR #38, opened 2020-04-25
test(template): add template module test suite

LayZeeDK:
Add test suite for the template module, verifying the declarables that it exposes.


BioPhoton:
This is a nice PR Thanks!

PR #40, opened 2020-04-25
feat: add back tslib dependency

LayZeeDK:
This is technically a breaking change, but it's default for Angular package libraries, so it shouldn't be a big deal.

Note that we currently use `tslib` v1.11.x in this workspace, but Angular CLI 9.0.x apps come with v1.10.x out-of-the-box.


BioPhoton:
Thanks @LayZeeDK

PR #41, opened 2020-04-25
feat: support Angular 8.0.0

LayZeeDK:
Greater than OR equal to 8.0.0.


BioPhoton:
Will consider it. For now, I close it.

Thanks a lot for the suggestion!

PR #42, opened 2020-04-25
chore: downgrade to minimum compatible package dependencies

LayZeeDK:
Downgrade to the minimum versions supported by an Angular CLI 9.0.x app.

NOTE: Consider dropping Angular 8.x support since we don't test that version.


BioPhoton:
Will consider it. For now, I close it.

Thanks a lot for the suggestion!

PR #43, opened 2020-04-25
chore: use Yarn

LayZeeDK:
Yarn is generally faster than NPM and resolves dependencies
deterministically.


BioPhoton:
ok. Will use it :)

PR #59, opened 2020-04-28
Removed duplicated angular.json config

kajetansw:
I noticed that in the `angular.json` there are duplicated configurations for three of the projects. This PR serves to remove them.

(Personally, I just wanted to say that I'm just beginning my adventure as an OSS contributor and I found this library very fascinating! 😊 I hope to learn and help if there will be a chance 🙂)


BioPhoton:
Thanks a lot @kajetansw ! 
<3

PR #83, opened 2020-05-06
feat(state) add signal$ to accumulationObservable

### LINKED ISSUE #60: @rx-angular/state reactive @Output binding needs improvement in usage
kind of refers to: #57 and #27. as discussed [here](https://github.com/BioPhoton/rx-angular/pull/57/files#r416960128) we need a seperate solution for handling `@Output()` bindings. The `select()` operator/method returns a shared and replayed observable which is not how angular typically handles Output bindings (`EventEmitter<T>` aka `Subject<T>`). Thus we have to come up with a solution for this.

We could combine the `$` property with a custom operator which has the same API as `select` but returns a cold observable. 
Another way to handle Output bindings would be to use `hold` and trigger a subject based on a selection. Exactly how stated in #27 .

BioPhoton:
closes #60


libs/state/spec/component-patterns/rx-state.component.spec.ts
```diff
@@ -0,0 +1,166 @@
...
+  });
+
+  it('should work wrong output initialisation', () => {
+    expect(parent.afterViewInit).toBeTruthy();
+    expect(parent.child.afterViewInit).toBeTruthy();
+
+    expect(parent.get().str).toBe(undefined);
+    expect(parent.get().strWrong).toBe(initialChildState.str);
+    expect(parent.child.get().str).toBe(initialChildState.str);
+  });
```
hoebbelsB, line 165: thought about this again. shareReplay in select() operator will still cause issues since the emitted value will we replayed after first emission. so every late subscriber gets a replayed value, which is invalid for output bindings. I guess we have to come up with a `selectCold` operator which does the same thing as select, but not replaying. We could wrap the cold one into the normal one in order to handle code duplication

BioPhoton: The subscriber is the output binding. 

I'm sure even if we would have multiple output bindings with the same selection we would not have a late subscriber.

But we can test it.

Would you sketch a test here roughly so I see the case to test?

BioPhoton: > `selectCold` :
I can't image what this should be... Can you elaborate?

hoebbelsB: will create a test for that, yes

BioPhoton: IS this related to this PR?
Otherwise lets finishe this one up.

hoebbelsB: finish it!

PR #85, opened 2020-05-08
fix(template) fixed a bug in the promise-duration-selector 

hoebbelsB:
…which caused crashes when zone was not included.

There were missing brackets.

### DIFF
--- a/libs/template/src/lib/core/render-aware/promise-duration-selector.ts
+++ b/libs/template/src/lib/core/render-aware/promise-duration-selector.ts
@@ -8,9 +8,10 @@ let resolvedPromise: Promise<void> | null = null;
 
 function getResolvedPromise(): Promise<void> {
   resolvedPromise =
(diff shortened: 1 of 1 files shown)


libs/template/src/lib/core/render-aware/promise-duration-selector.ts
```diff
@@ -8,9 +8,10 @@ let resolvedPromise: Promise<void> | null = null;
 
 function getResolvedPromise(): Promise<void> {
   resolvedPromise =
-    resolvedPromise || apiZonePatched('Promise')
+    resolvedPromise ||
+    (apiZonePatched('Promise')
       ? (getGlobalThis().__zone_symbol__Promise.resolve() as Promise<void>)
-      : Promise.resolve();
+      : Promise.resolve());
```
BioPhoton, line 14: We have to implement a test for that.

PR #87, opened 2020-05-08
Tour of heroes

hoebbelsB:
copied the original [angular tour of heroes](https://angular.io/tutorial) tutorial and used `@rx-angular/state` aswell as `@rx-angular/template` to manage state inside all components.

As a result I was able to completely turn of `NgZone`, which reduced the bundlesize (polyfills.ts) from `142kb` to `3.45kb`.

I have marked this branch as blocked, because it utilizes the strategy input of the `rxLet` Directive, which is currently broken. I've created another PR #86 where this is fixed.

I have probably found another bug concerning the strategies. Take a look at the `heroes.component.ts` aswell as `heroes.component.html`. I wanted to link the `configService.renderStrategy$` to the config parameter of the push pipe, which didn't work at all. That's why i hardcoded it in every template. @BioPhoton maybe you have some idea?


hoebbelsB:
#85 #88 have to be merged in order to get this one running

BioPhoton:
This is a really great idea!!

LOVE IT! Thank you soooo mush!

PR #95, opened 2020-05-09
Small typo fix, new links to select & connect.

Karnaukhov-kh:


libs/state/docs/snippets/behavior-subject-vs-rx-state.md
```diff
@@ -387,7 +387,7 @@ RxState service is in component providers. That means that lifecycle of this ser
   tasks$ = this.state.select("tasks");
 ```
 
-Visually it looks the same but select operator allows lot more possibilities than just passing keys. You can read about it [here](https://github.com/BioPhoton/rx-angular/blob/master/libs/state/docs/api.md#select). Also selection will be cached, distinct and undefined values will be filtered out.
+Visually it looks the same but select operator allows lot more possibilities than just passing keys. You can read about it [here](https://github.com/BioPhoton/rx-angular/blob/master/libs/state/docs/api/rx-state.md#select). Also selection will be cached, distinct and undefined values will be filtered out.
```
hoebbelsB, line 390:
```suggestion
Visually it looks the same but the select operator provides a lot more than just passing keys. You can read about it [here](https://github.com/BioPhoton/rx-angular/blob/master/libs/state/docs/api/rx-state.md#select). Also selection will be shareReplayed, distincted and undefined values will be filtered out.
```

libs/state/docs/snippets/behavior-subject-vs-rx-state.md
```diff
@@ -427,7 +427,7 @@ Now we need to update our `answerHandler$` so it will return an id of task that
   }
 ```
 
-Note that we removed `withLatestFrom(this.tasks$)` and using different approach with `connect`. First we tell to what field updates should be applied to, after that goes source of updates and the last one is `projectionFunction`. In this funciton first argument is our current state, second one is a value from our source. More on possible `connect` variants [here](https://github.com/BioPhoton/rx-angular/blob/master/libs/state/docs/api.md#connect).
+Note that we removed `withLatestFrom(this.tasks$)` and using different approach with `connect`. First we tell to what field updates should be applied to, after that goes source of updates and the last one is `projectionFunction`. In this funciton first argument is our current state, second one is a value from our source. More on possible `connect` variants [here](https://github.com/BioPhoton/rx-angular/blob/master/libs/state/docs/api/rx-state.md#connect).
```
hoebbelsB, line 430:
```suggestion
Note that we removed `withLatestFrom(this.tasks$)` in favor of the `projectionFunction` in `connect`.
First we define fields to be updated, then the source of the changes and lastly we provide the `projectionFunction`.
The functions' first first argument is the current state, the second is the change coming from our source. More on possible `connect` variants [here](https://github.com/BioPhoton/rx-angular/blob/master/libs/state/docs/api/rx-state.md#connect).
```

PR #96, opened 2020-05-11
modify loadingSignal behavior

bernatsampera:
Not sure what was the desired behavior of the filter, but the list reload makes much more sense this way


BioPhoton:
I just saw it in the demo!

Thanks a lot!

PR #99, opened 2020-05-13
Model & ViewModel with RxState snippet suggestion.

Karnaukhov-kh:


libs/state/docs/snippets/model-and-viewModel-with-rx-tate.md
```diff
@@ -0,0 +1,259 @@
+In this example we will see what is the deifference between Model and ViewModel and show different approaches how Model & ViewModel can be handled using RxState. First we will go with imperative approach and after that make it fully reactive.
+
+# What is what?
+### What is Model?
+Model is an object or class that holds our data.
```
BioPhoton, line 5:
```suggestion
A model is an object or class that holds our data. In our case, it is a property in our component called model
```

libs/state/docs/snippets/model-and-viewModel-with-rx-tate.md
```diff
@@ -0,0 +1,259 @@
...
+interface IViewList {
+  name: string;
+  items: string[];
+  itemsTotal: number;
+  isOpen: boolean;
+}
+```
+As you can see they are a bit different. Not all `IList` data needed in the View and at the same time `IViewList` contains some additional properties that are not in the `IList`.
+
+### ListService
```
BioPhoton, line 42: I would focus on the concept of model vs view model. A service here can maybe named by I see the code as overhead.
In general, we should shrink it down to its essence that is related to the model vs view model.

Karnaukhov-kh: Will remove service code and try to reduce this wall of text a bit. For now will convert this request to draft.

BioPhoton: But don't throw it away, it's good stuff there.

libs/state/docs/snippets/model-and-viewModel-with-rx-tate.md
```diff
@@ -0,0 +1,259 @@
...
+    this.toggle$.pipe(startWith(true))
+  ]).pipe(
+    map(([name, items, isOpen]) => ({
+      name,
+      items,
+      itemsTotal: items.length,
+      isOpen,
+      buttonLabel: isOpen ? "Close" : "Open"
+    })),
+    distinctUntilSomeChanged(['itemsTotal', 'isOpen])
```
hoebbelsB, line 134: maybe you take a look at the selectSlice operator? look in this PR: #119 there is a snippet about selectSlice and viewModel

Karnaukhov-kh: Yes, definitely will do.

PR #110, opened 2020-05-19
Global vs local state snippet

Karnaukhov-kh:
Basically a very-very tiny part of 'Layers of State' chapter + example of how to decide about state type.


BioPhoton:
Nice! 😍❤️

I'll do a review tomorrow!

PR #114, opened 2020-05-21
select slice operator

hoebbelsB:
implemented `selectSlice` operator. Scaffolded a test suite and written some docs aswell. Everything should get reviewed and improved :)

- [x] docs
- [x] tests
- [x] performance considerations (combineLatest vs. mergeAll)


libs/state/docs/api/operators/select-slice.md
```diff
@@ -0,0 +1,39 @@
...
+   title: string;
+   items: string[];
+   panelOpen: boolean;
+}
+const state$: Observable<MyState> = of({
+  title: 'myTitle',
+  items: ['foo', 'bar'],
+  panelOpen: true
+});
+const slice$ = state$.pipe(selectSlice('items', 'panelOpen')).pipe(tap(console.log)).subscribe();
```
Karnaukhov-kh, line 23: Just a thought:
It might be a bit confusing for some users.
- state.select('bar', 'foo'); will return nested property
- selectSlice('items', 'pannelOpen'); will return properties from the same level.
In addition distinctUntilSomeChnaged accepts strict array (without rest) of keys and is also mostly for managing viewModel.

Maybe just an array?

PR #120, opened 2020-05-22
feat(state): extend accumulator observable to take accumulation functions at runtime

BioPhoton:
# DRAFT

 extend accumulator observable to take accumulation funions at runtime

### DIFF
--- a/libs/state/src/lib/rx-state.service.ts
+++ b/libs/state/src/lib/rx-state.service.ts
@@ -15,7 +15,8 @@ import {
   pipeFromArray,
   stateful,
   WrongSelectParamsError,
-  isKeyOf
+  isKeyOf,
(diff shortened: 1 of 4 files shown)


libs/state/src/lib/rx-state.service.ts
```diff
@@ -78,6 +79,10 @@ export class RxState<T extends object> implements OnDestroy, Subscribable<T> {
     this.subscription.unsubscribe();
   }
 
+  setAccumulator(accumulatorFn: AccumulationFn) {
```
hoebbelsB, line 82: we need to test this method, otherwise it's fine i guess :)

BioPhoton: added tests

PR #126, opened 2020-05-29
fix: detect zone.js by using runOutsideAngular()

chrisguttandin:
This pull request aims to match the changes introduced here: ngrx/platform#2547.

It changes the way zone.js gets detected.


BioPhoton:
Thanks, A LOT for this fix!

PR #129, opened 2020-06-02
docs: Add ngrx selector example snippet

Phhansen:
Add a new example snippet using NgRx selectors with `@rx-angular/state`. Related to #100


BioPhoton:
Cool! Accepted and merged! <3

PR #161, opened 2020-06-26
Refactor get method

kubadospial:
references this issue: https://github.com/BioPhoton/rx-angular/issues/140


libs/state/docs/api/rx-state.md
```diff
@@ -61,6 +62,25 @@ if (!disabled) {
 }
 ```
 
+### get
+
+##### typeof: (k1: K1) => Partial<T>
+
+Read from the state in imperative manner by providing keys as parameters.
```
BioPhoton, line 69:
```suggestion
Read from the state in imperative manner by providing keys as parameters to reach deeply nested values.
```

libs/state/docs/api/rx-state.md
```diff
@@ -61,6 +62,25 @@ if (!disabled) {
...
+Returns the part of state object.
+
+_Example_
+
+```TypeScript
+// Access a single property
```
BioPhoton, line 75:
```suggestion
interface State {
  bar: { foo: `test`},
  baz: true
}

// Access a single property
```

libs/state/src/lib/core/utils/object-diver.ts
```diff
@@ -0,0 +1,10 @@
+import { isKeyOf } from './typing';
+
+export function objectDiver<T>(
+  obj: T | Partial<T>,
+  keys: string[]
+): T | Partial<T> {
+  const key = keys.shift();
+  const value = isKeyOf(key) ? obj[key] : null;
+  return !!value ? (keys.length > 0 ? objectDiver(value, keys) : value) : obj;
+}
```
BioPhoton, line 10: Here I suggest working immutable with the keys array as we would mutate a user variable.
Furthermore, TS should work correctly. Have a look at the pluck operator or similar to get an idea of the typing.
maybe consider a more intuitive naming as e.g. `safePluck`.

```suggestion
export function objectDiver<T>(
  obj: Partial<T>,
  keys: ????[]
): ??? {
  let prop = obj;
  keys.forEach((key) => {
    if(prop[key] !== undefined) {
      prop = prop[key];
    }
  })
  return prop;
}
```

libs/state/src/lib/rx-state.service.ts
```diff
@@ -94,8 +95,69 @@ export class RxState<T extends object> implements OnDestroy, Subscribable<T> {
...
+   *
+   * @return T | Partial<T>
+   */
+
+  get<K1 extends keyof T>(k1?: K1): T | Partial<T>;
+
+  get<K1 extends keyof T, K2 extends keyof T[K1]>(
+    k1?: K1,
+    k2?: K2
+  ): T | Partial<T>;
```
BioPhoton, line 123: isnt it this?
```suggestion
  ): T | T[K1][K2];
```

JakubDospial: https://github.com/BioPhoton/rx-angular/pull/161/commits/715df6a0f63a7e724a6ebe5bda19a8b3f6c5c1cf

libs/state/src/lib/core/utils/object-diver.ts
JakubDospial: https://github.com/BioPhoton/rx-angular/pull/161/commits/cf04a730b730345277c473f302a516bf5697f9b5
https://github.com/BioPhoton/rx-angular/pull/161/commits/c8e03aea2a094c6eaa642aff5e010dad51203e99

libs/state/spec/core/utils/safe-pluck.spec.ts
```diff
@@ -0,0 +1,18 @@
+import { safePluck } from '../../../src/lib/core/utils';
+
+describe('safePluck', () => {
+    const bar = 'hello'
+    const obj = { foo: { bar } };
+
```
BioPhoton, line 6:
```suggestion
    it('should not touch passed arguments', () => {
        const args = ['foo', 'bar'];
        expect(safePluck(args.length)).toBe(2);
    })
```

PR #169, opened 2020-07-05
Transformation helpers.

Karnaukhov-kh:
For real.


Karnaukhov-kh:
State tests are working now. Issue was because our docker has only 2 cores and 4 GB of ram.  Problem fixed with --runInBand param that runs all tests sequentially.

https://jestjs.io/docs/en/troubleshooting#tests-are-extremely-slow-on-docker-andor-continuous-integration-ci-server

Template tests are failing (but template related files are not modified in this PR)

PR #173, opened 2020-07-06
Global render queue

BioPhoton:


libs/state/src/lib/rx-state.service.ts
```diff
@@ -3,6 +3,7 @@ import {
   isObservable,
   Observable,
   OperatorFunction,
+  Subject,
```
BioPhoton, line 6:
```suggestion
```

BioPhoton:
closed in favor of #260

PR #175, opened 2020-07-08
Log errors in accumulator

Karnaukhov-kh:
### Issue

When an error happens inside the connect method it is not appearing in the console. Only can be reproduced if we use select() to get data from the source. If we use $ everything is fine.

### Guess

Sometime after 1.0 we introduced publish() + publishReplay(1) in state accumulator.
```
 const signal$ = merge(
    stateObservables.pipe(
      distinctUntilChanged(),
      mergeAll(),
      observeOn(queueScheduler)
    ),
    stateSlices.pipe(observeOn(queueScheduler))
  ).pipe(
    scan(stateAccumulator, {} as T),
    tap(newState => (compositionObservable.state = newState)),
    publish()
  );
  const state$ = signal$.pipe(publishReplay(1));
```
And after that errors disappeared. If we keep `state$ = signal$` without publishReplay errors appear as they should do.

### Fix

Not the perfect solution but gets the job done. We can adjust tap a tiny bit.
```
tap(
    newState => (compositionObservable.state = newState),
    error => console.error(error)
),
```

### DIFF
--- a/libs/state/src/lib/core/accumulation-observable.ts
+++ b/libs/state/src/lib/core/accumulation-observable.ts
@@ -6,7 +6,8 @@ import {
   queueScheduler,
   Subject,
   Subscribable,
-  Subscription
+  Subscription,
+  EMPTY
 } from 'rxjs';
 import {
   distinctUntilChanged,
@@ -16,7 +17,8 @@ import {
   publishReplay,
   scan,
   tap,
-  withLatestFrom
+  withLatestFrom,
+  catchError
 } from 'rxjs/operators';
 
 export type AccumulationFn = <T>(st: T, sl: Partial<T>) => T;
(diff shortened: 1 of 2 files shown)


libs/state/src/lib/core/accumulation-observable.ts
```diff
@@ -52,7 +52,10 @@ export function createAccumulationObservable<T extends object>(
       (state, [slice, stateAccumulator]) => stateAccumulator(state, slice),
       {} as T
     ),
-    tap(newState => (compositionObservable.state = newState)),
+    tap(
+      newState => (compositionObservable.state = newState),
+      error => console.error(error)
+    ),
```
BioPhoton, line 58:
```suggestion
    ),
    catchError(e => EMPTY)
```

Let's catch it here to forcefully not get lost...

libs/state/src/lib/core/accumulation-observable.ts
```diff
@@ -52,7 +54,11 @@ export function createAccumulationObservable<T extends object>(
...
+    catchError(e => EMPTY),
```
BioPhoton, line 61:
```suggestion
    // @Notice We catch the error here as it get lost in between `publish` and `publishReplay`. We return empty to
    catchError(e => EMPTY),
```

PR #179, opened 2020-07-12
fix(state): fixed selectSlice typing

### LINKED ISSUE #177: selectSlice still a problem?
Upgraded to 1.3 with the hope of being able to use `selectSlice` rather than a combination of `distinctUntilSomeChanged` and `map`, and it doesn't seem to be giving the result I expected.

Here's the existing code I was hoping to be able to refactor:

image

Here's the refactored code:

image

The typescript compiler isn't happy with this as can be seen:

image

I thought I should be able to use selectSlice as above. Am I missing something?

hoebbelsB:
- removed `| null` from selectSlice's typing. this breaks apps which run in strictMode.
- added usage to `selectSlice.spec.ts` to verify typing in strictMode
- added usage in `experiments` app to verify typing in non-strictMode

closes #177

### DIFF
--- a/libs/state/spec/rxjs/operators/selectSlice.spec.ts
+++ b/libs/state/spec/rxjs/operators/selectSlice.spec.ts
@@ -61,8 +61,7 @@ describe('selectSlice operator', () => {
       expectObservable(
         e1.pipe(
           selectSlice(['val']),
-          map(({ val }) => ({ val })) // this is here to test if the typings work in strict mode. we had some
-          // struggles ;)
+          map(({ val }) => ({ val })) // this is here to test if the typings work in strict mode
         )
       ).toBe(expected, values);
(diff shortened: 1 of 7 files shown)


libs/state/spec/rxjs/operators/selectSlice.spec.ts
```diff
@@ -60,7 +60,9 @@ describe('selectSlice operator', () => {
 
       expectObservable(
         e1.pipe(
-          selectSlice(['val'])
+          selectSlice(['val']),
+          map(({ val }) => ({ val })) // this is here to test if the typings work in strict mode. we had some
+          // struggles ;)
```
BioPhoton, line 65:
```suggestion
          map(({ val }) => ({ val })) // this is here to test if the typings work in strict mode
```

Issue #113, opened 2020-05-21
Create Global State example


BioPhoton:
# Injecton Token:

```typescript
@Injectable({ providedIn: 'root' })
export class MessageService {
  readonly messages$ = this.state.select('messages');

  constructor(private state:RxState<MessageServiceState>()) {
    this.state.set(initState);
  }

  add(message: string) {
    this.state.set('messages', oldState => [...oldState.messages, message]);
  }
}
```
Mention boildr plate best one

# Compistion and manual implementation:

```typescript
@Injectable({ providedIn: 'root' })
export class MessageService implement OnDetroy {
 private readonly state = new RxState<MessageServiceState>();
  readonly messages$ = this.select('messages');

  constructor() {
  
    this.state.set(initState);
  }

  add(message: string) {
    this.set('messages', oldState => [...oldState.messages, message]);
  }

ngOnDetroy () {
this.state$.ngOnDetroy (); 
}

}
```

Mention we don't expose the full API => avoid misusage to avoid

# Extending the global service:

```typescript
@Injectable({ providedIn: 'root' })
export class MessageService extends RxState<MessageServiceState> {
  readonly messages$ = this.select('messages');

  constructor() {
    super();
    this.set(initState);
  }

  add(message: string) {
    this.set('messages', oldState => [...oldState.messages, message]);
  }

  clear() {
    this.set({ messages: [] });
  }
}
```

Mention downsides expose private API

Karnaukhov-kh:
Test

BioPhoton:
related to #111

BioPhoton:
https://github.com/cartant/ts-action
We could use it in our examples

tayambamwanza:
@BioPhoton I'm trying to migrate from @codewithdan/observable-store to @rx-angular/state, it looks pretty similar here, but I managed to only find this issue, where can I find this on offical docs site or is it yet to be updated?

This is the one I find on the docs site: [Using RxState as Global State](https://github.com/rx-angular/rx-angular/blob/master/libs/state/docs/snippets/manage-global-state.md)

BioPhoton:
We decided to skip tsaction example for now.
if you like you can create a demo. Also and especially for Dan's store. 😁👍

tayambamwanza:
@BioPhoton thanks, I managed to succesfully re-implement on my side, it's actually a lot less code thats needed for rx-angular/state anyway, so I'll contribute a demo at some point.

Issue #111, opened 2020-05-19
Using rx-state as global state (best practice)


andreElrico:
What is the best practice to use rxState as global state?

I have this setup currently.

+ create injection token 

`export const GLOBAL_RX_STATE = new InjectionToken('GLOBAL_RX_STATE');`

--------------------------------------------------
`in app.module.ts`

create instance with initial state

```
const _RX_STATE_INSTANCE =
(function(initialValue) {
    const ans = new RxState();
    ans.set(initialValue);

    return ans;
})(INITIAL_STATE);
```

+ add it to the providers in `app.module`'s decorator

```
    providers: [
        ...
        { provide: GLOBAL_RX_STATE, useValue: _RX_STATE_INSTANCE},
        ...
    ],
```

Karnaukhov-kh:
Hi.

Just a suggestion. You can have a service with ({providedIn: 'root'}) that extends RxState.

Example: https://github.com/BioPhoton/rx-angular/blob/7d9d6999f491915f9ab39f832c717ca70e0c2c47/libs/ngx-rx-state/src/lib/rx-global-state.service.ts

BioPhoton:
related to #113

nartc:
![image](https://user-images.githubusercontent.com/25516557/83892325-2169fc80-a714-11ea-8654-39d9c497dadc.png)
Have your `providedIn: 'root'` service extends RxState. Works super nicely.

BioPhoton:
closed in favor of #113

PR #180, opened 2020-07-13
chore(docs): Update snippet 'global-state-vs-local-state'

Phhansen:
Add code examples


Karnaukhov-kh:
Hi. 

Looks good :). There is a request for "RxState as a global state service" snippet (#113 & #111). "Moving the tasks array to our global state" part covers this case really well as for me. So maybe it can be a separate snippet?

Phhansen:
@Karnaukhov-kh that is a good idea! I'll look into it later tonight 👍

BioPhoton:
Hi @Phhansen 

What's the status on that? Is this finished and discussed stuff is somewhere else?

Just want to get an idea of how we deal with the feedback.

Phhansen:
Hi @BioPhoton, 

Sorry for the delay! I've been away on holiday with my family and just got back this weekend. I'll be at my computer tonight and able to finish this up.

There has been no discussion anywhere else apart from this PR.

Should I make another snippet and include a few ways of doing a global service? Using `injection token`, `service` with `providedIn: "root"` and such?

BioPhoton:
@Karnaukhov-kh and I discussed that and somewhere put our decision regarding whst the best way for a global service is. I'll try if I can point you there later.

Phhansen:
And i´m back from holidays! Any chance you´ve found your discussion, and what actions am I to take from now?

@BioPhoton

BioPhoton:
@Phhansen I guess we can also adopt the way how to use it globally. 
We will just adopt the PR. You can start.

Phhansen:
@BioPhoton I'm still a little unsure what you want me to start on. You mentioned some notes, but I haven't seen anything yet. I'll get right to it once I'm sure what it is :D

libs/state/docs/snippets/global-state-vs-local-state.md
```diff
@@ -31,18 +31,172 @@ Let's take a look at a simple todo app. This app has 2 views.
...
+```typescript
+import { GLOBAL_RX_STATE } from './rx-state';
+
+@Component({
+  selector: 'my-app',
+  templateUrl: './app.component.html',
+  styleUrls: [ './app.component.css' ]
+})
+export class AppComponent  {
+constructor(@Inject(GLOBAL_RX_STATE) private state, private tasksService: TasksService) {
```
BioPhoton, line 143: How are those 2 services related to ech other? Also the state is not used..

BioPhoton:
Hi @Phhansen some related issue is here #113 .

The token approach is ok. 

If you just could move the code in a separate file. So we have it separate form this topic. 

Thanks! :)

libs/state/docs/snippets/global-state-vs-local-state.md
Phhansen: Inside the `constructor` there is the `connect` to the tasksService.

```
    /* Fetch tasks from backend */
    this.connect("tasks", this.tasksService.fetchTasks());
```

I think my original thought was that if two components are using the same state, they shouldn´t both call the API, so the parent should. At least from my memory.

If anything needs changing I'll do just that.

BioPhoton:
Nice! Thanks for the contribution!! <3

PR #194, opened 2020-07-28
Added warning about unpatch & routing

Karnaukhov-kh:


libs/template/docs/unpatch.md
```diff
@@ -30,6 +31,10 @@ The `unpatch` directive solves this problem in a convenient way:
 </button>
 ```
 
+
+> **Warning**: Do not use `[unpatch]` on elements that should trigger navigation (with `routerLink` directly or with method bound to `(click)` or other events.
```
Karnaukhov-kh, line 35:
```suggestion
> **Warning**: Do not use `[unpatch]` on elements that should trigger navigation (with `routerLink` directly or with method bound to `(click)` or other events).
```

libs/template/docs/unpatch.md
```diff
@@ -4,6 +4,7 @@ The `unpatch` directive helps developers to partially deactivate `NgZone`, as we
 of unnecessary renderings through zones `addEventListener` patches.
 It can be used on any element you apply to event bindings.
 
+
```
hoebbelsB, line 7: what is this? :D

libs/template/docs/unpatch.md
```diff
@@ -30,6 +31,10 @@ The `unpatch` directive solves this problem in a convenient way:
 </button>
 ```
 
+
+> **Warning**: Do not use `[unpatch]` on elements that should trigger navigation (with `routerLink` directly or with method bound to `(click)` or other events).
+
+
```
hoebbelsB, line 37: thank you!

hoebbelsB, line 37:
```suggestion

> **Warning**: Do not use `[unpatch]` on elements that should trigger navigation (with `routerLink` directly or with method bound to `(click)` or other events). Otherwise you will end up having a 'Navigation triggered outside Angular zone, did you forget to call "ngZone.run()"?' warning.


```

hoebbelsB:
had a discussion with @BioPhoton about this. We could inject the `routerLink` Directive as `@Self() @Optional()` and do a check if it's there. If it's present, do not patch any events which are registered by `routerLink`. This would at least prevent the issues with the directive.
When using `unpatch` on a button which in turn runs `router.navigate` somewhere else, the warning will appear again. This is "by design"....

libs/template/docs/unpatch.md
BioPhoton, line 7:
```suggestion
```

BioPhoton:
created https://github.com/BioPhoton/rx-angular/issues/225
so we can mere :)

PR #197, opened 2020-07-31
Bugfix/fix return type for get method

kubadospial:
references this issue: https://github.com/BioPhoton/rx-angular/issues/193


libs/state/src/lib/rx-state.service.ts
```diff
@@ -115,33 +115,30 @@ export class RxState<T extends object> implements OnDestroy, Subscribable<T> {
    * @return T | Partial<T>
    */
 
-  get<K1 extends keyof T>(k1?: K1): T | Partial<T>;
+  get<K1 extends keyof T>(k1?: K1): T[K1];
```
BioPhoton, line 118: souldnt this be  T |  T[K1]? Also if we are abet to pass also no param we may move that in a custom overload?

kubadospial: If you take a look in the `safe-pluck.ts` you will see that if you pass 4 keys into the function it is possible to return the value for the 2nd key (if the 3rd key returns `undefined`). So logically yes! it should return `T | T[K1]` but if I keep `T |` TS will return error as below.

this is how I reproduce it:
**rx-state.service.ts:**
image
**component.ts:**
image
error message:
`Type 'SomeClass | { entity: SomeClass; }' is not assignable to type 'SomeClass'.
  Property 'test' is missing in type '{ entity: SomeClass; }' but required in type 'SomeClass'.`

The fix that I provided resolves this error.

christianacca: > Also if we are abet to pass also no param we may move that in a custom overload?

That was what I was thinking. So instead of just `get<K1 extends keyof T>(k1?: K1): T[K1];`, there is:
```ts
get(): T;
get<K1 extends keyof T>(k1: K1): T[K1];
```

BioPhoton:
I fount couple of more typing issues. 

See safe-pluck here: https://github.com/BioPhoton/rx-angular/blob/fix-return-type-for-get-method/libs/state/src/lib/core/utils/safe-pluck.ts
and the state (wip) here: https://github.com/BioPhoton/rx-angular/blob/fix-return-type-for-get-method/libs/state/src/lib/rx-state.service.ts

JakubDospial:
> I fount a couple of more typing issues.
> 
> See safe-pluck here: https://github.com/BioPhoton/rx-angular/blob/fix-return-type-for-get-method/libs/state/src/lib/core/utils/safe-pluck.ts
> and the state (wip) here: https://github.com/BioPhoton/rx-angular/blob/fix-return-type-for-get-method/libs/state/src/lib/rx-state.service.ts

if it's not relevant with this issue please open separate issue and I will handle it on another PR

BioPhoton:
I @JakubDospial as suggested please take the propper implementation from https://github.com/BioPhoton/rx-angular/tree/fix-return-type-for-get-method.

Also please reuse the safePluck in distinctUntilSomeChanged.ts

libs/state/src/lib/core/utils/safe-pluck.ts
```diff
@@ -1,34 +1,98 @@
...
+    | [K1, K2, K3, K4, K5]
+    | [K1, K2, K3, K4, K5, K6]
 ):
-  | T
   | T[K1]
   | T[K1][K2]
   | T[K1][K2][K3]
   | T[K1][K2][K3][K4]
   | T[K1][K2][K3][K4][K5]
   | T[K1][K2][K3][K4][K5][K6] {
```
hoebbelsB, line 81:
```suggestion
  | T[K1][K2][K3][K4][K5][K6] | null | undefined {
```

libs/state/src/lib/core/utils/safe-pluck.ts
```diff
@@ -1,34 +1,98 @@
...
-    if (isKeyOf(key) && !!prop[key]) {
+  if (!keys || keys.length <= 0) {
+    throw new Error('No keys given to pluck');
+  }
+  if (!stateObject) {
+    throw new Error('No object to pluck from');
+  }
+  if (Object.keys(stateObject).length <= 0) {
+    throw new Error(`Can't pluck from empty object`);
+  }
```
hoebbelsB, line 90:
```suggestion
  // needed to match null and undefined conventions of RxAngular core components
  if (stateObject == null) {
    return stateObject;
  }
  if (!keys || keys.length <= 0 || Object.keys(stateObject).length <= 0) {
    return undefined;
  }
```

libs/state/spec/core/utils/safe-pluck.spec.ts
```diff
@@ -1,22 +1,48 @@
...
+    expect(() => safePluck({}, args)).toThrowError(
+      "Can't pluck from empty object"
+    );
+  });
+
+  it('should throw error when keys are not provided', () => {
+    expect(() => safePluck(obj, null as any)).toThrowError(
+      'No keys given to pluck'
+    );
+  });
```
hoebbelsB, line 43:
```suggestion
  it('should return undefined when object is not valid', () => {
    expect(safePluck([undefined] as any, ['prop1'] as any)).toEqual(undefined);
  });

  it('should return undefined when keys are not provided', () => {
    expect(safePluck(obj, null as any)).toEqual(undefined);
  });

  it('should return undefined when incorrect keys are provided', () => {
    expect(safePluck(obj, ['incorrect-key'] as any)).toEqual(undefined);
  });
```

libs/state/spec/core/utils/safe-pluck.spec.ts
```diff
@@ -1,22 +1,34 @@
...
+  it('should not touch passed arguments', () => {
+    const args = ['foo', 'bar'] as any;
+    safePluck(obj, args);
+    expect(args.length).toBe(2);
+  });
 
-    it('should return whole `obj` when keys are not provided', () => {
-        expect(safePluck(obj, [])).toEqual(obj);
-    })
+  it('should return `bar` when keys are provided', () => {
```
hoebbelsB, line 15:
```suggestion
  it('should return value of last key', () => {
```

libs/state/spec/core/utils/safe-pluck.spec.ts
```diff
@@ -1,22 +1,34 @@
...
+    expect(safePluck(obj, null as any)).toEqual(undefined);
+  });
+
+  it('should return undefined when incorrect keys are provided', () => {
+    expect(safePluck(obj, ['incorrect-key'] as any)).toEqual(undefined);
+  });
+
+  it('should return undefined when incorrect keys are provided', () => {
+    expect(safePluck(obj, ['incorrect-key'] as any)).toEqual(undefined);
+  });
```
hoebbelsB, line 33:
```suggestion
  describe('edge cases', () => {
    it('should return undefined when object is not valid', () => {
      expect(safePluck([undefined] as any, ['prop1'] as any)).toEqual(undefined);
      expect(safePluck(undefined as any, ['prop1'] as any)).toEqual(undefined);
      expect(safePluck('' as any, ['prop1'] as any)).toEqual(undefined);
      expect(safePluck(0 as any, ['prop1'] as any)).toEqual(undefined);
      expect(safePluck([], ['length'])).toEqual(undefined);
      expect(safePluck({} as any, ['prop1'] as any)).toEqual(undefined);
    });

    it('should return null when object is null', () => {
      expect(safePluck(null as any, ['prop1'] as any)).toEqual(null);
      expect(safePluck(null as any, [] as any)).toEqual(null);
      expect(safePluck(null as any, null as any)).toEqual(null);
    });

    it('should return undefined when keys are invalid', () => {
      expect(safePluck(obj, null as any)).toEqual(undefined);
      expect(safePluck(obj, [{}] as any)).toEqual(undefined);
      expect(safePluck(obj, [0] as any)).toEqual(undefined);
      expect(safePluck(obj, [''] as any)).toEqual(undefined);
      expect(safePluck(obj, [undefined] as any)).toEqual(undefined);
      expect(safePluck(obj, [] as any)).toEqual(undefined);
    });
  });
```

hoebbelsB: adapted test cases to really state what is getting returned and making sure every edge case is tested

libs/state/src/lib/core/utils/safe-pluck.ts
```diff
@@ -1,34 +1,96 @@
 import { isKeyOf } from './typing';
+import { isObject } from 'util';
+
+export function safePluck<T extends object, K1 extends keyof T>(
+  stateObject: T,
+  keys: [K1]
```
hoebbelsB, line 6:
```suggestion
  keys: K1 | [K1]
```
safePluck can work with just providing a single key as well -> for convenience

libs/state/src/lib/core/utils/safe-pluck.ts
```diff
@@ -1,34 +1,96 @@
...
+  if (!keys || keys.length <= 0 || Object.keys(stateObject).length <= 0) {
+    return undefined;
+  }
+  // clone keys in order to not mutate input
+  const stateKeys = [...keys];
+  let prop = stateObject[stateKeys.shift() as K1];
+  stateKeys.forEach((key) => {
+    if (isObject(prop) && isKeyOf(key)) {
       prop = prop[key];
     }
```
hoebbelsB, line 95:
```suggestion
  // needed to match null and undefined conventions of RxAngular core components
  // safePluck(null) -> return null
  // safePluck(undefined) -> return undefined
  // safePluck(obj, ['wrongKey']) -> return undefined
  // safePluck(obj, ['correctKey']) -> return value of key
  // safePluck(obj, '') -> return undefined
  // safePluck(obj, null) -> return undefined
  if (!isDefined(stateObject)) {
    return stateObject;
  }
  if (!isDefined(keys)) {
    return undefined;
  }
  // sanitize keys -> keep only valid keys (string, number, symbol)
  const keysArr = (Array.isArray(keys) ? keys : [keys])
    .filter(k => isKeyOf<T>(k));
  if (keysArr.length === 0 || !isObject(stateObject) || Object.keys(stateObject).length === 0) {
    return undefined;
  }
  let prop = stateObject[keysArr.shift() as K1];

  keysArr.forEach((key) => {
    if (isObject(prop) && isKeyOf(key)) {
      prop = prop[key];
    }
```

libs/state/src/lib/rxjs/operators/distinctUntilSomeChanged.ts
```diff
@@ -1,15 +1,7 @@
...
-
-/**
- * @internal
- */
-function safePluck<T extends object, K extends keyof T>(obj: T, key: K): T[K] {
-  return obj !== null && obj !== undefined
-    ? obj[key]
-    : ((obj as unknown) as T[K]);
-}
+import { safePluck } from '../../core/utils/safe-pluck';
```
BioPhoton, line 4: Nice!!

BioPhoton:
@hoebbelsB @Karnaukhov-kh 

Is this ready for merge?

libs/state/spec/core/utils/safe-pluck.spec.ts
```diff
@@ -14,6 +14,7 @@ describe('safePluck', () => {
 
   it('should return value of last key', () => {
     expect(safePluck(obj, ['foo', 'bar'])).toEqual(bar);
+    expect(safePluck(obj, 'foo')).toEqual(obj.foo);
```
hoebbelsB, line 17: ty :)

hoebbelsB:
> @hoebbelsB @Karnaukhov-kh
> 
> Is this ready for merge?

@BioPhoton yes from my point of view!

PR #199, opened 2020-08-02
Scheduling change detection on error and complete and aborting on teardown

kajetansw:
## Problem

When using `rxLet` directive and binding an observable to it, internally, change detection is scheduled whenever an observable notification (next/error/complete) is sent. There's a possibility (a slight, but...) that a notification is sent and before the actual change detection is triggered, the component is destroyed and stops existing. In other words - change detection is triggered on the non-existing component. That may result with errors or unexpected behaviors.

The other problem this PR solves is triggering change detection on error and complete notifications by using `RenderStrategy#detectChanges` method instead `RenderStrategy#scheduleCD`. By using the second one we gain some rendering optimizations specific for the provided rendering strategy. 

## How to reproduce?

The best way to actually see it is to change `libs/template/src/lib/core/render-aware/render-aware_creator.ts`, lines 88 and 92 (`master` branch) and use `scheduleCD()` method instead of `detectChanged()`. 

```ts
// libs/template/src/lib/core/render-aware/render-aware_creator.ts

...
tap({
// handle "error" and "complete" notifications for Observable from template
  error: err => {
    console.error(err);
    if (cfg.updateObserver.error) {
      cfg.updateObserver.error(err);
      currentStrategy.detectChanges(); // change to --> currentStrategy.scheduleCD();
    }
  },
  complete: cfg.updateObserver.complete
    ? () => currentStrategy.detectChanges() // change to --> currentStrategy.scheduleCD();
    : undefined
})
...
```

Tests start to fail on assertions from the `detectChanges` function from the core Angular. This is because Angular is not happy with triggering change detection on component that doesn't actually exist anymore. Tests are so fast, that they destroy a component under test before the final change detection happens.

## Solution

Solution is to define some custom teardown logic for an observable from the view and abort scheduled CD there. This is done with a custom `finalizeWithScheduledCD` operator.

### DIFF
--- a/libs/template/spec/let/let.directive.template-binding.all.spec.ts
+++ b/libs/template/spec/let/let.directive.template-binding.all.spec.ts
@@ -56,8 +56,6 @@ describe('LetDirective when template binding with all templates', () => {
   it('should render "complete" template on the observable completion', () => {
     // The resulting synchronous notification sequence is: 1,2,3,complete
     component.value$ = of(1, 2, 3);
-    // The resulting synchronous notification sequence is: 1,2,3,complete 
-    component.value$ = of(1,2,3);
     fixture.detectChanges();
     expectContentToBe('complete');
   });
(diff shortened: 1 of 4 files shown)


BioPhoton:
I can't make the tests fail. Even if I remove the Aboretcontroller handling completely :D

I will try to enforce the issue first and that use your fix again.

PR #215, opened 2020-08-09
docs(test-coverage): added test coverage to docs folder

BioPhoton:
This PR implements testbade generation and moves the output in the docs folder.

This has a couple of implications
- This means we would self host our test coverage bades
- We can have badges per repo
- we can hose Istambul output as-is
- we lose/drop corealls
- we have to integrate test-docs generation in the release step 

Let me know your thoughts on that.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 # See http://help.github.com/ignore-files/ for more about ignoring files.
 
 # compiled output
+/docs
 /dist
 /tmp
 /out-tsc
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -1,6 +1,7 @@
 # See http://help.github.com/ignore-files/ for more about ignoring files.
 
 # compiled output
+/docs
```
BioPhoton, line 4:
```suggestion

```

PR #222, opened 2020-08-12
Fix TemplateManager detaching template too often

hoebbelsB:
# Description

The current implementation of `TemplateManager` is _detaching_ the `ViewContainerRef` on every call to `insertEmbeddedView`.

```ts
// template-manager_creator.ts line 43 (why "_" btw?)
insertEmbeddedView(name: RxTemplateName) {
  if (templateCache.has(name)) {
    // detach currently inserted view
    viewContainerRef.detach();

    if (viewCache.has(name)) {
      viewContainerRef.insert(viewCache.get(name));
    } else {
      // creates and inserts view to the view container
      const newView = viewContainerRef.createEmbeddedView(
        templateCache.get(name),
        viewContext
      );
      viewCache.set(name, newView);
    }
  }
},
```

This method gets called on every new value emitted by the `UpdateObserver` via the `LetDirective`

```ts
// let.directive.ts line 179

next: (value: U | null | undefined) => {
      this.templateManager.insertEmbeddedView('rxNext');
      this.templateManager.updateViewContext({
        $implicit: value,
        rxLet: value,
   });
},
```

This leads to a complete re-render of the component on any next value. We do not really want that I guess. It certainly breaks the complete behavior of `animations` bound by `*rxLet`.

# Steps to reproduce

I have implemented an example in the experiments app which showcases this exact behavior.
* comment out my fix/or build with the master branch
* `yarn nx s experiments`
* navigate to the "animations" section
* Press the "Toggle grow" button
* to see that it is really an issue of the `*rxLet`, just use the `push` instead. It will work

# Fix
My suggestion is very simple. Just hold the "activeView" as a local variable and call `ViewContainerRef#detach` only if it changes. This way the bug is solved.

I am happy to hear your feedback!


BioPhoton:
Nice catch!

We lost that on our way :D

BioPhoton:
Hi @hoebbelsB Why did you close the PR?

I didn't take a look, at the code, so I may miss something....

hoebbelsB:
> Hi @hoebbelsB Why did you close the PR?
> 
> I didn't take a look, at the code, so I may miss something....

image

BioPhoton:
UUuuuupppps :D

BioPhoton:
Added initial tests to TemplateManager. Even thou there was a good test coverage for from reason... Hmmm.
Added Errors and found a couple of bugs in the test setup, also in our demo app.

The caching mechanism is now better documented and detaches and inserts take the optional argument to show that we remove and push to an array at a position. And therefore we have only one template visible at a time.


IMHO We have to refactor the displayView completely or create a new method before having this part super clean.
rxFor will most probably demand that. But.... that's a long road to there.. :D

I also left @Todos for @kajetansw to give a bit more information on why we do that. Even if I knew all that it took me a while to really understand why certain things got called. Please resolve the open TODOs

libs/template/spec/core/utils/template-manager_creator.spec.ts
```diff
@@ -0,0 +1,183 @@
...
+
+  it('addTemplateRef should throw if no TemplateRef is passed', () => {
+    testComponent.manager.addTemplateRef('templateRefA', testComponent.templateRefA);
+    expect(function() {
+      testComponent.manager.addTemplateRef('templateRefA', testComponent.templateRefA);
+    })
+      .toThrow(new Error('Updating an already existing Template is not supported at the moment.'));
+  });
+
+  it('insertEmbeddedView should create and insert a created view out of a registered template and increment internal Caches', () => {
```
Karnaukhov-kh, line 99:
```suggestion
  it('displayView should create and insert a created view out of a registered template and increment internal Caches', () => {
```

libs/template/spec/core/utils/template-manager_creator.spec.ts
```diff
@@ -0,0 +1,183 @@
...
+    expect(_m.templateCache.size).toBe(1);
+    expect(_m.viewCache.size).toBe(0);
+    expect(testComponent.viewContainerRef.length).toBe(0);
+    templateManager.displayView('templateRefA');
+    expect(_m.templateCache.size).toBe(1);
+    expect(_m.viewCache.size).toBe(1);
+    expect(testComponent.viewContainerRef.length).toBe(1);
+  });
+
+  it('insertEmbeddedView should create and insert a created view out of a registered template and call createEmbeddedView', () => {
```
Karnaukhov-kh, line 111:
```suggestion
  it('displayView should create and insert a created view out of a registered template and call createEmbeddedView', () => {
```

libs/template/spec/core/utils/template-manager_creator.spec.ts
```diff
@@ -0,0 +1,183 @@
...
+    templateManager.addTemplateRef('templateRefA', testComponent.templateRefA);
+    testComponent.viewContainerRef.createEmbeddedView = createSpy('createEmbeddedView');
+    testComponent.viewContainerRef.insert = createSpy('insert');
+    templateManager.displayView('templateRefA');
+    expect(testComponent.viewContainerRef.createEmbeddedView).toHaveBeenCalledTimes(1);
+    expect(testComponent.viewContainerRef.insert).toHaveBeenCalledTimes(0);
+  });
+
+  it('insertEmbeddedView should throw if a unregistered registered template is used', () => {
```
Karnaukhov-kh, line 120:
```suggestion
  it('displayView should throw if an unregistered registered template is used', () => {
```

PR #235, opened 2020-08-15
State docs fixes

Karnaukhov-kh:
### What Problem/Issue is it trying to solve?

Fixes in GitHub `.md` files for upcoming web docs.

### Change List

- Relative links for images replaced with https://raw.githubusercontent.com/.
- ⚠️ Coverage badges links also changed to raw file links.
- Relative links in the state docs replaced with global links.
- $ heading in `rx-state.md` changed to $ (state observable). Otherwise, it's not working as anchor link :)

Issue #217, opened 2020-08-09
State 1.4.0 Release


BioPhoton:
Todos:
- [x] RxState.get #193 #197 
- [x] Typeinference #200 
- [x] github pages 
  - [x] Docs
  - [x] Test Coverage

PR #237, opened 2020-08-15
State documentation for web page

Karnaukhov-kh:
### Merge After

#235 - changes in GitHub `.md` files needed to correctly display content on web page.

### What Problem/Issue is it trying to solve?

#217 - Docs checkbox :)

### Change List

- [Docsify](https://docsify.js.org/#/) generator added that will parse `.md` files to `html`.
- Docs skeleton pages added to `docs/web` folder. `.md` files form the `libs/state/docs` are imported to those pages.
- Michael added [netlifyCMS](https://www.netlifycms.org/) that will help us manage `.md` files better. Setup files are in `docs/admin`.
- Netlify collections were changed. Now we have separate collections for each subfolder in `docs/web` and `libs/{libName}/docs`

### How to Test it

- Checkout branch
- run `yarn run nx serve-docsify docs` and open localhost:3000 to see how web page for the docs will look like
- run `yarn run nx serve-admin docs` and open localhost:5000 to see how admin page for `.md` files look like


angular.json
```diff
@@ -513,15 +513,35 @@
...
+              {
+                "command": "npx netlify-cms-proxy-server"
+              }
+            ]
+          }
+        },
+        "serve-docsify": {
+          "builder": "@nrwl/workspace:run-commands",
+          "options": {
+            "parallel": true,
```
BioPhoton, line 537: We only have one command here. I suggest to remove it .
```suggestion
```

libs/state/src/lib/rx-state.service.ts
```diff
@@ -34,7 +34,7 @@ type ProjectValueReducer<T, K extends keyof T, V> = (
  * @description
  * RxState is a light-weight reactive state management service for managing local state in angular.
  *
- * ![state logo](https://raw.githubusercontent.com/BioPhoton/rx-angular/master/libs/state/images/state_logo.png)
+ * ![state logo](https://raw.githubusercontent.com/BioPhoton/rx-angular/master/docs/images/state/state_logo.png)
```
BioPhoton, line 37: I saw the issue about renaming master to main. Should we consider this here? Or would this introduce too much over these and we create a new issue for the steps todo?

angular.json
Karnaukhov-kh: Agree, thanks 👍

libs/state/src/lib/rx-state.service.ts
Karnaukhov-kh: It might create some issues.

- All open PRs and GitHub pages targeting master. Not sure the target branch will be changed automatically.
- How it will affect git actions and other automation tools that also target master?
- We will need to update links (but that's not the biggest effort)

GitHub will release tools for seamless migration later this year. 
More info here: https://github.com/github/renaming

PR #239, opened 2020-08-16
fix(template): replaced custom logic with materialized

BioPhoton:
This PR is an example on how we could also unify the channels. 
But I 'm sure the custom operator is faster.

loc vs maintenance .. opinions?


kajetansw:
1. Why didn't I know about that `materialize` operator earlier...?
2. I would add this after line 94 (after `tap(cfg.updateObserver)` part) to log the error:
```ts
catchError(err => {
  console.error(err);
  return EMPTY;
}),
```
3. If it basically does what we did with our custom logic, I would keep it. Do you think it would give us trouble in the future?

BioPhoton:
We will keep it as is :)

PR #241, opened 2020-08-16
Move unused code to experimental

BioPhoton:


libs/template/src/index.ts
```diff
@@ -3,7 +3,7 @@ export {
   priorityTickMap,
   SchedulingPriority,
 } from './lib/render-strategies';
-export { getZoneUnPatchedApi, isNgZone, isViewEngineIvy } from './lib/core';
```
Karnaukhov-kh, line 6: We need to add it somewhere to make it accessible by `apps/experiments`. Or remove this check from `config-pannel.component.ts`. Otherwise, experiments build will fail. And experiments build is running for each PR.

libs/template/src/index.ts
```diff
@@ -3,7 +3,7 @@ export {
...
+export { getZoneUnPatchedApi, isNgZone } from './lib/core';
```
BioPhoton, line 6:
```suggestion
// @TODO clarify if we should expost this
export { getZoneUnPatchedApi, isNgZone, isViewEngineIvy } from './lib/core';
export { getZoneUnPatchedApi, isNgZone } from './lib/core';
```

kajetansw:
Hi, @BioPhoton! I had the same build error, but it went away after merging with master ;)

PR #254, opened 2020-08-23
fix(template): fix let directive initial view

hoebbelsB:
### Description

The let directive currently calls `this.templateManager.displayView('rxNext');` in the `ngOnInit` method when no `rxSuspense` template was found. This is not only an issue as this is against the idea of "lazy rendering". 
If your value you bound to the `LetDirective` does not emit any value until the `LetDirective` renders it's initial view, you will end up in runtime errors since all your ViewBindings are now bound to an undefined object.
Also, in my opinion the `LetDirective` should really not render _ANYTHING_ until the first "defined" ( !== undefined ) value appears.  Of course only if you haven't provided a `rxSuspense` template.

### Steps to reproduce

* revert changes in `let.directive.ts`
* `yarn nx s tour-of-heroes`
* navigate to a detail-view from the dashboard
* you will get an `core.js:4196 ERROR TypeError: Cannot read property 'name' of undefined` error

image

### DIFF
--- a/libs/template/spec/let/let.directive.next.spec.ts
+++ b/libs/template/spec/let/let.directive.next.spec.ts
@@ -57,10 +57,10 @@ describe('LetDirective when nexting values', () => {
     expect(componentNativeElement).toBeDefined();
   });
 
-  it('should render undefined as value when initially undefined was passed (as no value ever was emitted)', () => {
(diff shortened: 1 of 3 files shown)


BioPhoton:
Very nice!

PR #272, opened 2020-08-29
docs(state): fix wrong property name

sebaferreras:
Just a very minor fix in the docs – the name of the property should be `creatureToUpdate` instead of `creatureToRemove`.


Karnaukhov-kh:
Hi. Good catch, thank you 👍

sebaferreras:
Hi @Karnaukhov-kh :) 

I submitted the PR directly from Github (since it was a very minor thing) but noticed that all checks have failed. Next time should I do something different?

Thanks again!

Karnaukhov-kh:
No, this is fine. Looks like right now our CI has some issue with accessing forked repos. We will look into it :).

PR #275, opened 2020-08-29
docs(setAccumulator): Create docs for API, usage and rx-state

### LINKED ISSUE #262: API docs for `setAccumulator` are missing
# Description

The docs in the source code for the `setAccumulator` method is missing.

Go there and see no docs.

# Environment

@rx-angular/state@1.3.4
```
-->

# Related to Other Issues

n/a

# Tasks to Resolve This

Write docs

# Notes

n/a

### PR DESCRIPTION
**Pull request by BioPhoton**
closes #262

### DIFF
--- a/libs/state/src/lib/rx-state.service.ts
+++ b/libs/state/src/lib/rx-state.service.ts
@@ -1,24 +1,17 @@
 import { Injectable, OnDestroy } from '@angular/core';
+import { isObservable, Observable, OperatorFunction, Subscribable, Subscription, Unsubscribable } from 'rxjs';
 import {
-  isObservable,
-  Observable,
-  OperatorFunction,
-  Subscribable,
-  Subscription,
-  Unsubscribable,
-} from 'rxjs';
-import {
+  AccumulationFn,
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: libs/state/src/lib/rx-state.service.ts [hunk 1]
```diff
@@ -79,6 +72,22 @@ export class RxState<T extends object> implements OnDestroy, Subscribable<T> {
...
+   *
+   * Allows to customize state accumulation function.
+   * This can be helpful to implement deep updates and tackle other immutability problems in a custom way.
+   * @example
+   *
+   * ```typescript
+   * const myAccumulator = (state: MyState, slice: Partial<MyState>) => ({
+   * ...state,
+   * ...slice
+   * });
```
Karnaukhov-kh, line 86:
```suggestion
  * const myAccumulator = (state: MyState, slice: Partial<MyState>) => deepCopy(state, slice);
```

Karnaukhov-kh: Example right now is the default behavior. Let's go with something different to not confuse people

PR #276, opened 2020-08-29
Harmonize reactive context

BioPhoton:
This PR normalized the 4 states for all templates.

It introduces:

the 4 states suspense, next, error and complete
RxTemplateObserver holds the Observer callbacks plus the suspense callback
RxViewContext which holds the default properties for a ViewContext

### DIFF
--- a/libs/template/src/lib/core/model.ts
+++ b/libs/template/src/lib/core/model.ts
@@ -1,0 +1,21 @@
+import { NextObserver } from 'rxjs';
+
+export interface RxViewContext<T> {
+  // to enable `let` syntax we have to use $implicit (var; let v = var)
+  $implicit: T;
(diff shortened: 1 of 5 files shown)


libs/template/src/lib/core/model.ts
```diff
@@ -0,0 +1,21 @@
...
+  // set context var suspense to true (var$; let c = $suspense)
+  $suspense: boolean;
+}
+
+/**
+ * @description
+ *
+ * Callbacks for the templates main states: suspense, next, error, complete
+ */
+export interface RxTemplateObserver<T> extends NextObserver<T> {
```
hoebbelsB, line 19: love that!

hoebbelsB (approved):
nothing to add, looks great! TY!!

Issue #202, opened 2020-08-06
Template Beta Release


BioPhoton:
Things to be done for the beta release:

The following parts should be documented in code as well as readme, 100% unit tested, and typed including angulars type guards for templates.

See: 
- [![Coverage Status](https://github.com/BioPhoton/rx-angular/blob/github-pages/docs/test-coverage/template/jest-coverage-badge.svg)](https://biophoton.github.io/rx-angular/test-coverage/template/lcov-report/index.html)

- [x] core folder
  - [x] API Docs
  - [x] Tests
- [x] Push (Features: receive observables, typed template, render with passed strategy or default)
  - [x] API Docs
  - [x] Tests
  - [x] MD Docs
- [x] Let
  - [x] MD Docs
  - [x] API Docs
  - [x] Tests
  - [x] Suspense Template
  - [x] Next Template
  - [x] Error Template
  - [x] Complete Template
  - [x] #222 
- [x] Strategies
  - [x] MD Docs
  - [x] Default is `local`
  - [x] Global
      - [x] API Docs
      - [x] Tests
  - [x] Local
      - [x] API Docs
      - [x] Tests
  - [x] Native
      - [x] API Docs
      - [x] Tests
  - [x] Noop
      - [x] API Docs
      - [x] Tests
- [x] Strategy switch bug (switch strategy at runtime implemented in `render_aware`)
  - [x] Merge PR
  - [x] Tests

Released but experimental stuff (located in `experimental folder` does not have to match the requirements mentioned above.)

kajetansw:
Markdown and API docs for the `LetDirective` provided with https://github.com/BioPhoton/rx-angular/pull/253.

kajetansw:
Supplement tests for the `LetDirective` provided with https://github.com/BioPhoton/rx-angular/pull/265.

PR #280, opened 2020-08-30
test(coalescing-manager): implemented coalescing-manager tests

hoebbelsB:
implemented missing coalescing-manager tests in regards of #202 

image


BioPhoton:
I guess we can remove the default [here](https://github.com/BioPhoton/rx-angular/blob/master/libs/template/src/lib/core/utils/coalescing-manager.ts#L39) as they make no sense...

hoebbelsB:
> I guess we can remove the default [here](https://github.com/BioPhoton/rx-angular/blob/master/libs/template/src/lib/core/utils/coalescing-manager.ts#L39) as they make no sense...

exactly what I was thinking

BioPhoton:
Let me know when you are done!

hoebbelsB:
> Let me know when you are done!

@BioPhoton done

PR #295, opened 2020-09-05
Error values in rx-let

### LINKED ISSUE #285: let directive: pass error value as context to error template when using template anchors

Currently if using template anchors the error template will receive a context $error = true.
With no context of the actual error

https://github.com/BioPhoton/rx-angular/blob/e1fe3b381d85aaf232873a6655bed08b533692d1/libs/template/src/lib/let/let.directive.ts#L334

# Solution

Pass the error as the implicit context for the view:

`this.templateManager.updateViewContext({
       $implicit: error,
       $error: true,
  });`

BioPhoton:
This PR introduces a new value for the `$error` property of `RxViewContext`.

Before:
`$error: boolean`

After:
`$error: false | Error`

Here the question is, if we assign the error value to the `$implicite` property too.
I decided to go with the assignment to the `$error` property as it could have nice opportunities to have the last value before the error also accessible. 

The relevant commit is here:
https://github.com/BioPhoton/rx-angular/pull/295/commits/3120fe06b4491b6a8541a1325b85f4b9b5c7c3dd
I adopted the tests to fit the new behavior.

Furthermore, I grouped the examples folder and introduced 2 new demos for error handling.
I also started to create a set of ghost templates for faster example crafting with suspense templates.

closes #285


codingnuclei:
LGTM

libs/template/src/lib/let/let.directive.ts
```diff
@@ -29,7 +29,7 @@ type RxTemplateName = 'rxNext' | 'rxComplete' | 'rxError' | 'rxSuspense';
 
 export interface LetViewContext<T> extends RxViewContext<T>  {
   // to enable `as` syntax we have to assign the directives selector (var as v)
-  rxLet: T;
+  rxLet: T | Error;
```
BioPhoton, line 32:
```suggestion
  rxLet: T;
```

PR #298, opened 2020-09-06
test(static-schedule): fix tests

BioPhoton:


libs/template/src/lib/render-strategies/static/static-schedule.ts
```diff
@@ -14,10 +14,11 @@ export function staticSchedule(
...
   const sub = priorityTickMap[priority].subscribe(
-    () => tryExecuteWork(),
-    (error) => console.error(error),
-    // on complete abort further executions
-    () => abC.abort()
+    {
+      complete: () => tryExecuteWork(),
+      // on complete abort further executions
+      next: () => abC.abort()
+    }
```
hoebbelsB, line 21:
```suggestion
    {
     // on complete abort further executions
     complete: () => abC.abort(),
     next: () => tryExecuteWork()
    }
```

hoebbelsB: I think this should be the correct order :+1:

hoebbelsB (changes requested):
Thx for the PR, I have suggested a fix for you :).
By looking at the code, isn't it weird that we call abort after we executed the work? I don't think this should be the case. Or is there any deeper reason behind this?

Also: aren't we executing `abort` multiple times? currently it gets executed on next() and complete(). Since all of your ticks complete directly after emitting a value, we are executing the `abort` method multiple times

PR #300, opened 2020-09-06
Poc ChangeDetection over EmbeddedView

BioPhoton:
This PR is an experiment only and should collect all needed information for the refactoring as stand-alone click demos.
This is required to do proper performance analysis and gives us an overview of future features and their challanges.

The following, directives should be covered with the poc examples:
- NgForOf
- NgForOfContext
- NgIf
- NgIfContext
- NgSwitch
- NgSwitchCase
- NgSwitchDefault


hoebbelsB:
@BioPhoton Take a look at the switch directive :+1:

libs/template/src/lib/let/let.directive.ts
```diff
@@ -262,6 +262,7 @@ export class LetDirective<U> implements OnInit, OnDestroy {
     templateRef: TemplateRef<LetViewContext<U | undefined | null> | null>
   ) {
     this.templateManager.addTemplateRef('rxComplete', templateRef);
+    this.templateManager.addTemplateRef('rxComplete', templateRef);
```
kajetansw, line 265:
```suggestion
```

PR #302, opened 2020-09-08
Poc rxfor

hoebbelsB:
Experimental PR for upcoming `*rxFor` Directive.

This PR covers two implementations of `rxFor` prototypes:
a very basic prototype to get an impression of what has to be done in order to render multiple templates on a per template basis.
It currently re-creates every template on each change, so there is not that much to see in the example application.

* `yarn nx s experiments`
* navigate to `for-poc/basic` route

And a more advanced prototype which uses the `IterableDiffer` in order to keep track of changes coming from the input observable. Here you can see how only affected templates get rendered when the input values change.

* `yarn nx s experiments`
* navigate to `for-poc/advanced` route

 In the example there is a button which leads to an immutable update of the source value. Try it out :)

### DIFF
--- a/apps/experiments/src/app/app.menu.ts
+++ b/apps/experiments/src/app/app.menu.ts
@@ -6,6 +6,7 @@ import { MENU_ITEMS as LET_MENU_ITEMS } from './let/let.menu';
 import { MENU_ITEMS as MIXED_MENU_ITEMS } from './mixed/mixed.menu';
 import { MENU_ITEMS as CD_OPERATORS_MENU_ITEMS } from './cd-operators/cd-operators.menu';
 import { MENU_ITEMS as RX_STATE_MENU_ITEMS } from './state/rx-state.menu';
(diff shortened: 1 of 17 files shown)


BioPhoton:
Nice @hoebbelsB 

Could you check out #300 and add your example with the differ as 04 there?
Also please try to reduce the code to the minimal amount relevant for your example.

BioPhoton (changes requested):
Nice!

Please reduce the code a lill as suggested. 

Also I think we can poc all features of the for in the experiments. 
I like the differ usage. We should definitely use it. :)

hoebbelsB:
> Nice!
> 
> Please reduce the code a lill as suggested.
> 
> Also I think we can poc all features of the for in the experiments.
> I like the differ usage. We should definitely use it. :)

 @BioPhoton I have just added the iterable version to your PR. Just plain inserted, not reduced. I will do that in the next session

BioPhoton:
Nice. Will close this one in favor of #300

PR #305, opened 2020-09-09
Update usage.md

NechiK:
Replaced old method `setState` with `set`, replaced `this.select()` with `this.state.select()` in places where `RxState` was provided as dependency.


Karnaukhov-kh:
Hi @NechiK. Thank you for this PR 👍 . Looks good to me :)

PR #308, opened 2020-09-11
test(global-strategy): fix tests for markDirty

BioPhoton:


Karnaukhov-kh:
Some of the global strategy tests depending on `markDirty`. It is available only in Ivy.

By default, tests are running with ViewEngine. But we need Ivy. Adding `compilerOptions` or `angularCompilerOptions` with `"enableIvy": true` to tsConfigs doesn't make any difference. So the only way I've found was to run `ngcc` before tests. But that's not the best solution.

If we call `markDirty` without Ivy we see next error `TypeError: Cannot read property 'nodeIndex' of null`.

hoebbelsB:
@BioPhoton thought about adding a second target for the ivy related tests. Doing so would not help us since the target which collects information about code coverage would ignore the ivy related tests.

I am running out of ideas.

PR #309, opened 2020-09-12
WIP. Test new CI tests pipeline

Karnaukhov-kh:


Karnaukhov-kh:
@BioPhoton this is the attempt to run tests in CI with `--runInBand` flag. And it is slower than the current version (34-36 sec with runInBand vs 25 without). So unfortunately it will not help, closing this PR :)

PR #312, opened 2020-09-14
Proposal for updating tests for ɵmarkDirty

kajetansw:
As we are clearly not able to test the internals of the `ɵmarkDirty` function, and that is checking if a proper flag was set on the `__ngContext__` of the component  - we could change the approach.

I trust the Angular team to properly test the internals of the `ɵmarkDirty` function - they created it. Rather than ensuring for the second time that the `ɵmarkDirty` is working as expected, we should focus on our side of the code.

That's why my proposal is rather simple - to check if the call on the `ɵmarkDirty` was made on the component, rather than check the internal functionality of `ɵmarkDirty`.

---

If you think that this solution is not satisfactory, you can go ahead and reject it. It's only a proposal :)

### DIFF
--- a/libs/template/spec/render-strategies/render-strategies/global.strategy.spec.ts
+++ b/libs/template/spec/render-strategies/render-strategies/global.strategy.spec.ts
@@ -1,5 +1,6 @@
 // tslint:disable-next-line:nx-enforce-module-boundaries
 import { getStrategies } from '../../../src/lib/render-strategies';
+import * as AngularCore from '@angular/core';
 import { CallsExpectations, getMockStrategyConfig, testStrategyMethod } from '../../fixtures';
 import { async, ComponentFixture, TestBed } from '@angular/core/testing';
 import { ChangeDetectorRef, Component } from '@angular/core';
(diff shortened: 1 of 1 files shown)


hoebbelsB:
it's fine for me to go with this approach a little while until we find a way to test it properly. We expect a very special behavior when using `markDirty`, our tests should make sure this behavior does not change over time. Any other opinions? @BioPhoton @Karnaukhov-kh

LayZeeDK:
If we're using Jest, we can use `jest.mock('@angular/core');` to do what @kajetansw is suggesting.

kajetansw:
@LayZeeDK Makes things even simpler, thanks!

But for the future reference, is there any way to run tests with Ivy without running `ngcc` before each test run? We wanted to get ahold of the `__ngContext__` of the component inside the test but right now is undefined...

Just setting `enableIvy` flag for the compiler doesn't seem to work.

LayZeeDK:
Probably not as long as we have the Angular Compatibility Compiler and this repository uses a single `package.json`. [Angular will eventually replace the Angular Compatibility Compiler with the Angular Linker](https://github.com/angular/angular/issues/38366).

Karnaukhov-kh (approved):
Hi @kajetansw. Thanks a lot you for this suggestion 👍. Looks good to me. Tests working, coverage for strategies is 100%. However, in total it's still not a 100. There is an untested branch in render-aware. But it is outside of this PR scope.

@BioPhoton @hoebbelsB what do you think? :)

P.S. @LayZeeDK huge thank you for this clarification 😃

BioPhoton:
THX!!! for that one!

Issue #188, opened 2020-07-26
CI - Adopt affected build to only use target version


BioPhoton:
This should fix the failing PRs from forked repositories.

PR #316, opened 2020-09-16
Fix CI checkout from the forked repository 

kajetansw:
## Issue
This PR serves to solve https://github.com/rx-angular/rx-angular/issues/188. With it all potential contributors that use forked repository to create PRs can automatically trigger CI build without failure on the checkout phase.

## Solution
This change configures the CI to point `actions/checkout@v2` action to the correct repository. Previously, each time this action was triggered it pointed to the `rx-angular/rx-angular` repository. This is why it failed when PR was created from the forked repo, because the base branch didn't actually exist in the `rx-angular/rx-angular` repo.

To check if this change doesn't affect PRs from the base repo, the https://github.com/rx-angular/rx-angular/pull/326 draft PR was created to test the behavior.

Screenshots below prove that for the two cases (for the base repo and the forked one) CI points to the right one:

image

image

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -20,12 +20,10 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
-      if: github.event_name == 'pull_request'
       with:
-        fetch-depth: 0
         ref: ${{ github.event.pull_request.head.ref }}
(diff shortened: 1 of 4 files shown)


BioPhoton (approved):
You are our special force!!!

Thank you soo much!!

PR #322, opened 2020-09-19
Improve Document: Manage Global State

adnanebrahimi:
Type of PR:

- Reporting a bug
- Discussing the current state of the code
- Submitting a fix ✔
- Proposing new features
- Becoming a maintainer

Summary:

With the current document of [Managing Global State](https://github.com/rx-angular/rx-angular/blob/master/libs/state/docs/snippets/manage-global-state.md) new developers will following the snippet code to implement it in their code. the code that they implement will not work. because in the snippet code in the AppComponent they way of calling the `connect()` method is wrong: `this.connect("tasks", this.tasksService.fetchTasks())`. 

the correct way is to call the `connect()` method of `state` property: `this.state.connect("tasks", this.tasksService.fetchTasks())`


BioPhoton (approved):
Thanks a lot @adnanebrahimi! <3

PR #328, opened 2020-09-21
feat(code-component): implement code component into demo app

BioPhoton:
# WIP


BioPhoton:
After some research I believe I have to copy:
https://github.com/googlearchive/code-prettify/blob/master/src/prettify.js into our codebase.

Furthermore I would need to use webpack `raw-loader` to get the string of ts files.

PR #339, opened 2020-09-25
docs(experiments): change-detection demos

BioPhoton:
# WIP

There is a problem with the markForCheck demo, other than this it should be fine. Maybe somebody can have a look


hoebbelsB:
@BioPhoton the `markForCheck` demos were broken because the button events are `unpatched`. I have added methods for `markDirty` and a patched version of `markForCheck` as well to see the difference.

BioPhoton:
Very nice @hoebbelsB, Thx for your time.

Imho we should somehow document or at least put some text some where for us internally to explain this situation.

PR #349, opened 2020-10-02
fix(transformation-helpers): log warning only if source is not empty.

Karnaukhov-kh:
Empty array considered as valid input for `toDictionary` helper. So there should be no warning about unexpected input.

thx @hoebbelsB


hoebbelsB (approved):
nice finding and thx for fixing it. looks fine for me. Do you think there is any way of testing if this additional if is causing the performance to decrease?

hoebbelsB:
can you fix the failing checks?

Karnaukhov-kh:
> nice finding and thx for fixing it. looks fine for me. Do you think there is any way of testing if this additional if is causing the performance to decrease?

Sure, will test today

> can you fix the failing checks?

Yup

Karnaukhov-kh:
@hoebbelsB looks like no big difference in performance. Tested on a huge amount of data (300.000 elements). `performance.now()` approach.

image

About errors - you or Michael already fixed it, so I just merged master to this branch 😄.

PR #361, opened 2020-10-10
(POC) Change strategy at runtime

Karnaukhov-kh:
Service holds:
- Default strategy name
- Default out of viewport strategy name

You can:
- Get current strategy name
- Get current out of viewport strategy name
- Set current strategy name
- Set current out of viewport strategy name


BioPhoton:
Love it!! Thanks! this helps to setup the POC for embeddedView rendering!

PR #362, opened 2020-10-10
docs(rxLet): add preloading techniques example

BioPhoton:
Implementation:

```html
      <img [src]="url" *rxLet="valP.imgUrl$; let url; suspense:sV;"/>
      <ng-template #sV><img [src]="placeholder"/></ng-template>
```

```html
    <ng-container [ngComponentOutlet]="c" *rxLet="component$; let c; suspense:suspenseView"></ng-container>
    <ng-template #suspenseView><rxa-list-item-ghost></rxa-list-item-ghost></ng-template>
```
:)


kajetansw:
Nice example! One thing that I've noticed while going through the code.

While clicking the "Reset" you invoke the `PrimitivesProviderService.resetObservables` method and select Observables for the `RxState` via `select` and `hold` over and over again without disposing previously "hooked" Observables. That results in stacking up a big amount of Observables in the memory (bigger with any click, in fact).

To try this out by just putting some `console.log` at the beginning of the `PrimitivesProviderService.resetObservables` method and see how many logs you get with each consecutive click of the "Reset" button.

Maybe injecting `RxState` via DI and then invoking `this.state.ngOnDestroy(); this.state = new RxState();` at the beginning of `PrimitivesProviderService.resetObservables` would solve the problem :)

BioPhoton:
@kajetansw could you PR the fix in a separate one, please?

Also, I added another example for components :) What do you say?

BioPhoton:
The demos app will now be deployed on GitHub pages whenever we push to master 🚀🚀🚀

kajetansw:
> @kajetansw could you PR the fix in a separate one, please?
> 
> Also, I added another example for components :) What do you say?

Here you go: https://github.com/rx-angular/rx-angular/pull/363 :) Resetting image via the "Reset" button seem to be working now. Also, internal Observables are unsubscribed from when resetting.

kajetansw:
> The demos app will now be deployed on GitHub pages whenever we push to master 🚀🚀🚀

Cool, thank you! Just a nitpick: wouldn't it be more descriptive and approachable (for others) if you would do it as a separate PR? This change is not really relevant to the "Preloading techniques example" PR 😛

.github/workflows/build-and-test.yml
```diff
@@ -1,7 +1,7 @@
 # This workflow will do a clean install of node dependencies, build the source code and run tests across different versions of node
 # For more information see: https://help.github.com/actions/language-and-framework-guides/using-nodejs-with-github-actions
 
-name: rx-angular CI
+name: Rx-Angular CI
```
BioPhoton, line 4:
```suggestion
name: rx-angular CI
```

BioPhoton:
Moved GitHub action into separate PR: https://github.com/rx-angular/rx-angular/pull/364

PR #365, opened 2020-10-10
RenderAware fix

hoebbelsB:
fixed custom `renderWithLatestStrategy` operator. It completed instantly which sometimes leads to over-emitting push pipes. The `coalesceWith` operator emits instantly when the subscription completes.


libs/template/src/lib/core/render-aware/render-aware_creator.ts
```diff
@@ -104,7 +113,7 @@ function renderWithLatestStrategy<T>(
       withLatestFrom(strategyChanges$),
       // always use latest strategy on value change
       switchMap(([renderValue, strategy]) =>
-        of(renderValue).pipe(strategy.rxScheduleCD)
+        concat(of(renderValue), NEVER).pipe(strategy.rxScheduleCD)
```
BioPhoton, line 116: Lovely :) hahaha

PR #369, opened 2020-10-11
feat(templateManager): add getEmbeddedView tests, fix selector

BioPhoton:


libs/template/src/lib/core/utils/template-manager_creator.ts
```diff
@@ -86,6 +95,14 @@ export function createTemplateManager<C extends object, N extends string = strin
       }
     },
 
+    getEmbeddedView(name: N): EmbeddedViewRef<C> {
+      if (viewCache.has(name)) {
+        return viewCache.get(name);
+      } else {
+        return undefined;
```
kajetansw, line 102: Nitpick: `Map.prototype.get()` seems to return `undefined` by itself when no value is provided for a given key.

Reference: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/get

libs/template/src/lib/core/utils/template-manager_creator.ts
```diff
@@ -19,6 +19,15 @@ export interface TemplateManager<C extends object, N extends string = string> {
    */
   addTemplateRef(name: N, templateRef: TemplateRef<C>): void;
 
+  /**
+   * @description
+   * Returns a template from the internal templateRefCache map.
+   *
+   * @param name
+   * @param templateRef
```
kajetansw, line 27: This param does not exist.

kajetansw:
Aside from two things above - looks fine by me 👍

libs/template/src/lib/core/utils/template-manager_creator.ts
```diff
@@ -86,6 +95,14 @@ export function createTemplateManager<C extends object, N extends string = strin
       }
     },
 
+    getEmbeddedView(name: N): EmbeddedViewRef<C> {
+      if (viewCache.has(name)) {
+        return viewCache.get(name);
+      } else {
+        return undefined;
+      }
```
BioPhoton, line 103:
```suggestion
 return viewCache.get(name);
```

BioPhoton:
```suggestion
      return viewCache.get(name);
```

libs/template/src/lib/core/utils/template-manager_creator.ts
BioPhoton, line 27:
```suggestion
```

BioPhoton:
Unfortunately I can't apply my own change suggestions :D

PR #371, opened 2020-10-11
Home screen for demos app

kajetansw:
As we are preparing to go "public:" with the Demos app (deploying it via GitHub Actions, see https://github.com/rx-angular/rx-angular/pull/364) I thought it would be a good idea to have some basic home page with short descriptions of each demo section.

Feel free to suggest changes to provided descriptions, they don't necessarily have to be considered final. Especially the "Integrations" section, that I just didn't know how to describe.

If you think any other for of a home page would be better - feel free to suggest it 😊


apps/demos/src/app/home/home.component.ts
```diff
@@ -0,0 +1,32 @@
...
+      Demos presenting fundamental rules and functionalities existing in Angular that are good to know before you start your journey with RxAngular.
+
+      <li><h4>🧰 Template</h4></li>
+      Playground for different functionalities of the <code>@rx-angular/template</code> package.
+
+      <li><h4>📋 Tutorials</h4></li>
+      Tutorials for features existing in the <code>@rx-angular</code> packages.
+
+      <li><h4>🧮 Integrations</h4></li>
+      ???
```
BioPhoton, line 20: Rx-Angular/state integrated with different libs and patterns? We will have the heroes demo, the example with global state, and  others there

PR #379, opened 2020-10-15
Chrome Devtools Protocol integrated with Cypress tests

kajetansw:
# What 

This PR (the source branch itself, to be precise) serves as a shareable POC for integrating Chrome Devtools Protocol with our Cypress tests' configuration.

# Why

This change would potentially enable us to generate reports (flame charts) programmatically - using automated e2e tests. Those charts are normally accessible from the Chrome Devtools via "Performance" and "JavaScript Profiler" tabs.

# How

### What is CDP?

I configured Cypress to work with the [`chrome-remote-interface`](https://github.com/cyrus-and/chrome-remote-interface) library by enabling it via Cypress [tasks](https://docs.cypress.io/api/commands/task.html). See `apps/demos-e2e/src/plugins/index.js` for a reference.

[Chrome Devtools Protocol docs](https://chromedevtools.github.io/devtools-protocol/) show the whole API from this protocol. `chrome-remote-interface` is a library that provides JS counterparts for the available methods of the CDP. I used `Profiler` methods for generating "JavaScript Profiler" reports and `Tracing` for "Performance" reports.

### What tests are there?

For now, there are 2 example tests and 2 flame chart reports generated per test = 4 reports generated after the whole test suite. Tests are performed on the "RxLet vs Push pipe" demo page and the single test consists of clicking the "Toggle" button 6 times resulting in generating a list of 10000 elements 3 times. Each action has a 1 second of artificial delay so that the results on the flame chart would be more readable.

### How to run tests?

To run tests run:
```
yarn nx run demos-e2e:e2e --watch
```

The `--watch` flag is optional. Report files will be generated in the directory of your codebase. Those files can be loaded to the "JavaScript Profiler" or "Performance" tabs in Devtools. Or you can use [Devtools Timeline Viewer](https://chromedevtools.github.io/timeline-viewer/) to also see it.


apps/demos-e2e/src/integration/app.spec.ts
```diff
@@ -1,13 +1,57 @@
...
+  cy.wait(WAIT_DURATION);
+  cy.get('button').contains('Toggle').click();
+  cy.wait(WAIT_DURATION);
+}
+
+function testOpeningAndClosingToggleThreeTimes() {
+  openAndCloseToggle();
+  openAndCloseToggle();
+  openAndCloseToggle();
+}
```
BioPhoton, line 57: should we move that in a po file?

apps/demos-e2e/src/plugins/index.js
```diff
@@ -1,22 +1,154 @@
...
+        `./tracing${params.suffix ? '_' + params.suffix : ''}_${currentDateString()}.json`,
+        JSON.stringify(tracingEvents)
+      );
+
+      return Promise.resolve(tracingComplete);
+    },
+  });
 };
+
+function ensureRemoteDebuggingPort(args) {
```
BioPhoton, line 125: Maybe tiny comments why this code is here etc?

BioPhoton (commented):
Very nice snippets!!! <3

apps/demos-e2e/src/integration/app.spec.ts
kajetansw: It's good to set up some basic good practices for writing future tests, event if it is just an example test. So I moved it to a PO file as suggested :)

apps/demos-e2e/src/plugins/index.js
kajetansw: Sure, I've added some more to be more clear for any future devs!

BioPhoton (approved):
Approved! 👍 🥇

PR #380, opened 2020-10-15
Polish ev

BioPhoton:
I had to expose effect related stuff from state
- https://github.com/rx-angular/rx-angular/pull/380/files#diff-8f05a977001447d7dd928512bfa843a648c269217f27bbb1b637d6c025a251af

and adopted and edge case in template
- https://github.com/rx-angular/rx-angular/pull/380/files#diff-8eb54d0227046f1976377cd690353201e4f4c8b22c3a52767df6640f317ce04c


libs/template/spec/core/utils/template-manager_creator.spec.ts
```diff
@@ -146,6 +146,15 @@ describe('TemplateManager', () => {
...
 
+  it('activeView should return name of the active view after activated', () => {
+    expect(templateManager.activeView()).toBe(undefined);
+  });
+
+  it('activeView should return name of the view if set', () => {
+    expect(() => templateManager.displayView('templateRefA'))
+    expect(templateManager.activeView()).toBe('templateRefA');
+  });
+
```
BioPhoton, line 157:
```suggestion
```

PR #384, opened 2020-10-16
feat(template-manager): Add optional id to embeddedView name

BioPhoton:
This change is needed to have multiple EmbeddedViews for the same template


hoebbelsB (approved):
i can see the purpose, is there a need to type it more strictly in future? anyways, it doesn't matter for now, since we need this feature

Issue #203, opened 2020-08-06
Template RC preconditions 


BioPhoton:
**Things to be done for the RC releases:**

The following parts should be documented in code as well as readme, 100% unit tested, and typed including angulars typeguards for templates.
- [x] ❗ `nx update`   
- [x] ❗ Demos App  
   - [x] Zone-Full, Zone-Less
   - [x] Performance Section with:
     - [x] Generic/recursive configurable components to control workload (global render queue demo)
     - [x] Standard for: nested structures, lists, frequent updates
     - [x] List of multiple components each having their own let directive. 
      - [x] Changes received over Input bindings
      - [x] Changes received over global service
- [x] ❗🔍 Ensure TemplateManager has no negative performance impact (alpha.0 vs alpha.1)
- [x] ❗🔍 Evaluate the difference of schedule and coalesce or separate #388
  -  ~~Provide flame charts and other docs to save the knowledge~~
- [x] Implementing beta feedback
  - [x] Harmonize naming
  - [x] Measure Performance
  - ~~Docsument PushPipe with getters~~
  - ~~🐞 Fix Routing bug (Navigation triggered outside NgZone)~~
- [x] ⚙️ 
  - [x] 🛂 Test coverage badge
- [x] Strategies
  - [x] 🔍 Clarify usecases of `detectChanges` (usage works atm only with local or detach)
  - [x] Detach
    - [x] 📖 
    - [x] 📚
    - [x] 🛂
- [x] ✍️ Introduce afterRender signal
  - [x] Setup usecase in [`experiments`](https://github.com/rx-angular/rx-angular/tree/master/apps/experiments) app that shows the need
  - [x] Implement signal for imperative and reactive style
  - [x] 💻 Imperative and reactive
  - [x] 📖 
  - [x] 🛂
-  ~~Abort controller chaining testing and solid implementation. [Look at Pr in RxJS](https://github.com/ReactiveX/rxjs/issues/3122)~~ (postponed as removed from strategies in pipes and directives => moved in RenderService)
   -  ~~📖~~ 
   -  ~~🛂~~
- ~~Check if every source file is covered by a separate spec file (avoid indirectly tested lines)~~ (postponed)
- [x] 🔍 Template Typing https://github.com/rx-angular/rx-angular/issues/201
  - [x] Ensure it only happens in VsCode
- [x] 💪 Refine usage of schedulers (current mode, queue scheduler, postTask). 
  - [x] Harmonize signatures #294
[...]

PR #388, opened 2020-10-19
strategy change demo

Karnaukhov-kh:
What's included:
- `rx-let-poc.service` that wraps CDRef, holds current visible/out of viewport strategies and set/get functionality.
- `rx-let-poc.directive` that uses `rx-let-poc.service` to change strategies.
- RxLet POC demo with the possibility to change strategy at runtime (in experiments).

Removed:
- `strategies-setup.service` and all related stuff removed from src/experimental 😬.

related #203


libs/template/src/lib/core/render-aware/interfaces.ts
```diff
@@ -5,6 +5,8 @@ export interface StrategySelection {
   [strategy: string]: RenderStrategy;
 }
 
+export type RenderStrategyFactory = (config: RenderStrategyFactoryConfig) => StrategySelection
+
```
BioPhoton, line 9:
```suggestion
```

libs/template/src/lib/render-strategies/util/util.ts
```diff
@@ -1,5 +1,19 @@
 import { fromEvent, Observable } from 'rxjs';
 import { takeUntil } from 'rxjs/operators';
+import { RenderStrategy, StrategySelection } from '@rx-angular/template';
+
+
+
```
BioPhoton, line 6:
```suggestion

```

BioPhoton (commented):
changed in libs look good

Issue #303, opened 2020-09-08
Insert browser support


BioPhoton:
https://godban.github.io/browsers-support-badges/

PR #392, opened 2020-10-22
add upsert transformation helper

hoebbelsB:
### Description

I noticed that an `upsert` transformation helper is missing. Actually, I was in need of this by myself, so I decided to quickly create a PR for this.
The `upsert` transformation helper inserts new items if they don't exist in the source and updates them when they are present. It shares the same API as `update` and behaves the same in every edge case.
I copied the specs I needed  from `insert` and `update` and used them for the `upsert` spec.

Documentation is still missing. Otherwise, I am happy about any feedback :+1:


libs/state/src/lib/transformation-helpers/array/upsert.ts
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  if (!sourceDefined || !source.length || !sourceIsArray) {
+    return [...updatesAsArray] as T[];
+  }
+
+  const inserts: T[] = [];
+  const updates: Record<number, Partial<T>> = {};
+  updatesAsArray.forEach(item => {
+    const match = source.findIndex(sourceItem => valuesComparer(item as T, sourceItem, compare));
```
Karnaukhov-kh, line 31: Do we care about IE 11? If yes, `findIndex` is not supported in this browser-like application.

libs/state/src/lib/transformation-helpers/array/upsert.ts
```diff
@@ -0,0 +1,48 @@
+import { isDefined } from '../../core/utils/guards';
+import { valuesComparer } from '../_internals/valuesComparer.util';
+import { ComparableData } from '../interfaces/comparable-data-type';
+
+export function upsert<T extends object>(
```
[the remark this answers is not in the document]
hoebbelsB: I already thought about this but came to the conclusion that this would ultimately make things less performant. Upsert would have to do all the edge case checks (like valid input, is source of correct type, ...etc.) before it can decide to insert and upsert - where all of those checks run again.

So in the end we have 'duplicated' code, but it's more performant. do I miss something?

libs/state/src/lib/transformation-helpers/array/upsert.ts
hoebbelsB: is this really a thing? this library gets transpiled, doesn't it?

libs/state/src/lib/transformation-helpers/array/upsert.ts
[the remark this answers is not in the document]
Karnaukhov-kh: Checked insert/update. You are right. What we can consider as an option in the future is to refactor stuff a bit and move edge cases handling logic to separate helper methods - one for arrays and the second one for objects.

libs/state/src/lib/transformation-helpers/array/upsert.ts
Karnaukhov-kh: In ng10+ yes, in previous versions, there might be some issues. 

But the proper fix for previous ng versions and old browsers will be to downgrade ng-packagr back to version 9.X.X, so it will produce fesm5 and esm5. Not sure if we should do that :)

hoebbelsB: i don't know if we should take care of browsers which are deprecated especially when keeping in mind that we are offering cutting edge technology with rx-angular. @BioPhoton what's your opinion about this?

libs/state/src/lib/transformation-helpers/array/upsert.ts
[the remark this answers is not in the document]
hoebbelsB: yes, there is room for improvement!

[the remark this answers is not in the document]
BioPhoton: > Checked insert/update. You are right. What we can consider as an option in the future is to refactor stuff a bit and move edge cases handling logic to separate helper methods - one for arrays and the second one for objects.

That sounds nice!! And the code would be way easier to understand.

libs/state/src/lib/transformation-helpers/array/upsert.ts
BioPhoton: > Do we care about IE 11? If yes, `findIndex` is not supported in this browser-like application.

We should discuss that one soon! #303

libs/state/src/lib/transformation-helpers/array/upsert.ts
```diff
@@ -0,0 +1,56 @@
...
+  });
+  const updated = Object.keys(updates).length === 0 ?
+                  source :
+                  source.map((item, i) => {
+                    const updatedItem = updates[i];
+                    if (isDefined(updatedItem)) {
+                      if (isObjectGuard(item)) {
+                        return { ...item, ...updatedItem };
+                      } else {
+                        return updatedItem as T;
```
Karnaukhov-kh, line 50: `return updatedItem as T;` is now not covered by tests. And `update` helper works only with objects. I'm not sure how to cover it 🤔.

Karnaukhov-kh (commented):
Hi @hoebbelsB. Not sure why the build fails but otherwise only a small comment about update part. Other parts LGTM.

libs/state/src/lib/transformation-helpers/array/upsert.ts
hoebbelsB: @BioPhoton imo IE11 should not get supported by a cutting edge library. And didn't angular support default IE11 support as well?

hoebbelsB: but if it's just about this line, i wont complain if we change it :+1:

PR #403, opened 2020-10-27
use concatMap, not switchMap to avoid loosing deletes

datumgeek:


BioPhoton:
Good catch!!

datumgeek:
@BioPhoton - can you help with this?

> Your PR was submitted to a repository that is not participating in Hacktoberfest. Maintainers of the repository can add the "hacktoberfest" topic to their repository if they wish to participate. Alternatively, an individual PR can be opted-in with a maintainer adding the "hacktoberfest-accepted" label to the PR.

BioPhoton:
I added the label. Is this OK?

edbzn:
@BioPhoton I think there's other places where this change should belong (demos and docs). Also, what about adding a note about `RxState#hold` and lossy operators? I think many people will use `switchMap` by default as I did.

PR #414, opened 2020-11-03
Zone flag handler

BioPhoton:
setup initial zoneConfigurations helper


apps/demos/src/zone/event-names.ts
```diff
@@ -0,0 +1,120 @@
...
+  'dragover',
+  'dragstart',
+  'drop'
+];
+
+export const formControlsEvents: (keyof WindowEventMap)[] = [
+  'change',
+  'blur',
+  'focus',
+  'contextmenu',
```
hoebbelsB, line 95: is this only for formControls? i guess contextmenu is available on most elements isn't it?

apps/demos/src/zone/zone-flags-configurator.ts
```diff
@@ -0,0 +1,90 @@
...
+type RuntimeConfigurationMethods = {
+  [disabledFlag in ZoneRuntimeConfigurationsKey]: (isDisabled?: boolean) => void;
+}
+
+const zoneDisable = '__Zone_disable_';
+const zoneIgnore = '__Zone_ignore';
+const zoneSymbol = '__zone_symbol__';
+
+
+export interface ZoneFlagsConfigurator {
```
hoebbelsB, line 35: will there be a property named `enable` in future? otherwise I would reconsider the naming here

apps/demos/src/zone/zone-flags-configurator.ts
```diff
@@ -0,0 +1,90 @@
...
+  },
+  target: {
+    disable: GlobalTargetConfigurationMethod
+  }
+  runtime: {
+    disable: RuntimeConfigurationMethods
+  },
+}
+
+function getZoneFlagsConfigurator(): ZoneFlagsConfigurator {
```
hoebbelsB, line 50:
```suggestion
function createZoneFlagsConfigurator(): ZoneFlagsConfigurator {
```

I think this is a creator function. So let's name it like one :)

apps/demos/src/zone/zone-flags-configurator.ts
```diff
@@ -0,0 +1,90 @@
...
+      disable: zoneRuntimeConfigurationsKeys
+        .reduce((map, prop) => ({
+          ...map,
+          [prop]: (isDisabled: boolean = true) => cfg[zoneSymbol + prop] = isDisabled
+        }), {} as RuntimeConfigurationMethods)
+    }
+  };
+}
+
+export const zoneFlagsConfigurator = getZoneFlagsConfigurator();
```
hoebbelsB, line 90:
```suggestion
export const zoneFlagsConfigurator = createZoneFlagsConfigurator();
```

hoebbelsB (commented):
What a GREAT PR!!! I have added some comments about minor naming stuff.

Also, I cannot get the demo to run properly. Google maps won't show me anything :( Is there any configuration missing from my side?

apps/demos/src/zone/event-names.ts
BioPhoton: true! What should be the group?

apps/demos/src/zone/zone-flags-configurator.ts
BioPhoton: No. disable could get false, but false is the default. As all this will be in one file I can't imagine setting it to false and then true again.

What would be a better naming?

apps/demos/src/polyfills.ts
```diff
@@ -52,10 +52,12 @@
  *
  */
 
+import './zone-flags'; // Included before zone.js
```
hoebbelsB: we need some way to load this only on environments, any clue how?

apps/demos/src/zone/zone-flags-configurator.ts
hoebbelsB: okay I got it wrong somehow. Improvements could only be made with some kind of `Record` and a ConditionalType which maps e.g. `global` to `GlobalDisableConfigurationMethods`. Lets go with this version for now i'd say. Otherwise we would need our typing magician @kajetansw

apps/demos/src/zone/event-names.ts
hoebbelsB: i'd say it's a global event. From what I can see in the [docs](https://developer.mozilla.org/en-US/docs/Web/API/Element/contextmenu_event) the event exists on `Element`

hoebbelsB (commented):
got you wrong with the typings of the `Configurator`, sorry. Should we wait for an answer of @kajetansw regarding this or just go with this version? We won't publish this API right away anyway, do we?

Otherwise i've left a comment regarding the `contextmenu` event

apps/demos/src/polyfills.ts
BioPhoton: Target configurations and multiple polyfill files. Or custom build scripts

apps/demos/src/zone/event-names.ts
BioPhoton: So I create one with subgroups

hoebbelsB:
@BioPhoton what about this one? any decisions made so far?

PR #417, opened 2020-11-08
Bootstrap components with/without dependencies

Karnaukhov-kh:
This tiny demo allows users to switch between multiple components.

We have a component that displays a `text`.

- Component without dependencies. `text` property defined inside component.
- Service injected into component. `text` property defined inside the service.
- Component extends a service. `text` property defined inside the service.
- RxState injected into component. `text` property is a part of state.
- Component extends RxState. `text` property is a part of state.
- Component with Async Pipe. `text$` property defined inside component and displayed with `async` pipe.
- Component with Push Pipe. `text$` property defined inside component and displayed with `push` pipe.
- Component with Let Directive. `text$` property defined inside component and displayed with `rxLet` directive.

P.S. Measurements added to measurements repo.


BioPhoton:
Could we please get a *ngFor with ~1000 of each and add timing marks?
I guess this would make it way easier to spot the diff in the flames and also give us a faster way to measure the diff.

Issue #405, opened 2020-10-30
Schematics


BioPhoton:
# Problem Solved By The Feature

We have no schematics :(

# Solution

Schematics are implemented

**State:**
- [ ] `add` (no Modules to register, no postinstall work)
- [ ] `generate` component 
  - [ ] Ask for target component
  - [ ] Ask for inject or extend

**Template:**
- [ ] `add`
  - [ ] Add `LetModule`, `PushModule`, `UnpatchEventsModule` to `app.module.ts`
- [ ] `migration` minimal setup (rename `*rxLet="o$, suspense: view"` to `*rxLet="o$, rxSuspense: view"`)
- [ ] `generate` think about usecases???

sis0k0:
.

edbzn:
:wave: I would be happy to help on this too.

BioPhoton:
Maybe we can move in the base and do different schematics in different PRs? @sis0k0

edbzn:
I started bootstraping schematics files and I'm wondering how to organize it, I see two different alternatives : 

1- We put schematics directly in it's related library under a `schematics` directory. The library holds its schematics files including the build configuration. _This means we need to change the build process to transpile schematics along the library and put these files in the distribution folder._ 

2- We put schematics in dedicated publishable libraries seperated from core libraries (eg: `schematics-state`, `schematics-template`). We can use an a plugin here to let Nx handle the build process for us. _This way we can normally use the `affected:build` command._

What are your thoughts on this?

BioPhoton:
I guess at lease nd-add should be in the libs folders. 

**Folder Structure:**
```
 libs/
├── state
│   ├── src
│   ├── schematics
│   │   └── commands
│   │       └── ng-add
│   │           ├── index.md
│   │           ├── index.ts
│   │           └── schema.json
│   ├── package.json
│   └── collection.json
├── template
│   ├── src
│   ├── schematics
│   │   └── commands
│   │       └── ng-add
│   │           ├── index.md
│   │           ├── index.ts
│   │           └── schema.json
│   ├── package.json
│   └── collection.json
└── schematics
    ├── src
    ├── schematics
    │   └── commands
    │       └── ng-add
    │           ├── index.md
    │           ├── index.ts
    │           └── schema.json
    ├── package.json
    └── collection.json
```

Not sure if we have that much schematics to ship different libs for state and template...
But in any case we should coordinate with @sis0k0 i believe

sis0k0:
[...]

PR #419, opened 2020-11-09
ng-add schematics

edbzn:
This PR adds **ng-add** schematics for both `@rx-angular/state` and `@rx-angular/template` (#405)


BioPhoton:
minor:

Could we use following folder struct:

```
 libs/
├── state
│   ├── src
│   ├── schematics
│   │   └── commands
│   │   │      └── ng-add
│   │   │              ├── index.md
│   │   │              ├── index.ts
│   │   │              └── schema.json
│   │   └── utils
```

edbzn:
@BioPhoton Done. We will need to reuse some functions between schematics, not sure what will be the best place for it, for now I leave it as it.

libs/state/schematics/utils/npm.ts
```diff
@@ -0,0 +1,36 @@
...
+  name: string;
+  version: string;
+}
+
+export function getLatestNodeVersion(
+  packageName: string
+): Promise<NpmRegistryPackage> {
+  const DEFAULT_VERSION = 'latest';
+
+  return new Promise((resolve) => {
```
sis0k0, line 13: I'd suggest keeping track of the versions in a local file instead of checking the npm registry.

libs/state/schematics/commands/ng-add/index.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+  newProjectRoot: 'projects',
+  version: '10.1.0',
+  defaultProject: 'state',
+};
+
+const defaultAppOptions = {
+  name: 'state',
+};
+
+describe('ng-add function', () => {
```
sis0k0, line 17:
```suggestion
describe('ng-add schematic', () => {
```

sis0k0 (approved):
I've added a couple of non-blocking comments. Overall it looks great! Good job!

libs/state/schematics/utils/npm.ts
edbzn: The good part with checking the npm registry is that we don't need to manually update the package version on each release.

edbzn:
I've added a couple of things since last review :

- ng-add schematic for `@rx-angular/template`
  - Add `@rx-angular/template` to package.json dependencies
  - Import `LetModule`, `PushModule` and `ViewportPrioModule` in the given @NgModule (default to app.module)

I see one improvement that we can do : sharing some rules between schematics instead of duplicating them, maybe by using a dedicated library?

Also I didn't added any documentation, I think there's some things to do here.

sis0k0:
I'll take a look at the changes. I like the idea of having a shared library!

BioPhoton:
@edbzn Thanks so much for this PR!
Definitely one of the most meaningful things added to the lib. <3 <3 <3

PR #425, opened 2020-11-13
docs: add pokemon w/ pagination demo

nartc:


BioPhoton:
Thanks @nartc for this PR!!!

I'll have a closer look soon, but wanted to check back quickly to give feedback. :)

apps/demos/src/app/features/integrations/pokemon-pagination/states/pokemon-state.service.ts
```diff
@@ -0,0 +1,105 @@
...
+          filteredResult: data.results,
+          originalResult: data.results,
+        }))
+      );
+  }
+
+  private queryEffect$() {
+    return this.select('query') // watch for query changed
+      .pipe(
+        distinctUntilChanged(), // rate limiting
```
edbzn, line 94: Minor : I think `distinctUntilChanged` here is superfluous since `select` already applies it via the [`stateful`](https://github.com/rx-angular/rx-angular/blob/8b83a6f4773fbc52f2cb31a5c4ecd0d03fd7536c/libs/state/src/lib/rxjs/operators/stateful.ts#L110) operator.

nartc: good catch! thanks

BioPhoton:
Thanks so much for this PR <3 and sorry for the delayed merge!

Issue #426, opened 2020-11-13
Calling update on an empty array with a matcher inserts the updated item in the array


Seqi:
# Description
I have some code that calls update on an array with an object. It works like this:

State:
```
{
   items: Item[]
   selectedItem: Item
}
```

If selected item changes in any way, I want to reflect that change back into the items list. 

```
    // Set the item into selected item
    this.connect(this.setSelectedItem$, (state, item) => setProp(state, 'selectedItem', item))

        // Whenever the selected item changes, feed that change back into the items list
     this.connect('items', this.select('selectedItem'), (state, updatedItem) => 
      update(state.items, updatedItem, 'id')
    )
  }
```

# Expected Behavior
The `items` array remains empty, as no item matched the item to update using the criteria. I do not expect **update** to perform an **insert** action.

# Actual Behavior
The `selectedItem` is inserted into the array as a new item.

# Steps to Reproduce the Issue

Run the following code:

```
    let result = update([], { id: 1 }, 'id')
```

Observe how the result contains the item, even though there are no items in the source array to match the criteria to.

Based on your source, "if source doesnt exist, then initialise the source with the update" seems to be the logic. Is this intended? I would assume that if no matching item in the source would cause no update to happen. If insertion was intended, then `insert` would be used. 

# Tasks to Resolve This

The line causing this is here:
https://github.com/rx-angular/rx-angular/blob/master/libs/state/src/lib/transformation-helpers/array/update.ts#L100

I imagine all this will need is something like:

```
  if (!sourceDefined || (!source.length || !sourceIsArray) {
    return [] as T[];
  }
```

or a less impactful change like:
```
  if (!sourceDefined || !sourceIsArray) {
    return [...updatesAsArray] as T[];
  }
```
so that if the state is in an expected, but empty state, nothing unexpected happens.

# Notes
I have fixed this in my application by checking for an empty array in state myself, and if it is, not calling update.

If this is working as intended, please let me know and close this up and i'll add these empty checks where necessary.

Seems there is some interest (or at least a few comments) in an `upsert` transformer (https://github.com/rx-angular/rx-angular/pull/392/files), which seems to stress the need for `update` to never `insert`. 

Have added a PR to demonstrate suggested changes. Happy to modify if suggestions given.

PR #428, opened 2020-11-13
Do not insert updated item(s) if the source object is invalid or empty

Seqi:
Suggestion of changes for #426 . Happy to modify on suggestion.


Karnaukhov-kh:
Hi @Seqi. Thank you for the feedback & the PR 👍. I'm pretty sure that we should not touch an empty array and it was a bug. But not sure if we should not initialize undefined value and override not valid value. @hoebbelsB @BioPhoton @kajetansw  what do you think?

BioPhoton:
@Karnaukhov-kh could you help me to follow up here?
I'm was not fully involved in the decision making of transformation helpers

hoebbelsB (approved):
LGTM thx

PR #432, opened 2020-11-17
feat(template):detach on complete and error

BioPhoton:
This PR tries to evaluate the pros/cons edgecases of detaching the EmbeddedView after error or complete.

### DIFF
--- a/apps/demos/src/app/shared/rx-angular-pocs/render-stragegies/default-strategies.ts
+++ b/apps/demos/src/app/shared/rx-angular-pocs/render-stragegies/default-strategies.ts
@@ -3,10 +3,6 @@ import { ɵmarkDirty as markDirty } from '@angular/core';
 import { coalesceWith, priorityTickMap, SchedulingPriority } from '@rx-angular/template';
 import { tap } from 'rxjs/operators';
 
-export function getDefaultStrategyNames(): string[] {
-  return Object.keys(getDefaultStrategyCredentialsMap());
-}
-
 export function getDefaultStrategyCredentialsMap(): StrategyCredentialsMap {
   return {
     'global': globalCredentials,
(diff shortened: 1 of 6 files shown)


hoebbelsB (changes requested):
this is a very nice idea. but I think we need to move this into a strategy or something. it will lead to problems, when changes from outside should get reflected in those templates.

just a little example:

```html

            <!-- does not work when template gets detached -->
            <ng-container>{{ errorTmplState$ | push }}</ng-container>
            <!-- works fine when template gets detached -->
            <ng-container *rxLet="errorTmplState$; let state">{{ state }}</ng-container>
```

BioPhoton:
Moved it into the local strategies. I had to extend the work function with the current notification to know the active template

apps/demos/src/app/shared/rx-angular-pocs/render-stragegies/default-strategies.ts
```diff
@@ -18,7 +14,14 @@ export function getDefaultStrategyCredentialsMap(): StrategyCredentialsMap {
...
   name: 'local',
-  work: (cdRef) => cdRef.detectChanges(),
+  work: (cdRef, _, notification) => {
+    if(['rxComplete', 'rxError'].includes(notification.kind)) {
+      cdRef.detach();
+    } else {
+      cdRef.reattach();
+      cdRef.detectChanges();
+    }
+  },
```
hoebbelsB, line 24: so the error and complete template will not receive any changes? don't they receive some context which should get updated?
Also, I think this implementation is very slow, we should store the methods in a map and just call

```ts
notificationWorkMap[notification.kind]();
```

this way we save 1.000s of iterations over the same array again and again.

apps/demos/src/app/shared/rx-angular-pocs/render-stragegies/model.ts
```diff
@@ -1,7 +1,8 @@
 import { Observable } from 'rxjs';
 import { ChangeDetectorRef } from '@angular/core';
+import { RxNotification } from '@rx-angular/template';
 
-export type RenderWork = <T = unknown>(cdRef: ChangeDetectorRef, context: any) => void;
+export type RenderWork = <T = unknown>(cdRef: ChangeDetectorRef, context?: any, notification?: RxNotification<T>) => void;
```
hoebbelsB, line 5: I think this can be quite helpful :)

hoebbelsB (changes requested):
do we really want to have the complete and error templates cold?

apps/demos/src/app/shared/rx-angular-pocs/render-stragegies/default-strategies.ts
BioPhoton: "cold" was confusing at the beginning for me. But got it. 
You mean values from the observable. 

There is no next value after error or complete that could occur from the observable. 
And the triggers for the template has to behave the same. Detach temple and no other CDs affect it. A potential next value will reattach the view.

For the map:
Sure we have to make to nice! Fir the Poc it was important to see what we need to change in the strategy config.

So I will skip any changes on low level implementations.

PR #437, opened 2020-12-11
fix stop stream when throw error in hold method

### LINKED ISSUE #435: Throw error in hold stream stopped all other hold streams
# Description

If one of the hold streams has an error when all other hold streams stoped

Check  [stackblitz](https://stackblitz.com/edit/angular-ivy-mrmn6v?file=src%2Fapp%2Fapp.component.ts) example

```
this.hold(interval(1000), i => console.log(i));
this.hold(
  of(null).pipe(
    delay(5000),
      mergeMap(() => throwError("error"))
  )
);
```

Result in console:
`0
1
2
3
4
error`

# Related to Other Issues

#293 

# Solution

Add catchError in hold definition and return EMPTY when having some error

```
hold<S>(
    obsOrObsWithSideEffect: Observable<S>,
    sideEffectFn?: (arg: S) => void
  ): void {
    if (typeof sideEffectFn === 'function') {
      this.effectObservable.nextEffectObservable(
        obsOrObsWithSideEffect.pipe(catchError((e)=> {
          console.error(e)
          return EMPTY
        }),tap(sideEffectFn))
      );
      return;
    }
    this.effectObservable.nextEffectObservable(obsOrObsWithSideEffect.pipe(catchError((e)=> {
      console.error(e)
      return EMPTY
    })));
  }

```

allfayn:
This PR aims to fix  #435

### DIFF
--- a/libs/state/src/lib/rx-state.service.ts
+++ b/libs/state/src/lib/rx-state.service.ts
@@ -547,9 +547,7 @@ export class RxState<T extends object> implements OnDestroy, Subscribable<T> {
     obsOrObsWithSideEffect: Observable<S>,
     sideEffectFn?: (arg: S) => void
   ): void {
-    const sideEffect = obsOrObsWithSideEffect.pipe(catchError((e)=> {
-      return EMPTY
-    }))
+    const sideEffect = obsOrObsWithSideEffect.pipe(catchError(e => EMPTY))
     if (typeof sideEffectFn === 'function') {
       this.effectObservable.nextEffectObservable(
         sideEffect.pipe(tap(sideEffectFn))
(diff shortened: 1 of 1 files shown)


libs/state/src/lib/rx-state.service.ts
```diff
@@ -547,13 +547,16 @@ export class RxState<T extends object> implements OnDestroy, Subscribable<T> {
     obsOrObsWithSideEffect: Observable<S>,
     sideEffectFn?: (arg: S) => void
   ): void {
+    const sideEffect = obsOrObsWithSideEffect.pipe(catchError((e)=> {
+      return EMPTY
+    }))
```
BioPhoton, line 552: Nitty change suggestion
```suggestion
    const sideEffect = obsOrObsWithSideEffect.pipe(catchError(e=> EMPTY))
```

PR #449, opened 2021-01-05
Update get-global-this.ts

mhartington:


apps/demos/src/app/rx-angular-pocs/cdk/utils/get-global-this.ts
```diff
@@ -11,5 +11,15 @@
...
+    'undefined' != typeof globalThis
+      ? globalThis
+      : 'undefined' != typeof window
+      ? window
+      : void 0 !== n
+      ? n
+      : 'undefined' != typeof self
+      ? self
+      : {};
+  return l;
```
BioPhoton, line 24:
```suggestion
  return 
    'undefined' != typeof globalThis
      ? globalThis
      : 'undefined' != typeof window
      ? window
      : 'undefined' != typeof self
      ? self
      : {};
```
I removed the line you mentioned.
Nit Pic: return it directly.

BioPhoton (approved):
Thanks, mike for that nice fix!!
Was not able to fix is since..... long :D 
<3<3<3

PR #458, opened 2021-01-12
Migrate Nx workspace to v11.1.2

edbzn:
This PR does the following:

- Run the migration script to bump Nx workspace to v11.1.2
- Replace the library build builder `@nrwl/angular:package` by `@nrwl/angular:ng-packagr-lite` which is more optimized


BioPhoton:
sry, merged dependabot first.. can u please rebase

PR #456, opened 2021-01-12
Clean up the basic tutorial

mhamri:
#WHY

the current basic tutorial isn't clear to read, the files are not comparable against each others. there are typos and missing context. 

#WHAT 

- add more description to the files. 
- show how the tutorial is structured. 
- make files comparable against each other
- add more link to any files discussed in the tutorial

PR #459, opened 2021-01-13
Fix instructions for basic tutorial

rene-leanix:
Here's a quick fix to allow users to run the basic tutorial again.

I only saw now that #456 also addresses this and adds a lot of other improvements. Feel free to close my PR then.


BioPhoton:
I will follow your suggestion and close it in favor of #456

PR #467, opened 2021-01-22
fix(viewport-prio): disconnect the observer to allow the directive to be GCed

arturovt:
Hey! I didn't open an issue since this seems to be a tiny change. Currently, the `IntersectionObserver` is not disconnected and prevents the `ViewportPrioDirective` and `LetDirective` to be garbage collected.


BioPhoton:
Hi @arturovt ,

THX for that one!

PR #469, opened 2021-01-23
fix(template): support server-side rendering for the `unpatch` directive

arturovt:


BioPhoton:
Very nice one!!

This reminds me to setup SSR in our demos app!

PR #480, opened 2021-02-02
build: do not publish Ivy compiled packages

arturovt:
Note that `package` builder doesn't support configurations. You can check it with:

```js
$ node
{ execSync } = require('child_process')
packageImpl = require('@nrwl/angular/src/builders/package/package.impl')
run = packageImpl.run
packageImpl.run = (options, context) => {
  console.log(options.tsConfig);
  return run(options, context);
}
execSync('yarn build template --prod --skip-nx-cache').toString()
```

### DIFF
--- a/angular.json
+++ b/angular.json
@@ -514,11 +514,6 @@
           "options": {
             "tsConfig": "libs/state/tsconfig.lib.json",
             "project": "libs/state/ng-package.json"
-          },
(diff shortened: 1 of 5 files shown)


BioPhoton:
Thanks a lot for your time @arturovt ! 🙏🙏🙏

PR #481, opened 2021-02-02
build: remove barrel files for modules and directives since it breaks metadata collection

arturovt:
**Before:**

image

---

**After:**

image


BioPhoton:
Sweet! 
Nice catch!! Thanks for spending your time here! 🙌

PR #489, opened 2021-02-02
feat(cdk): setup render-strategies and template-management

BioPhoton:


libs/cdk/src/lib/render-strategies/utils.ts
```diff
@@ -0,0 +1,82 @@
...
+          : strategies[defaultStrategyName]
+      )
+    );
+}
+
+export function onStrategy<T>(
+  value: T,
+  strategy: StrategyCredentials,
+  workFactory: (value: T, work: RenderWork, options: { scope?: any }) => void,
+  options: { scope?: any } = {}
```
Karnaukhov-kh, line 74: Should scope be a specific type/enum?

libs/cdk/src/lib/template-management/list-manager.ts
```diff
@@ -0,0 +1,191 @@
...
+  notifyInjectingParentIfNeeded,
+  strategyHandling,
+  TNode,
+} from './utils';
+
+export interface ListManager<T, C> {
+  nextStrategy: (config: string | Observable<string>) => void;
+
+  render(changes$: Observable<NgIterable<T>>): Observable<any>;
+}
```
Karnaukhov-kh, line 36: Shoul be `Observable<C>` instead of `Observable<any>`?

libs/cdk/src/lib/template-management/model.ts
```diff
@@ -0,0 +1,80 @@
...
+interface RxContext {
+  // set context var complete to true (var$; let e = $error)
+  $error: false | Error;
+  // set context var complete to true (var$; let c = $complete)
+  $complete: boolean;
+  // set context var suspense to true (var$; let s = $suspense)
+  $suspense: any;
+}
+
+export interface RxViewContext<T> extends NgViewContext<T>, RxContext {
```
Karnaukhov-kh, line 43: Is it needed?

libs/cdk/src/lib/template-management/template-manager.ts
```diff
@@ -0,0 +1,167 @@
...
+                  const context = getContext[kind](value);
+                  const view = <EmbeddedViewRef<C>>viewContainerRef.get(0);
+                  Object.keys(context).forEach((k) => {
+                    view.context[k] = context[k];
+                  });
+                  work(view, options.scope, notification);
+                }
+                activeTemplate = templateName;
+              }
+              // { scope: viewContainerRef.get(0) }
```
Karnaukhov-kh, line 143: Should be removed?

libs/cdk/src/lib/template-management/utils.ts
```diff
@@ -0,0 +1,536 @@
...
+          detectChanges(parentComponent);
+        },
+        { scope: parentComponent }
+      )
+    );
+  }
+  return behaviors;
+}
+
+export function templateHandling<N, C>() {
```
Karnaukhov-kh, line 292: Return type is missing

libs/cdk/src/lib/template-management/utils.ts
```diff
@@ -0,0 +1,536 @@
...
+        // console.log('notify injectingView', injectingViewCdRef);
+        work(injectingViewCdRef, options.scope);
+      }
+      //  scopeOnInjectingViewContext
+    ).pipe(ignoreElements())
+         : (([] as unknown) as Observable<never>);
+}
+
+// TNode is a component that was projected into another component (virtual parent)
+export function getVirtualParentNotifications$(
```
Karnaukhov-kh, line 265: Return type is missing

libs/cdk/src/lib/template-management/utils.ts
```diff
@@ -0,0 +1,536 @@
...
+      );
+    }
+    return isTemplateRefOrNull;
+  }
+}
+
+export function strategyHandling(
+  defaultStrategyName: string,
+  strategies: StrategyCredentialsMap
+) {
```
Karnaukhov-kh, line 331: Return type is missing

libs/cdk/src/lib/template-management/utils.ts
```diff
@@ -0,0 +1,536 @@
...
+  );
+  return {
+    next(name: string | Observable<string>) {
+      strategyName$.next(name);
+    },
+    strategy$,
+  };
+}
+
+export function templateTriggerHandling<T>() {
```
Karnaukhov-kh, line 347: Return type is missing

libs/cdk/src/lib/template-management/utils.ts
```diff
@@ -0,0 +1,536 @@
...
+  const trigger$ = templateTriggerSubject.pipe(mergeAll());
+  return {
+    next(templateName: Observable<RxNotification<T>>) {
+      templateTriggerSubject.next(templateName);
+    },
+    trigger$,
+  };
+}
+
+export function getHotMerged<U>() {
```
Karnaukhov-kh, line 358: Return type is missing

libs/cdk/src/lib/template-management/utils.ts
```diff
@@ -0,0 +1,536 @@
...
+export function getListTemplateManager<
+  C extends { updateContext: (context: RxListViewComputedContext<T>) => void },
+  T
+>(templateSettings: {
+  viewContainerRef: ViewContainerRef;
+  templateRef: TemplateRef<C>;
+  createViewContext: CreateViewContext<T, C>;
+  updateViewContext: UpdateViewContext<T, C>;
+  patchZone: NgZone | false;
+}) {
```
Karnaukhov-kh, line 379: Return type is missing

libs/cdk/src/lib/render-strategies/utils.ts
BioPhoton: A non-primitiveIMHO. let me check.

libs/cdk/src/lib/template-management/list-manager.ts
BioPhoton: fixed

libs/cdk/src/lib/template-management/model.ts
BioPhoton: no :)
Thanks!

libs/cdk/src/lib/template-management/template-manager.ts
BioPhoton: done

libs/cdk/src/lib/render-strategies/utils.ts
BioPhoton: Done. typed to boject

libs/cdk/src/lib/template-management/utils.ts
BioPhoton: done

libs/cdk/src/lib/template-management/utils.ts
BioPhoton: Done

libs/cdk/src/lib/template-management/utils.ts
BioPhoton: Done

libs/cdk/src/lib/template-management/utils.ts
BioPhoton: Done

libs/cdk/src/lib/template-management/utils.ts
BioPhoton: Done

libs/cdk/src/lib/template-management/utils.ts
BioPhoton: Done

libs/cdk/src/lib/model.ts
```diff
@@ -0,0 +1,3 @@
+export interface CoalescingOptions {
+  scope?: object;
```
hoebbelsB, line 2: I think
```ts
type coalescingObj = Record<string | number | symbol, unknown>;
``` is the way to go?

File: libs/cdk/src/lib/render-strategies/model.ts [hunk 12]
```diff
@@ -0,0 +1,37 @@
...
+import { ChangeDetectorRef } from '@angular/core';
+
+export enum RxNotificationKind {
+  suspense = 'suspense',
+  next = 'next',
+  error = 'error',
+  complete = 'complete'
+}
+
+type NotificationExtract = 'value' | 'hasValue';
```
hoebbelsB, line 11:
```suggestion
type NotificationPick = 'value' | 'hasValue';
```
or maybe `NotificationSlice`

libs/cdk/src/lib/render-strategies/utils.ts
```diff
@@ -0,0 +1,83 @@
...
+  StrategyCredentialsMap,
+  RxNotification,
+  RxNotificationKind,
+  RxErrorNotification,
+  RxCompleteNotification
+} from './model';
+import { CoalescingOptions } from '../model';
+
+
+const rxJsToRxA = {
```
hoebbelsB, line 15: we should turn on `strict mode`

libs/cdk/src/lib/render-strategies/model.ts
```diff
@@ -0,0 +1,37 @@
...
+export type RxNextNotification<T> = Pick<Notification<T>, NotificationExtract> & { kind: RxNotificationKind } & {error: boolean} & {complete: boolean};
+export type RxSuspenseNotification = Pick<Notification<unknown>, NotificationExtract> & { kind: RxNotificationKind.suspense } & {error: false} & {complete: false};
+export type RxErrorNotification = Pick<Notification<unknown>, NotificationExtract> & { kind: RxNotificationKind.error } & {error: any} & {complete: false};
+export type RxCompleteNotification = Pick<Notification<unknown>, NotificationExtract> & { kind: RxNotificationKind.complete } & {complete: boolean} & {error: false};
+export type RxNotification<T> = RxNextNotification<T> | RxSuspenseNotification | RxErrorNotification | RxCompleteNotification;
+
+export type RenderWork = <T = unknown>(
+  cdRef: ChangeDetectorRef,
+  scope?: object,
+  notification?: RxNotification<T>
```
hoebbelsB, line 22: do we provide an operator like `rxMaterialize` ? and do we export it? Otherwise this can get confusing for people when they want to provide their own `StrategyCredentials`

libs/cdk/src/lib/render-strategies/utils.ts
```diff
@@ -0,0 +1,83 @@
...
+  return {
+    ...notification,
+    error: false,
+    complete: false,
+    kind
+  };
+
+};
+
+export function rxMaterialize<T>(): OperatorFunction<T, RxNotification<T>> {
```
hoebbelsB, line 43: ah, `rxMaterialize` :laughing:

libs/cdk/src/lib/template-management/index.ts
```diff
@@ -0,0 +1,4 @@
+export * from './utils';
```
hoebbelsB, line 1: do we really want a `utils` for each sub-folder? maybe we are able to combine all utils into a single utils folder/file. but this does not have to be in the scope of this PR

libs/cdk/src/lib/render-strategies/utils.ts
hoebbelsB:
```suggestion
const rxJsToRxA: Record<'N' | 'E' | 'C', RxNotificationKind> = {
```

https://github.com/ReactiveX/rxjs/blob/master/src/internal/Notification.ts

`NotificationKind` is `@deprecated`

libs/cdk/src/lib/template-management/model.ts
```diff
@@ -0,0 +1,75 @@
...
+  suspense = 'rxSuspense'
+}
+
+export const enum ListChange {
+  insert,
+  remove,
+  move,
+  update,
+  context,
+}
```
hoebbelsB, line 27: https://github.com/ReactiveX/rxjs/blob/master/src/internal/Notification.ts
looking here it says: `NotificationKind is deprecated as const enums are not compatible with isolated modules. Use a string literal instead.`

libs/cdk/src/lib/template-management/model.ts
```diff
@@ -0,0 +1,75 @@
...
+export interface RxListViewComputedContext<T> {
+  index: number;
+  count: number;
+}
+
+export interface RxListViewContext<T extends Record<string | number | symbol, any>, U extends NgIterable<T> = NgIterable<T>, K = keyof T> extends RxListViewComputedContext<T> {
+  $implicit: T;
+  item$: Observable<T>;
+  updateContext(newProps: RxListViewComputedContext<T>): void;
+}
```
hoebbelsB, line 49:
```suggestion
export interface RxListViewContext<T extends Record<string | number | symbol, unknown>, U extends NgIterable<T> = NgIterable<T>, K = keyof T> extends RxListViewComputedContext<T> {
  $implicit: T;
  item$: Observable<T>;
  updateContext(newProps: RxListViewComputedContext<T>): void;
}
```

libs/cdk/src/lib/template-management/model.ts
```diff
@@ -0,0 +1,75 @@
...
+
+export type CreateViewContext<T, C> = (value: T, customProps?: object) => C;
```
hoebbelsB, line 51:
```suggestion
export type CreateViewContext<T, C> = (value: T, customProps?: Partial<C>) => C;
```

libs/cdk/src/lib/template-management/template-manager.ts
```diff
@@ -0,0 +1,168 @@
...
+  strategyHandling,
+  templateHandling,
+  templateTriggerHandling,
+  TNode,
+} from './utils';
+import { CoalescingOptions } from '../model';
+
+export interface RenderAware<T> {
+  nextStrategy: (nextConfig: string | Observable<string>) => void;
+  render: (values$: Observable<T>) => Observable<any>;
```
hoebbelsB, line 36:
```suggestion
  render: (values$: Observable<T>) => Observable<T | false | Error>;
```
hmm any seems kind of fishy... is this one correct?

libs/cdk/src/lib/template-management/template-manager.ts
```diff
@@ -0,0 +1,168 @@
...
+>(config: {
+  renderSettings: RenderSettings<T, C>;
+  templateSettings: TemplateSettings<T, C>;
+  templateTrigger$?: Observable<RxNotificationKind>;
+  notificationToTemplateName: {
+    [rxKind: string]: (
+      value: T,
+      templates: { get: (name: N) => TemplateRef<C> }
+    ) => N;
+  };
```
hoebbelsB, line 61: can we make a `type` out of it?
 suggestion: 
```ts

Record<RxNotificationKind, (
      value: T,
      templates: { get: (name: N) => TemplateRef<C> }
    ) => N>
```

libs/cdk/src/lib/template-management/template-manager.ts
```diff
@@ -0,0 +1,168 @@
...
+  return {
+    addTemplateRef: templates.add,
+    addTrigger: triggerHandling.next,
+    nextStrategy: strategyHandling$.next,
+    render(values$: Observable<T>): Observable<any> {
+      return values$.pipe(
+        map((o) => (isObservable(o) ? o : of(o))),
+        distinctUntilChanged(),
+        switchAll(),
+        distinctUntilChanged(),
```
hoebbelsB, line 105:
```suggestion
        ngInputFlatten(),
```

libs/cdk/src/lib/template-management/template-manager.ts
```diff
@@ -0,0 +1,168 @@
...
+                  const context = getContext[kind](value);
+                  const view = <EmbeddedViewRef<C>>viewContainerRef.get(0);
+                  Object.keys(context).forEach((k) => {
+                    view.context[k] = context[k];
+                  });
+                  work(view, options.scope, notification);
+                }
+                activeTemplate = templateName;
+              }
+              // @TODO specify scoping
```
hoebbelsB, line 144: so did we specify it? :D

hoebbelsB (changes requested):
ty! have some questions regarding typings.

And we should discuss a folder structure in a more streamlined way, we currently have 100 different `utils` folders... :laughing:

libs/cdk/src/lib/render-strategies/model.ts
BioPhoton: Yes, IMHO we do. 
This is used for triggers. But ATM we only use this in `rxLet`. Potentially we can also use it in `rxContext`?

libs/cdk/src/lib/render-strategies/utils.ts
BioPhoton: Docs are missing

libs/cdk/src/lib/template-management/index.ts
BioPhoton: The idea for utils per subfolder was to organize and share folder internal (not exported) functions across a folder.
Functions used in multiple folders would put in your suggested public utils. I refrained to to this so far because introducing it later leads to more thoughts on what we really need to export.

libs/cdk/src/lib/template-management/model.ts
BioPhoton: If we use it only folder internal it should not matter IMHO

libs/cdk/src/lib/template-management/template-manager.ts
BioPhoton: Done

libs/cdk/src/lib/template-management/template-manager.ts
BioPhoton: Its now called `coerceAndSwitchDistinct` :D

BioPhoton: done

libs/cdk/src/lib/template-management/template-manager.ts
BioPhoton: If there is no scoping I would like to comment why.

libs/cdk/src/lib/template-management/template-manager.ts
BioPhoton: What this reminds me of is we are inconsistent across `TemplateManager` and `ListManager` in the render function output.

TemplateManager:
```typescript
export interface RenderAware<T> {
  nextStrategy: (nextConfig: string | Observable<string>) => void;
  render: (values$: Observable<T>) => Observable<any>;
}

export interface TemplateManager<
  T,
  C extends RxViewContext<T>,
  N = RxBaseTemplateNames | string
> extends RenderAware<T> {
  addTemplateRef: (name: N, templateRef: TemplateRef<C>) => void;
  addTrigger: (trigger$: Observable<RxNotification<T>>) => void;
}
```
vs
```typescript
export interface ListManager<T, C> {
  nextStrategy: (config: string | Observable<string>) => void;
  render(changes$: Observable<NgIterable<T>>): Observable<C>;
}
```

I like the version where we return the context. It contains all information possible:
- RxContext => suspense, next, error, complete
- additional context variables depending on the directive
-

libs/cdk/src/lib/template-management/model.ts
```diff
@@ -0,0 +1,79 @@
...
+import { Observable } from 'rxjs';
+import { StrategyCredentialsMap } from '../render-strategies/model';
+
+export type rxBaseTemplateNames = 'rxError' | 'rxComplete' | 'rxSuspense';
+
+export enum RxBaseTemplateNames {
+  error = 'rxError',
+  complete = 'rxComplete',
+  suspense = 'rxSuspense'
+}
```
BioPhoton, line 19: The naming could map with the input binding of directives? 
`*rxLet="errorTpl: myTpl"`

```typescript
export enum RxBaseTemplateNames {
  error = 'errorTpl',
  complete = 'completeTpl',
  suspense = 'suspenseTpl'
}
```

libs/cdk/src/lib/template-management/template-manager.ts
hoebbelsB: but `render(changes$: Observable<NgIterable<T>>): Observable<C>;` makes no sense at all.
`<C>` is the `RxViewContext` for a single item, but the renderCallback of `rxFor, (ListManager)` should return all items applied to the view? So just `NgIterable<T>` 

Otherwise we could think about emitting `<void>` as `renderCallback` as a starting point. I have used them a lot and never found myself in a situation where i needed the informationen passed via the callback.

libs/cdk/src/lib/template-management/template-manager.ts
hoebbelsB: the problem here is that the template-manager potentially has multiple scopes (1 for each `EmbeddedViewRef` and 1 for the `Component` itself).

in a perfect world we would scope each operation to the template `EmbeddedViewRef` when possible. When creating the first `EmbeddedViewRef` it won't be possible, since the scope gets created in the function which we would like to scope with it... ? :D

I don't even know if we really need to scope to a single `EmbeddedView`

libs/cdk/src/lib/template-management/model.ts
hoebbelsB: will every directive have it's own template though?

libs/cdk/src/lib/template-management/template-manager.ts
BioPhoton: > <C> is the RxViewContext for a single item, but the renderCallback of rxFor
argh. right! Lets void here. agree

BioPhoton (commented):
Wrote some thought on public operators

libs/cdk/src/lib/model.ts
BioPhoton: done

libs/cdk/src/lib/render-strategies/model.ts
BioPhoton: `rxMaterialize` is moved in lib root utils

libs/cdk/src/lib/template-management/model.ts
BioPhoton: Not always imho, but its good to have a 'link' in the names

libs/cdk/src/lib/template-management/utils.ts
```diff
@@ -0,0 +1,735 @@
...
+export function coerceDistinctWith<T>(flattenOperator?: OperatorFunction<ObservableInput<T>, T>) {
+  flattenOperator = flattenOperator || switchAll();
+  return (o$: Observable<Observable<T> | T>) =>
+    o$.pipe(
+      coerceObservableWith(),
+      distinctUntilChanged(),
+      flattenOperator,
+      distinctUntilChanged()
+    );
+}
```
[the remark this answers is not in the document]
BioPhoton: Moved files.

libs/cdk/src/lib/template-management/template-manager.ts
BioPhoton: In `rxLet` there could e multiple sync emissions from some global state or other sources.

libs/cdk/src/lib/template-management/index.ts
BioPhoton: To elaborate a bit more, here my idea on how we could start to organize dependencies:

if we organize CDK in sub-packages we have the following folder structure:

```terminal
cdk/src/lib/
├── sub-package1/
│   ├── index.ts
│   ├── model (single file or folder)
│   ├── utils (single file or folder)
│   └── feature (single file or folder)
├── sub-package2/
│   └── ...
└── core/
    ├── index.ts
    ├── model (single file or folder)
    └── utils (single file or folder)
```

- The `index.ts` in every folder defined the public exports and hides domain internal code away.
- The `core` folder contains abstract shared code.
- The `sub-package` folders define a specific domain e.g. `template-management` and contain domain dedicated code. 
- In some cases, we will reuse sub-packages in others. Here we should try to have as minimal relations as possible.

libs/cdk/src/lib/template-management/model.ts
BioPhoton: I saw u changed it. close this here...

libs/cdk/src/lib/template-management/template-manager.ts
BioPhoton: So could you add a comment or todo so we don't forget to polish here?

libs/cdk/src/lib/template-management/index.ts
mitschmidt: The folder structure looks good to me so far. For code that is shared across `sub-package1, ... n` one might consider `common` instead of `core`? `core` suggests that it is required by all other subfolders, or might be required for the entire library to work. `common` sounds more like that code **may** but does not have to be used by other parts. What do you think?

BioPhoton: Very good! Yes.
Angular cdk e. G. Has a coerce folder. Which for sure is a common like thing.

libs/cdk/src/lib/template-management/template-manager.ts
BioPhoton: @hoebbelsB u can also put it here and I move it into code

PR #491, opened 2021-02-02
feat(cdk): add event-manager

BioPhoton:
Not sure if this is worth maintaining or if we can just add it as a snippet. WDYS?

### DIFF
--- a/libs/cdk/src/lib/event-manager/index.ts
+++ b/libs/cdk/src/lib/event-manager/index.ts
@@ -1,1 +1,1 @@
-export { ShareEventManager } from './shared-event-manager';
+export { SharedEventManager } from './shared-event-manager';
--- a/libs/cdk/src/lib/event-manager/index.ts
+++ b/libs/cdk/src/lib/event-manager/index.ts
@@ -1,0 +1,1 @@
+export { ShareEventManager } from './shared-event-manager';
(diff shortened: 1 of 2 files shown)


libs/cdk/src/lib/event-manager/shared-event-manager.ts
```diff
@@ -0,0 +1,51 @@
...
+type EventName = string;
+type SharedEvent = Observable<Event> | null | undefined;
+
+interface SharedEventsContainer {
+  window: Map<EventName, SharedEvent>;
+  document: Map<EventName, SharedEvent>;
+}
+
+@Injectable()
+export class ShareEventManager extends EventManager {
```
Karnaukhov-kh, line 18:
```suggestion
export class SharedEventManager extends EventManager {
```

libs/cdk/src/lib/event-manager/index.ts
```diff
@@ -0,0 +1 @@
+export { ShareEventManager } from './shared-event-manager';
```
Karnaukhov-kh, line 1:
```suggestion
export { SharedEventManager } from './shared-event-manager';
```

Karnaukhov-kh (commented):
Can we make it "experimental" and see how it goes? The idea is pretty nice and adds up to the zone-agnostic mental model.

BioPhoton:
I close here and reopen in new PR

PR #504, opened 2021-02-10
chore(build): remove deep import

arturovt:
> Warning: Entry point '@rx-angular/template' contains deep imports into 'node_modules/rxjs/internal/types'. This is probably not a problem, but may cause the compilation of entry points to be out of order.


BioPhoton:
THX!!!

PR #533, opened 2021-02-20
fix(template): reduce public API

BioPhoton:


hoebbelsB:
I would propose to only export:

* `RxPushModule`
* `RxLetModule`
* `RxForModule`
* `RxSwitchModule`
* `RxContextModule`

Imo `UnpatchEventsModule` should be part of `@rx-angular/cdk` since it has nothing to do with the view anyway.

What about `ViewportPrioModule`? Do we want to keep it exported for now?

BioPhoton:
Unpatch is a directive used in the template. Imho definitely in template

hoebbelsB:
okay, do we have other opinions about this?
Thing is, everything else `NgZone` or `unpatching` related is part of `cdk`.

The definition "is used in the view, thus it is template" is not a strong argument imo. If you take a look at other `cdk` libraries (angular/cdk, taiga/cdk) there are lots of `Directives` and `Pipes`.

Karnaukhov-kh:
Very interesting topic.

BTW. Do we handle `unpatch` as experimental still? 😅

IMHO. I see 3 possible points of view here:

1. Everything that we provide in the `template` is observable-related (structural directives and pipes). So everything that is more like building blocks and additional optimization things could be part of `CDK`.
2. In `template`, we provide things that are focusing on the performance of the views. So `unpatch` and `viewportPrio` in the future could be part of `template`.
3. In `template` we provide counterparts of existing Angular features.

I'm more leaning towards the point that `template` should contain direct counterparts of existing Angular features and be partially a drop-in replacement for "CommonModule". Other cool stuff users can consume from `CDK`.

P.S. will strategies be part of `CDK`?

BioPhoton:
>I'm more leaning towards the point that template should contain direct counterparts of existing Angular features and be partially a drop-in replacement for "CommonModule". Other cool stuff users can consume from CDK.

I like this the most

Issue #552, opened 2021-02-23
Fails to build with ngcc in vscode with angular language service


lincolnthree:
# Description

```
[ng] Build at: 2021-02-23T05:27:07.304Z - Hash: 4d037ff176977e7829cf - Time: 2850ms
[ng] Error: ./node_modules/@rx-angular/template/__ivy_ngcc__/fesm2015/rx-angular-template.js
[ng] Module build failed (from ./node_modules/@ngtools/webpack/src/ivy/index.js):
[ng] Error: ENOENT: no such file or directory, open '/Users/lincoln/projects/maritlabs/topdecked-unified/node_modules/@rx-angular/template/__ivy_ngcc__/fesm2015/rx-angular-template.js'
```

# Steps to Reproduce the Issue

1. Download example
2. cd example
3. npm i
4. open in vscode with language service

[example.zip](https://github.com/rx-angular/rx-angular/files/6029475/example.zip)

# Environment

# Related to Other Issues

Related to: https://github.com/angular/vscode-ng-language-service/issues/1144

# Tasks to Resolve This

Edit package.json to point to a `main:` source file that exists within the rx-angular distribution.

# Notes

They claim the issues is that package.json points to a file that does not exist:

https://github.com/angular/vscode-ng-language-service/issues/1144

lincolnthree:
https://github.com/angular/vscode-ng-language-service/issues/1144#issuecomment-784414486

hoebbelsB:
ty!

hoebbelsB:
i've just created a new project with the angular cli and added `@rx-angular/template` via `ng add @rx-angular/template`.

This is the outcome of node_modules:

![image](https://user-images.githubusercontent.com/4904455/108897334-455ccd80-7616-11eb-86e4-1412da997a23.png)

The file which is missing for you is actually here. A build with ivy is working as well. Can you provide some more information please? I will check out your project now

hoebbelsB:
@lincolnthree I have checked out your project and it builds fine? btw. next time please share a git repository... I actually wonder why I just trusted your zip file :O

hoebbelsB:
sorry, I obviously didn't catch that this is an issue with the language service itself. I've tested it with webstorm. @Karnaukhov-kh has some more insights about this!

Karnaukhov-kh:
Hi @lincolnthree. I'm not having any issues with build itself (in new project or your example).

[...]

PR #553, opened 2021-02-23
fix: ng-packagr builds

hoebbelsB:
As pointed out [here](https://github.com/rx-angular/rx-angular/issues/552) and [here](https://github.com/angular/angular/issues/40965) we have some issues with our builds. 

I think the stackblitz issue is connected as well.

image

After hours of trial and error I finally thought about testing `ng-packagr` without the `nrwl` plugin... So I basically just swapped the build command to `npx ng-packagr ...`. It turns out that the missing (`.umd.js`) are now generated correctly.

I will test this with vs-code and npm link tomorrow morning.


lincolnthree:
Yay!

hoebbelsB:
@BioPhoton @Karnaukhov-kh 
image

*rxLet is working with new vs-code-language-service !!!!

lincolnthree:
This looks good! Fix confirmed. However... there is another issue that seems to be undocumented. Not sure if it's caused by this or just by... other updates since the last version we've been able to use:

```
[ng] An unhandled exception occurred: The target entry-point "@rx-angular/template" has missing dependencies:
[ng]  - @rx-angular/cdk
```

Looks like we need to install `@rx-angular/cdk` manually as a dependency to get things working (additionally, version numbers don't match up on those libraries, which could confuse some folks)

PR #560, opened 2021-02-26
feat(push): implement proper patchZone flag

hoebbelsB:
`patchZone` wasn't implemented correctly. fixed that :)

I did not add it as a third argument (` v$ | push: 'strategy'; renderCallback; patchZone`) since it's kind of useless in this scenario. We should think about having the push pipe taking an object as argument, or some kind of tuple type.
The current solution is a pain if you want to configure something in the view.


hoebbelsB:
i know why the build is complaining. i make a separate pr for this

libs/template/src/lib/push/push.pipe.ts
```diff
@@ -79,34 +83,13 @@ export class PushPipe<U> implements PipeTransform, OnDestroy {
```
Karnaukhov-kh, line 83:
```suggestion
/** @internal */
```

libs/template/src/lib/push/push.pipe.ts
```diff
@@ -79,34 +83,13 @@ export class PushPipe<U> implements PipeTransform, OnDestroy {
...
-              scope,
-              strategy: strategy.name,
-              patchZone: this.strategyProvider.config.patchZone ? null : false,
-            }
-          )
-        )
-      )
-      .subscribe();
+    this.subscription = this.handleChangeDetection();
   }
```
Karnaukhov-kh, line 94:
```suggestion
/** @internal */
```

libs/template/src/lib/push/push.pipe.ts
```diff
@@ -140,4 +123,38 @@ export class PushPipe<U> implements PipeTransform, OnDestroy {
   ngOnDestroy(): void {
     this.subscription.unsubscribe();
   }
+
```
Karnaukhov-kh, line 126:
```suggestion
/** @internal */
```

Karnaukhov-kh (commented):
Thought about code docs vs public docs and realised that we do public docs a bit differently. Ignore that 😄.

hoebbelsB:
@BioPhoton @Karnaukhov-kh the build is crashing because only template gets build. It is missing `@rx-angular/cdk` while building though. It relies on that `@rx-angular/cdk` gets build first.
We can fix that either by configuring an `implicitDependency` in `nx.json` or by extending the build command of `@rx-angular/template` with `--with-deps`. 
What is the preferred way of doing so?

BioPhoton:
--with-depth

hoebbelsB:
> --with-depth

this will mean that changes in `@rx-angular/cdk` would make the github actions build it twice. This is not a big deal but it will result in more time consumption for the github actions

BioPhoton:
Hi @hoebbelsB, let me know if there is some progress.

hoebbelsB:
close in favor of #575

PR #561, opened 2021-02-26
feat(cdk): improve template manager view-context

hoebbelsB:
# Description

When an observable bound to an `rxLet` (TemplateManager) `completes` or `errors` it is not possible to display the latest value in the provided templates.
I have improved the `RxTemplateManager` and `templateNotifier` so that the `RxViewContext` will always contain the latest emitted value of the bound value.

A little example:
rx-let-viewcontext

Example code:

```html
<ng-template #error let-value let-error="$error">
      <div>
        <mat-icon class="error-icon">thumb_down</mat-icon>
        <h2>{{ error }}</h2>
        <strong>Last valid value: {{ value }}</strong>
      </div>
    </ng-template>
```

```html
<ng-template #complete let-value>
      <div>
        <mat-icon class="notification-icon complete-icon">thumb_up</mat-icon>
        <h2>Completed!</h2>
        <strong>Last valid value: {{ value }}</strong>
      </div>
    </ng-template>
```

Additionally we should talk about the naming of the variables inside the `RxViewContext`. I don't know why prefix all variables with `$`? (`let-error="$error"`) It is only good for `$implicit` since this is a reserved key of angular. All other variables should not contain this imo. @BioPhoton @Karnaukhov-kh . Any suggestions?

### DIFF
--- a/apps/demos/src/app/features/template/rx-let/basic/rx-let-basic.component.ts
+++ b/apps/demos/src/app/features/template/rx-let/basic/rx-let-basic.component.ts
@@ -1,5 +1,5 @@
 import { ChangeDetectionStrategy, Component } from '@angular/core';
-import { of, Subject } from 'rxjs';
+import { Subject } from 'rxjs';
 import { map } from 'rxjs/operators';
 
 @Component({
--- a/apps/demos/src/app/features/template/rx-let/basic/rx-let-basic.component.ts
+++ b/apps/demos/src/app/features/template/rx-let/basic/rx-let-basic.component.ts
@@ -1,5 +1,5 @@
 import { ChangeDetectionStrategy, Component } from '@angular/core';
-import { Subject } from 'rxjs';
+import { of, Subject } from 'rxjs';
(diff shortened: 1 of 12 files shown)


Karnaukhov-kh:
Mixed feelings. People are used to $implicit so maybe it will be easier for users to get used to $error, $complete and $suspense. And also complete and error are pretty common words, so with $ it might be easier to spot error and complete that you need.

apps/demos/src/app/features/template/rx-let/basic/rx-let-basic.component.ts
```diff
@@ -1,13 +1,13 @@
 import { ChangeDetectionStrategy, Component } from '@angular/core';
-import { Subject } from 'rxjs';
+import { of, Subject } from 'rxjs';
```
BioPhoton, line 2: remove unused import

BioPhoton (approved):
Nice! Like that ew got rid of the templateObserver side effect! 🦾

PR #567, opened 2021-02-27
Template manager tests

Karnaukhov-kh:


hoebbelsB:
@Karnaukhov-kh the build is failing for the same reason mentioned in #560 .
And shouldn't we move the `RxTemplateManager` tests to the `cdk` package?

Karnaukhov-kh:
@hoebbelsB, yes, same reason. Moving make sense, just tried to make them launch with more or less proper fixture. No real tests here, 90% of old tests were referencing api that doesn't exist anymore. Believe that Michael might be working on cdk tests right now, so this draft is just a signal that this spec file can run now.

PR #574, opened 2021-03-02
fix(build): make sure cdk gets build when template gets build

hoebbelsB:
set `@rx-angular/template` as `implicitDependency` for `@rx-angular/cdk`. This ensures `cdk` gets build whenever `template` gets build. This way we ensure `template` has a built version of `cdk` whenever it needs

### DIFF
--- a/angular.json
+++ b/angular.json
@@ -718,7 +718,7 @@
                 "command": "rimraf dist/libs/template"
               },
               {
-                "command": "ng run template:build-lib"
+                "command": "ng run template:build-lib --with-deps"
               },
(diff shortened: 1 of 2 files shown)


hoebbelsB:
okay, so this will end up in circular dependencies... so it will be `--with-deps` even if this means we sometimes build cdk twice?

hoebbelsB:
inserted `--with-deps` for the `@rx-angular/template` build. otherwise it will just be annoying to deal with. Maybe we find some build specialist at some point in time :)

PR #582, opened 2021-03-06
feat(cdk): add accumulateObservable creation function

BioPhoton:
# accumulateObservable 

- [x] implement operator
- [x] create tests
- [x] provode code docs

This new operator solves the following problems:

- Composition of observables to a state slice is often a wordy block of code
```typescript
const object$: Observable<{
   prop1: number,
   prop2: string,
   prop3: string
 }> = combineLatest([
    interval(42).pipe(filter(v => v !== undefined), distinctUntilChanged()),
    of('lorem').pipe(filter(v => v !== undefined), distinctUntilChanged()), 
    of(true).pipe(filter(v => v !== undefined), distinctUntilChanged())
]).pipe(
  map(([prop1, prop2, prop3]) => ({prop1, prop2, prop3})),
  shareReplay(1)
);
```
- Initializing `RxState` from multiple Observables is also wordy
```typescript
rxState.connect('prop1', interval(42));
rxState.connect('prop1', of('lorem'));
rxState.connect('prop1', of(true));
```
- Mixing of static and observable values is a mess
```typescript
rxState.connect('prop1', interval(42));
rxState.set({prop1:'lorem'});
rxState.connect('prop1', of(true));
```

# Solution 

A helper function abstracting the repetitive code.

Things the helper should take care of:

- filter out undefined values to avoid nullish check later 
- filter out emissions for unchanged values
- coalesce over emissions on a promise
- shares the composition
- replays latest emission

Composition of observables
```typescript
const object$ = accumulateObservables({
    prop1: interval(42),
    prop2: of('lorem'), 
    prop3: of(true)
});
```
Initializing `RxState`
```typescript
rxState.connect(accumulateObservables({
    prop1: interval(42),
    prop2: of('lorem'), 
    prop3: of(true)
}));
```

## Unsolved problems

The composition, even if shared is "cold" meaning controlled by the consumer. This can lead to unwanted behavior, especially within templates that get recreated e.g. `*ngIf`.

As this problem is closely related to the place where it is used the user should take care of controlling the subscriptions.
`RxState` for example but also any other state management helper can just take that over.

# Open Todos

- Mixing of static and observable values
```typescript
rxState.connect(accumulateObservables({
    prop1: interval(42),
    prop2: 'lorem', 
    prop3: of(true)
}));
```


hoebbelsB:
@BioPhoton made it green  :dancers:

Issue #584, opened 2021-03-07
Error in console if using Global state + nested objects in template


cakeinpanic:
# Description

Hi, I have noticed that if using global state, it provides `undefined` at first render, and if template uses nested fields – it causes error in console
```
export const initialState: GlobalState= {
  hello: 'Hello!'
}
constructor(@Inject(GLOBAL_RX_STATE) private state: RxState<GlobalState>){
     this.state.set(initialState);
 }

///

readonly state$: Observable<GlobalState> = this.state.select();
constructor(@Inject(GLOBAL_RX_STATE) private state: RxState<GlobalState>) {}

////

<div *rxLet="state$; let state; rxSuspense: suspenseTemplate">
{{state.hello}}
</div>
```
<img width="596" alt="Screen Shot 2021-03-07 at 21 27 46" src="https://user-images.githubusercontent.com/588916/110252461-32d78200-7f8e-11eb-9ad6-e89ae5863879.png">

# Steps to Reproduce the Issue

Here is [stackblitz](https://stackblitz.com/edit/rx-angular-state-global-issue)

# Environment
```
    "@angular/common": "~11.1.0",
    "@angular/compiler": "~11.1.0",
    "@angular/core": "~11.1.0",
    "@angular/platform-browser": "~11.1.0",
    "@angular/platform-browser-dynamic": "~11.1.0",
    "@rx-angular/cdk": ">=1.0.0-alpha.2",
    "@rx-angular/state": "^1.4.3",
    "@rx-angular/template": "^1.0.0-beta.24",
```

# Notes
I have debugged issue a bit and this patch fixes this exact problem. Dunno if it breaks anything else :)
related to file https://github.com/rx-angular/rx-angular/blob/master/libs/cdk/src/lib/template-management/template-manager.ts#L185
```
Index: libs/cdk/src/lib/template-management/template-manager.ts
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/libs/cdk/src/lib/template-management/template-manager.ts b/libs/cdk/src/lib/template-management/template-manager.ts
--- a/libs/cdk/src/lib/template-management/template-manager.ts	(revision f83788e47a5933a248d3d84c9e174190d5a50407)
+++ b/libs/cdk/src/lib/template-management/template-manager.ts	(date 1615049309031)
@@ -183,7 +183,8 @@
                     viewContainerRef.clear();
                   }
[...]

PR #586, opened 2021-03-08
fix bug in first template init

cakeinpanic:
fix for https://github.com/rx-angular/rx-angular/issues/584


hoebbelsB:
thank you katya!!! :+1:  this should work just fine. I have no idea why the pipeline is failing though.

hoebbelsB (approved):
LGTM

cakeinpanic:
I checked, pipeline fails cause of bug in labeler action: https://github.com/actions/labeler/issues/36

PR #587, opened 2021-03-08
Fix template schematics

edbzn:
Aims to fix template schematics.


hoebbelsB:
THX @edbzn !!!

so now we actually reverted everything back to the point where we build everything from the libs, not from the node_modules folder? Are we still sure `@rx-angular/cdk` does not get bundled with `@rx-angular/template`? 
We need to make sure that cdk is actually a peerDependency of template. I am not an expert at CI stuff, maybe u can enlighten me a bit about what u did here? :)

hoebbelsB:
btw. this is what @LayZeeDK says to this topic:

image

thats the link: https://github.com/rx-angular/rx-angular/blob/zone-patch-push-pipe/.github/workflows/build-and-test.yml#L42

edbzn:
@hoebbelsB Hi, to fix the build command I kept the --with-deps option in order to build the cdk package before the template. I fixed the tsconfig files to extend the base config as it's required to Nx to work properly. I also added a path alias to point to the cdk dist folder to make ng-packagr happy with the build step. As far as I know it's also required to make one punishable library dependent from another one. I also kept the cdk as a peerDependency of the template package.

hoebbelsB:
ok I need to check if anything is different. Just so you know, the last setup was basically copy/pasted by the ngrx-guys. All of those things you pointed out fixing were intentionally configured like this.
Either way, if your setup works I am more than happy to have someone on board who has a knowledge on this topic :100:

edbzn:
That's how we made it for one of our project, we had the exact same scenario. I don't know the previous solution with the copy in the node_modules folder and it did not worked on the master branch so I reverted with my solution. Also I was confused about the rollup config, why does it is here since we are using Angular CLI builders?

hoebbelsB:
I was confused as well :facepalm: I guess this was one of the things we copied from ngrx

hoebbelsB:
hi @edbzn I have checked out your branch, built all projects and tried the artefacts in other projects. 
I tried with `npm link` and with plain copy/pasting but I didn't manage to get anything work.

This is the outcome of a build with `IvyEnabled: false`: 

`
Error: ./src/app/app.component.ngfactory.js 20:160-175
"export 'LetDirective' (imported as 'i2') was not found in '@rx-angular/template'

Error: ./src/app/app.module.ngfactory.js 14:2567-2579
"export 'LetModule' (imported as 'i6') was not found in '@rx-angular/template'

Error: ./src/app/app.module.ngfactory.js 14:2581-2593
"export 'LetModule' (imported as 'i6') was not found in '@rx-angular/template'
`

This is the result with `Ivy` enabled:
`
Error: node_modules/@rx-angular/template/lib/let/let.module.d.ts:1:22 - error NG6002: Appears in the NgModule.imports of AppModule, but could not be resolved to an NgModule class.

This likely means that the library (@rx-angular/template) which declares LetModule has not been processed correctly by ngcc, or is not compatible with Angular Ivy. Check if a newer version of the library is available, and update if so. Also consider checking with the library's authors to see if the library is expected to be compatible with Ivy.

1 export declare class LetModule {
                       ~~~~~~~~~


Error: ./src/app/app.module.ts 17:12-21
"export 'LetModule' was not found in '@rx-angular/template'
`

edbzn:
@hoebbelsB Thanks for your time, I will check today but at first look we should not publish libraries with Ivy enabled, maybe it's a problem with npm link which does not trigger ngcc post install hook, in any case I need to dig into.

hoebbelsB:
I didn't build the lib with Ivy, just the other project. And as I said: I tried copy/paste and npm link. Both approaches work without your changes. @arturovt can actually confirm this behavior

edbzn:
@hoebbelsB How can it works if the build does not work on master, are you testing from a specified commit? It would help me to be able to compare.

hoebbelsB:
the build does work if you specify everything yourself. It was just the dependency which was not working. so when template has a change, it needs cdk to be built aswell. That is (besides the confusing setup) the only thing not working atm. If you do:

`yarn nx run cdk:build`
`yarn nx run template:build`

everything should be just fine

edbzn:
@hoebbelsB I just fixed the problem, I tested with and without Ivy in a sandbox app and it works, I cleaned the template and cdk tsconfig files to be in sync with what Nx originally generates so we will have a standard Nx config.

FYI even if we build the cdk first, the master build for the template library was not working for several reasons, the `ng ... --with-deps` throws because it's an Nx option, also I get an error with the `npc` package, but that's not a problem anymore.

I inspected some CI logs, for exemple [this one](https://github.com/rx-angular/rx-angular/runs/2077100786) or [this one](https://github.com/rx-angular/rx-angular/runs/2059870346) (and many others) we can see that it didn't triggered any build or test even if some projects were affected by the commit. It seems that you didn't get notified that something was wrong.

The problem comes from the CI configuration, when the CI get triggered by a master push directly, **not from pull requests**, the affected commands run with an incorrect `head` and `base` comparison, I think `{{ github.event.pull_request.head.ref }}` is not what you need. 

PS: happy to see the labeler in green :heart_eyes:

hoebbelsB:
@edbzn sorry was very busy the last days! Thx for further investigations :100: ! I will try to have a look here and at #597 over the weekend

hoebbelsB:
@edbzn sorry I had no time the last couple of days. Now that #597 is merged. Is it still advisable to introduce your changes?

edbzn:
@hoebbelsB Yes this PR primarily fixes the template schematic, the branch was rebased on what was merged so it just cleans up some minor details.

libs/template/ng-package.json
```diff
@@ -5,7 +5,7 @@
     "flatModuleFile": "template",
     "entryFile": "src/index.ts",
     "umdModuleIds": {
-      "@rx-angular/cdk": "rx-angular.cdk"
+      "@rx-angular/cdk": "@rx-angular/cdk"
```
hoebbelsB, line 8: i remember @arturovt telling me that the prevous `dot notation` was actually correct. I need to test this out

hoebbelsB:
ok @edbzn so now this PR only focuses on schematics. very cool :+1:  could u check the changes to `ng-package.json` again?

libs/template/ng-package.json
edbzn: I don't get the point of using another name here

edbzn:
After rebasing I'm facing a new issue, the `build` target was modified, it causes a side effect because previously a dedicated `build-schematics` target was run:

```json
"build-schematics": {
  "builder": "@nrwl/workspace:run-commands",
  "options": {
    "commands": [
      {
        "command": "npx tsc -p libs/template/tsconfig.schematics.json"
      },
      {
        "command": "npx cpx libs/template/schematics/collection.json dist/libs/template/schematics"
      },
      {
        "command": "npx cpx libs/template/schematics/src/**/files/** dist/libs/template/schematics/src"
      },
      {
        "command": "npx cpx libs/template/schematics/src/**/schema.json dist/libs/template/schematics/src"
      }
    ],
    "parallel": false
  }
},
```

This code is not triggered anymore when building the template, it causes an issue because ng-packagr is guessing `template/schematics` to be a secondary endpoint and it tries to build it using the `@nrwl/angular:package`, here are some warnings: 

```
WARNING: No name was provided for external module '@angular-devkit/schematics' in output.globals – guessing 'schematics'
WARNING: No name was provided for external module '@angular-devkit/schematics/tasks' in output.globals – guessing 'tasks'
WARNING: No name was provided for external module '@schematics/angular/utility/dependencies' in output.globals – guessing 'dependencies$1'
WARNING: No name was provided for external module 'http' in output.globals – guessing 'http'
WARNING: Creating a browser bundle that depends on Node.js built-in module ('http'). You might need to include https://github.com/ionic-team/rollup-plugin-node-polyfills
```

The schematics should be bundled using the TypeScript compiler, not ng-packagr, @arturovt any clue on how I can fix it without bringing back the old build target?

libs/template/ng-package.json
hoebbelsB: I have tested the build in another project and it seems to work just fine without any warnings. I will do some more testing

hoebbelsB:
@edbzn did u find out anything about the WARNINGS you receive?

libs/template/ng-package.json
arturovt: @hoebbelsB it will not show any warning during the build since UMD is a runtime:
1. you can check the `.umd.js` and it should show you the factory definition with `globals.umdId` here
2. try to create a simple project and add `@rx-angular/cdk` to webpack externals and try to import the UMD file through SystemJS for example (this is an example of sharing dependencies in micro-frontends)

As far as I can see in the latest version the UMD id is `rx-angular.cdk` (https://unpkg.com/@rx-angular/cdk@1.0.0-alpha.6/bundles/rx-angular-cdk.umd.js): `factory((global['rx-angular'] = global['rx-angular'] || {}, global['rx-angular'].cdk = {})`

edbzn:
I think it's because other projects don't import any node dependencies, I will take time to solve it soon, it's been open for too long.

libs/template/ng-package.json
hoebbelsB: thx @arturovt

edbzn: I will bring this back, thanks, @arturovt can you check [this comment](https://github.com/rx-angular/rx-angular/pull/587#issuecomment-802868720), please?

edbzn:
It's fixed, I had to revert some changes made in here https://github.com/rx-angular/rx-angular/commit/7961b2fad045cba0f80f222745715c75bbabbfc3 because it broke the schematic build. I tested using a link and running `ng g @rx-angular/template:ng-add`. Also, the doc should be updated, there's no --project option.

hoebbelsB:
@BioPhoton what about this one here? Can we sync up again?

BioPhoton:
Yeah way behind! Catch up with you all next week. (even if I'm of little help here 😅)

arturovt:
Hey guys. I have actually come up with a question. Why does `template/schematics` have a `package.json` file with an `ngPackage` property if this shouldn't be built through the `ng-packagr`?

The default `ng-packagr` lookup algorithm for secondary entrypoints is to glob all folders with `package.json` or `ng-package.json` files:
```ts
glob(`${packagePath}/**/{ng-package,package}.json`);
```

We'd want to remove that `package.json` (or at least this is my subjective opinion, but I don't really know why it's there).

Another question would be related to naming targets. As I said previously in my PR that Nx doesn't work like that and they're searching for the `build` and `prepare` targets (I'm not sure if we have to break Nx's standards):
```ts
export function getOutputsForTargetAndConfiguration() {
  if (options.outputPath) {
    return Array.isArray(options.outputPath)
      ? options.outputPath
      : [options.outputPath];
  } else if (target === 'build' || target === 'prepare') {
    return [
      `dist/${node.data.root}`,
      `${node.data.root}/dist`,
      `${node.data.root}/build`,
      `${node.data.root}/public`,
    ];
  } else {
    return [];
  }
}
```

Thus naming the target as `build-lib` will fail `--with-deps` to fail since packages are not able to discover other package they rely on.

@edbzn could you try to name the target as `prepare` and other target as `build` (which will run the whole workflow with building schematics)? I haven't tried the `prepare` target but I think it's worth giving a try.

This might be an issue in the future when we start splitting libs into smaller libs and that will cause troubles with builds for different packages since we'll end up doing some stuff "for the sake of fixing an issue" (e.g. adding TS paths to different TS configs, etc.).

edbzn:
Heya @arturovt, 

> I have actually come up with a question. Why does template/schematics have a package.json file with an ngPackage property if this shouldn't be built through the ng-packagr?

I also spotted this and it's fixed in this PR, now it executes a traditional tsc build.

> @edbzn could you try to name the target as prepare and other target as build (which will run the whole workflow with building schematics)? I haven't tried the prepare target but I think it's worth giving a try.

Nice suggestion!! Replacing the target name `build-lib` to `prepare` allows me to remove the path alias without breaking the build. Thanks!

edbzn:
Ok ... maybe I spoke too quickly, the target naming solution is not working. The following target is breaking the build : 

```json
"build": {
  "builder": "@nrwl/workspace:run-commands",
  "options": {
    "commands": [
      {
        "command": "nx run template:prepare"
      },
      {
        "command": "nx run template:build-schematics"
      }
    ],
    "parallel": false
  }
},
```

If I replace the one above with the original build target, then it's working :  

```json
"build": {
  "builder": "@nrwl/angular:package",
  "options": {
    "tsConfig": "libs/template/tsconfig.lib.json",
    "project": "libs/template/ng-package.json"
  }
},
```

No clue how to run the schematics build without modifying this one. I also tested another approach with adding a new project `"template-schematics"` with its own build command but it's not viable depending on the project's build order using the affected command. 

So I'm back to the path alias solution until I find a better way to do this.

PR #590, opened 2021-03-09
Fix change detection not running inside the zone when `patchZone` is truthy

arturovt:


libs/cdk/src/lib/template-management/template-manager.ts
```diff
@@ -192,7 +196,7 @@ export function createTemplateManager<
                   Object.keys(context).forEach((k) => {
                     view.context[k] = context[k];
                   });
-                  work(view, options.scope, notification);
+                  workFactory(() => work(view, options.scope, notification));
```
hoebbelsB, line 199: if we do it like this, we will call `detectChanges` for inserted views twice since it was added to the `createEmbeddedViewFactory`. 
Here is a suggestion to solve:

```ts
 const context = <C>getContext[kind](notification);
                if (isNewTemplate) { // template has changed (undefined => next; suspense => next; ...)
                  // handle remove & insert
                  // remove current view if there is any
                  if (viewContainerRef.length > 0) {
                    // patch removal if needed
                    workFactory(() => viewContainerRef.clear());
                  }
                  // create new view if any
                  if (template) {
                    // createEmbeddedView is already patched, no need for workFactory
                    templates.createEmbeddedView(templateName, context);
                  }
                } else if (template) { // template didn't change, update it
                  // handle update
                  const view = <EmbeddedViewRef<C>>viewContainerRef.get(0);
                  Object.keys(context).forEach((k) => {
                    view.context[k] = context[k];
                  });
                  // update view context, patch if needed
                  workFactory(() => work(view, options.scope, notification));
                }
                activeTemplate = templateName;
```

I mean the problem right now is that we have zone patch at different places. Inside of the `createEmbeddedViewFactory` and inside of `template-manager` and `list-view-handler`. 
We should remove the `createEmbeddedViewFactory` entirely and do everything with the easier `workFactory` approach. But lets skip this refactoring step until next week and make it work for now!

hoebbelsB (changes requested):
hi @arturovt THX again for your finding! Can u please have a look at my suggestion?

PR #597, opened 2021-03-11
chore(build): fix target names and configs

arturovt:
@BioPhoton @hoebbelsB I've debugged the issue and found the root cause. This is related to how builder targets are named right now.

First of all since we're using Nx workspace I've changed builders to native `@nrwl/angular:package` (this is one of the builders that supports `--with-deps` option along with lite ng-packagr).

`'rootDir' is expected to contain all source files` was thrown because Nx wasn't able to find `.d.ts` files for dependencies. In our case the `template` library depends on the `cdk` library. The builder algorithm looks as following:
1) it calls the `createProjectGraph` function which creates the project graph from all files located inside the `template` folder
2) then it calls the `calculateProjectDependencies` function which goes through the graph and collected dependencies and checks if there is any buildable dependency (`rx-angular/cdk` is a buildable dependency)
3) If it's buildable then it tries to collect `outputs` (where that buildable library could potentially be located), and this is **where it fails**. Outputs are collected correctly for dependencies which have `build` target name (and not `build-lib` in our case + it should have targets that actually builds, not runs commands)

Thus I've swapped `build` with `build-lib`, now the `build` target actually builds the library and `build-lib` goes through commands.


arturovt:
upd: `zone-config.spec.ts` is failing. Michael that's on your side :smile:

BioPhoton:
I will get some time over the weekend!!

edbzn:
Hi, I've also made some changes on my side (#405) related to this issue but I didn't have to change the build target name. :thinking:

arturovt:
Because you’re not using Nx builders.

edbzn:
@arturovt I think we should mix our changes.

BioPhoton:
@arturovt it's fixed now. Can u rebase?

PR #598, opened 2021-03-12
Update zone cfg api

cakeinpanic:


BioPhoton:
Thanks for the cleanup!! ❤

PR #599, opened 2021-03-13
Fix template and cdk universal build

### LINKED ISSUE #579: Universal Build Errors because of undefined reference to window
# Description

When doing a universal build with @rx-angular/template execution will fail because of the reference to window in https://github.com/rx-angular/rx-angular/blob/cfd47112c12bf7333e657c2f6b6e79d3d3eccda4/libs/cdk/src/lib/render-strategies/scheduler/scheduler.ts#L63. 

Include `@rx-angular/template` in a server context.

# Environment

"@angular/platform-server": "11.0.8",
"@rx-angular/cdk": "^1.0.0-alpha.4",
"@rx-angular/template": "1.0.0-beta.24",
"@angular-devkit/build-angular": "0.1100.5",
"@nrwl/cli": "11.2.12",

Ivy Workspace: Yes

# Tasks to Resolve This

Use setTimeout instead of window.setTimeout maybe? No Idea to be honest 😄

edbzn:
Fix issues related to universal build `ReferenceError: window is not defined` resolve #579. 

I moved the `getGlobalThis` function from template/core to cdk/utils to use it wherever it's needed.


edbzn:
There's still an error caused by a check [inside the lib](https://github.com/rx-angular/rx-angular/search?q=This+browser+doesn%27t+support+requestAnimationFrame), but that's not causing a build failure anymore.

```
This browser doesn't support requestAnimationFrame. Make sure that you load a polyfill in older browsers. https://reactjs.org/link/react-polyfills
```

edbzn:
I made some investigations with an SSR app and noticed the push pipe (and surely also the let directive) don't render any html in the server side, this PR fix the build step for Node.js runtime so we can bundle RxAngular for this environment but it does not fix the server side rendering. 

Using the native async pipe fix the problem but that's not what we want I guess. While investigating in this issue I noticed that the code use `requestAnimationFrame`, I imagine it's to queue rendering work, which obviously fails in Node.js. I don't know if it's the only breaking part but we can imagine fallback to `setImmediate` in the Node.js platform. 

Is it a goal for you to support server side rendering? Should I open another issue dedicated to this?

arturovt:
> Edit: I'm not convinced about exposing a getGlobalThis function in the CDK, maybe I should use the ɵglobal variable directly.

IMO you can use it directly, I don't think it should be exposed by the CDK package since it will create a tight coupling between other packages (that might exist in the future and that would want to use the `ɵglobal` variable) and the CDK package.

hoebbelsB:
i think we should merge https://github.com/rx-angular/rx-angular/pull/610 this first, than move on to this PR again?

edbzn:
@hoebbelsB sounds good

hoebbelsB (changes requested):
code looks good to me. Can we maybe have a very simple ssr setup in one of our demo applications just to see if things work as expected? This would also increase confidence for future changes that we don't break SSR builds again! Or is it too much of a request? sorry in that case :)

hoebbelsB:
okay after I realised the scope of #610 I guess we should just move on with this one here. I have tried at least a development build in another project with this exact PR. Do u have any test application or test scenario where we can test if your changes really fix universal builds?

edbzn:
I tested directly with one of my apps, available [here](https://github.com/angular-lyon/angular-lyon.github.io), but you're right we should add tests in this repo. What I would suggest is, if manual tests validate that it fixes the build, we should merge this branch first, and then set up an SSR app to integrate universal build in the CI with some automated tests on it. I can take care of it. 

What do you think? I would also suggest to open another issue related to the default strategy which does not work server-side,  ( related to this comment https://github.com/rx-angular/rx-angular/pull/599#issuecomment-798912431)

PR #602, opened 2021-03-14
Extend event names

BioPhoton:


libs/cdk/src/lib/zone-configuration/event-names.ts
```diff
@@ -68,52 +123,80 @@ export const touchEvents: (keyof WindowEventMap)[] = [
...
-  ...inputEvent,
-  ...keyboardEvent,
-  ...compositionEvent,
-  ...touchEvents,
-  ...formControlsEvents,
-  ...globalEvents,
-  ...xhrEvent,
-  ...websocketEvents,
+// XHREvent (group is here to make it easier to target HXR in angular applications. It contains redundant events e.g. 'error')
+export const xhrEvents = ['XHR', 'load', 'error'];
```
cakeinpanic, line 160: todo: add convenience method to disable XHR events altogether with XHR flag itself

BioPhoton: `zoneConfig.rxa.disable.XHR()` maybe?

Where `rxa` is the slot for convenience things?

PR #603, opened 2021-03-15
feat(cdk): infer `getZoneUnPatchedApi` return type and fix types for existing functions

arturovt:


libs/cdk/src/lib/zone-less/utils.ts
```diff
@@ -1,39 +1,27 @@
...
+>(name: N): (Window & typeof globalThis)[N];
+export function getZoneUnPatchedApi<T extends object, N extends keyof T>(
+  name: N,
+  target: T
+): T[N];
+export function getZoneUnPatchedApi<T extends object, N extends keyof T>(
+  name: N,
+  target?: T
+): T[N] {
+  target = target || ɵglobal;
```
edbzn, line 25: What about defining a default target as following :

```suggestion
  target: T = ɵglobal
): T[N] {
```

arturovt: Thanks, updated!

PR #609, opened 2021-03-18
[wip]bring back old typings + refactor flag setting code

cakeinpanic:


libs/cdk/src/lib/zone-configuration/zone-cfg.ts
```diff
@@ -84,80 +118,62 @@ function createZoneFlagsConfigurator(): ZoneConfig {
...
+     * Unpatch all related to XHR
+     **/
+    unpatchXHR: () => {
+      config.global.disable.XHR();
+      config.events.disable.UNPATCHED_EVENTS([...xhrEvent]);
+    },
+    /**
+     * Unpatch passive events https://developers.google.com/web/updates/2016/06/passive-event-listeners
+     **/
+    usePassiveScrollEvents: () => {
```
BioPhoton, line 132: The naming here is not intuitive in terms of the additional unpatch event.

libs/cdk/src/lib/zone-configuration/event-names.ts
```diff
@@ -156,8 +167,14 @@ export const websocketEvents: (keyof WebSocketEventMap)[] = [
...
-// XHREvent (group is here to make it easier to target HXR in angular applications. It contains redundant events e.g. 'error')
-export const xhrEvents = ['XHR', 'load', 'error'];
+/**
+ * Basic XHR Events
+ * [Load](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/load_event)
+ * [Error](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/error_event)
+ * There are more events you may want to unpatch https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest
+ */
+// XHREvent (group is here to make it easier to target XHR in angular applications. It contains redundant events e.g. 'error')
+export const xhrEvents = ['load', 'error'];
```
cakeinpanic, line 177: @BioPhoton I checked, there is no `XHR` event in MDN, also when request is sent using HttpClient, only two events get subscribed on – `load` and `error`, so I removed `XHR` from here

libs/cdk/src/lib/zone-configuration/zone-cfg.ts
cakeinpanic: done!
also reorganised code and interfaces here a bit, check it please

libs/cdk/src/lib/zone-configuration/event-names.ts
BioPhoton: 👍

libs/cdk/src/lib/zone-configuration/zone-cfg.ts
BioPhoton: 👍

PR #610, opened 2021-03-18
feat(cdk): decouple `zone-less` into its own self-contained library

arturovt:
I've moved `zone-less` folder into its own self-contained library so this API is not interconnected with `cdk` package anymore:

Now it's easy to use that API inside tests since Jest will not require mocking `MessageChannel`. This is super useful in Nx workspaces where many libraries can rely on the `zone-less` package but don't use `cdk` package inside tests.

P.S. no functional changes to the API, nothing has been touched, just moved into another folder.


hoebbelsB:
@arturovt I have checked out your branch to my local machine and studied the code a bit.
This would be a breaking change which we need to discuss with @BioPhoton. Btw. why didn't you delete the old zone-less files inside cdk itself?
We do need to publish this as a separate library, no?

arturovt:
> This would be a breaking change

Yuh, sure. What I'm thinking about is that we can provide out custom migration script that will work with `ng update` and update the user's code.

> hy didn't you delete the old zone-less files inside cdk itself?

Sry forgot to do that.

> We do need to publish this as a separate library, no?

No, we don't need. That would be a part of the cdk package.

hoebbelsB:
ok so then besides I think the folder structure appears to be a bit weird now, everything LGTM.

@BioPhoton ?

arturovt:
@BioPhoton @hoebbelsB I've merged with master, resolved conflicts, the CI seems to be happy now.

BioPhoton:
Hi @arturovt sry for the delay, I try to squeeze it in soon! <3

Issue #304, opened 2020-09-09
rxFor 


BioPhoton:
# rxFor

Current implementation available here (rxFor)[https://github.com/rx-angular/rx-angular/tree/master/apps/demos/src/app/rx-angular-pocs/template/directives/for] until released under `rx-angular/template`

Demo:
1. `git clone git@github.com:rx-angular/rx-angular.git`
2. `cd rx-angular`
2. yarn 
3. nx s
4. http://localhost:4200/rx-angular/demos/

## Case Study
The subject is a nested DOM structure rendered as a table with 10 rows and 5 columns every column contains 1 template expression.
The template structure is built with 2 nested `ngFor` directives.

**The situation explained in steps:**
- A list is initially rendered
- The list is updated 4 times in a row with breaks
- The list is updated as fast as possible for some seconds on a `setInterval`
- A list is destroyed

__`ngFor` - Initialize; 4 updates by button; some seconds; as fast as possible, destroy__
<img width="959" alt="ng-for-init-5-click -interval" src="https://user-images.githubusercontent.com/10064416/103499000-e74a1000-4e46-11eb-8d7e-a93e96fecade.PNG">

# Prior Ar

`*ngFor="let item of array"`

Local Variables:
- index
- first
- even
- odd
- last

```html
<div *ngFor="let item of array; trackBy: trackById">
  <div *ngFor="let child of item.children; trackBy: trackById">
   {{child}}
  </div>
</div>
```

# Problem Solved By The Feature

There are several key factors that have a negative impact on rendering performance. The most common of them are rendering big lists of data - or in other terms: lots of workloads.
For rendering lists of data as templates, we typically use the `ngFor` Directive.
Its implementation is pull-based and relies on the current rendering mechanism of dirty marking and top-down (depth-first) evaluation triggered by zone.js, thus only being able to render all changes **synchronously**.

## Bootstrapping

Render a number of items that happen synchronously. This means all items are created in the same browser tick.
<img width="828" alt="sync-rendering-of-items" src="https://user-images.githubusercontent.com/10064416/96523165-52873b80-1275-11eb-8cc9-5385d4544adf.PNG">

This is very good for small to mid-size lists or small lists with a lot of DOM structure. 
For larger lists, this results in blocking UI. 

## Process change detection and render work in chunks

`ngFor` is not able to execute its work in batches. This relates to blocking UI when `create`, `update`, and `destroy` items. 

```html
<ul>
    <ng-container>
[...]

PR #626, opened 2021-03-26
feat(cdk): improve rendering error handling

hoebbelsB:
Sorry, the demo application could be more complete with different scenarios which one could configure via the UI, I can improve at a later stage!

Addresses at least one paint point of #304 => error handling

Final result:

image

image

Open Todos:

- [x] Test behavior on error (.spec)
- [ ] restructure code => I guess this will be done with #632 
- [ ] decide on how the renderCallback will behave on an error case

## Description

currently, if any error gets thrown while `RxListManager` is performing `work` (e.g. `listViewHandler.moveView`), the error won't get caught and ends up being thrown to the user.
The `RxListManager` will stop working after this event happens.

**Example**

I have build a demo scenario where a child-component will cause different errors. I tried to focus on those erros related to `work` in the `RxListManager`. There are examples which are not really catchable for `RxListManager` but are caught by `*ngFor` due to the fact it uses `NgDoCheck`. I will explain later on.

The `ChildComponent` produces an exception in the `RxListManager` by removing its own node from the `DOM`. After this, the `move` operation will fail and throw an error.

```ts
// rxa-error-handling-child

ngDoCheck() {
    if (this._index %2 !== 0 && !this.removed && this.el.nativeElement.parentElement) {
      // remove this node from the DOM, so remove / move operations will fail on me :)
      this.el.nativeElement.parentElement.removeChild(this.el.nativeElement);
      this.removed = true;
    }
  }
```

image

Just for comparison, this is the output of `*ngFor`:

image

I have implemented a scenario which will cause errors for `add` and `update` scenarios as well. Feel free to play around with it!
** This is currently commented out **

```ts
// rxa-error-handling-child
_index: number;
  @Input() set index(index: number) {
    console.log(index);
    this._index = index;
    /*if (index > 20) {
      // throw an error to cause errors on add/update
      throw new Error('erororororo');
    }*/
  }
```

### Where to catch?

My first attempt was to insert a `catchError` pipe to the `render` method of `RxListManager`. It turned out that it won't do anything for us, because the error happens in a complete different context.
The only place where to possibly catch the error thrown by operations on the `ViewContainerRef` is inside the `work` method itself. This is where I ended up wrapping the code with a `try catch` block.

```ts
// list-template-manager.ts (getObservablesFromChangesArray)

try {
 // ...
 listViewHandler.insertView();
 listViewHandler.moveView();
// ...
} catch (e) {
  console.error('RxListManager received an error:, e);
}
```

With that in place errors will get caught correctly and the `RxListManager` won't stop working after receiving it.

In fact, it should interrupt processing the current batch and inform the user about the error.

Since there are a LOT of places where errors could potentially happen out of the regular context I decided to move the `try catch` directly inside the `onStrategy` creationFunction.

```ts
// onStrategy.ts

let error: Error;
  return of(value).pipe(
    strategy.behavior(() => {
      try {
        workFactory(value, strategy.work, options);
      } catch (e) {
        error = e;
      }
    }, options.scope || {}),
    switchMap(() =>
      error ? throwError(errorFactory(error, value)) : of(value)
    )
  );
```

Please note the `errorFactory`. Based on the original `Error` and the emitted `value` the user can decide how the outcoming error looks like and maybe add more information to the output.

It is not advised to mutate the `Error` or anything, instead the resulting interface looks like this:

```ts
export type RxRenderError<T> = [Error, T]; // [originalError, additionalContext]

export type RxRenderErrorFactory<T, E> = (error: Error, value: T) => RxRenderError<E>;
```

With that in place it is possible to pass the original thrown error plus any additional context to the next player.

In case of the `*rxFor / RxListManager` we want to inform the user about the item which led to the error. 

### How to Inform ?

Now that we figured out where to catch the original error we need a way to inform the user. Angular provides an `ErrorHandler` as root service out of the box which is also responsible to print out the errors occuring during `*ngFor` processes items.

The `ErrorHandler` is now part of the `RxRenderSettings`:

```ts

export interface RxRenderSettings<T, C> {
  cdRef: ChangeDetectorRef;
  eRef: ElementRef;
  parent: boolean;
  patchZone: NgZone | false;
  strategies: RxStrategies<string>;
  defaultStrategyName: string;
  errorHandler?: ErrorHandler;
}

```

It gets injected by the respective Directive (`rxFor, rxLet,...`) and passed to the creationFunction.
The `RxTemplateManager / RxListManager` will then use the `createDefaultErrorHandler` function to use the errorHandler:

```ts
// render-error.ts

export function createDefaultErrorHandler(
  _handler?: ErrorHandler
): ErrorHandler {
  const _handleError = _handler ? e => _handler.handleError(e) : console.error;
  return {
    handleError: error => {
      if (isRxRenderError(error)) {
        _handleError(error[0]);
        console.error('additionalErrorContext', error[1]);
      } else {
        _handleError(error);
      }
    }
  }
}

// list-template-manager.ts
const errorHandler = createDefaultErrorHandler(renderSettings.errorHandler);

/**/
         catchError(e => {
              partiallyFinished = false;
              errorHandler.handleError(e);
              return e;
            })
```

## Known issues

I don't know if its really important, but rendering output won't be the same for `rxLet` vs. `async` or `rxFor` vs. `ngFor`. This is because of the totally different implementations and different timings when changes are applied to the view. But in either case, nothing should stop working in case of an error.


hoebbelsB:
@BioPhoton can u check this one? I have included the error handling for *rxLet aswell. Implemented a demo for this.
I have removed the `console.error` call from `rxMaterialize`. Imo the user of `rxMaterialize` should decide what to do in case of an error.

BioPhoton:
> The ErrorHandler is now part of the RxRenderSettings
I like it a lot!

In general, this is all in all a well-fitting solution! So cool we have now CDK wide error handling (and nearly all communication ways) 

THX A LOT! 🤘🤘🤘

BioPhoton (approved):

PR #640, opened 2021-04-06
fix(demos): fix most of the demos

hoebbelsB:
Fixed the most important demos:

* renderCallback
* rxContext (fixed rxSwitch aswell)
* strategy comparison
* coalescing
* push
* rxLet basics


BioPhoton:
♥♥♥♥

PR #646, opened 2021-04-11
fix fromEvent always subscribing on document

cakeinpanic:
Bug: 
fromEvent always added listened on `window` element

[Stackblitz reproduction](https://stackblitz.com/edit/rx-angular-unpathced-bug)


BioPhoton:
Nice catch! 
Thanks a lot for your time!

Issue #661, opened 2021-04-16
Error logging in hold streams


d-stolyar:
# Description

Errors in hold streams are not logged

# Steps to Reproduce the Issue

Check [stackblitz](https://stackblitz.com/edit/angular-ivy-sndw3n?file=src%2Fapp%2Fapp.component.ts) example

```
this.hold(
  from([1, 2, 3]).pipe(switchMap(() => throwError("error"))),
  (value: number) => {
    console.log(value);
  }
);
```
Result in console:
` `
Expected result in console:
`error error error`
# Solution
Add console.error in hold definition
```
hold<S>(
  obsOrObsWithSideEffect: Observable<S>,
  sideEffectFn?: (arg: S) => void
): void {
  const sideEffect = obsOrObsWithSideEffect.pipe(catchError(e => {
    console.error(e)
    return EMPTY
  }))
  if (typeof sideEffectFn === 'function') {
    this.effectObservable.nextEffectObservable(
      sideEffect.pipe(tap(sideEffectFn))
    );
    return;
  }
  this.effectObservable.nextEffectObservable(sideEffect);
}
```

BioPhoton:
Thx for the issue!
Get back with updates.

BioPhoton:
Hi!

In the PR for RxEffects we inplemented angulars error handler. Would be col if we could do the same here.
https://github.com/rx-angular/rx-angular/pull/569/files#diff-9afb99d5c66b09318924448921fd9e1f7eefe3b2d5244af8f125266f57378c93R75

PR #678, opened 2021-04-22
added logging to the hold method #661

d-stolyar:


hoebbelsB:
hi @d-stolyar THX for the hotfix! :+1: 
Could you maybe add a test which covers this? This is already a regression and I'd like it to be the last one for swallowing errors in effects :)

hoebbelsB:
oh forgot to mention... If you need any help or don't want to write tests, just ping one of us :+1:

BioPhoton:
Hi @d-stolyar any updates on your end? We would love to get that change in :)

BioPhoton:
@distoliar any chance you can bring that in too?

https://github.com/rx-angular/rx-angular/issues/661#issuecomment-832913105

edbzn (commented):
Hey, I left two comments regarding the DI. Also, I'm wondering why the `ErrorHandler` is injected as an optional dependency?

libs/state/src/lib/rx-state.service.ts
```diff
@@ -547,7 +587,13 @@ export class RxState<T extends object> implements OnDestroy, Subscribable<T> {
     obsOrObsWithSideEffect: Observable<S>,
     sideEffectFn?: (arg: S) => void
   ): void {
-    const sideEffect = obsOrObsWithSideEffect.pipe(catchError(e => EMPTY))
+    const sideEffect = obsOrObsWithSideEffect.pipe(
+      catchError((e) => {
+        const errorHandler = inject(ErrorHandler, InjectFlags.Optional);
```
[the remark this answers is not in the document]
arturovt: Yeah, it actually will always fail here. Providing `InjectFlags` is a "hack" to make the runtime behaviour happy (since the `inject must be called blah-blah...` error is not thrown). I also can't think of a situation where `ErrorHandler` can be `null` since it's initially provided by Angular or overwritten by developers.

libs/state/spec/rx-state.service.spec.ts
```diff
@@ -590,5 +594,18 @@ describe('RxStateService', () => {
...
       expect(calls).toBe(3);
     }));
+
+    it('should invoke provided ErrorHandler', fakeAsync(() => {
+      const state = setupState({ initialState: initialPrimitiveState });
+      const customErrorHandler: ErrorHandler = {
+        handleError: jest.fn()
+      };
+      TestBed.overrideProvider(ErrorHandler, { useValue: customErrorHandler });
+      setCurrentInjector(TestBed.inject(Injector));
```
arturovt, line 604: We shouldn't be using this inside tests. Just remove the `inject` function and use the constructor injection as @edbzn proposed; therefore, you won't need to set the current injector.

libs/state/src/lib/rx-state.service.ts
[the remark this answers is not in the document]
d-stolyar: If injecting the ErrorHandler using the constructor:
```
constructor(private errorHandler: ErrorHandler) {}
```
will not be backward compatible with -  https://github.com/rx-angular/rx-angular/blob/master/libs/state/docs/usage.md#inherit
```
@Directive({
  selector: '[appStateful]',
})
export class StatefulComponent extends RxState<{ foo: number }> {
  readonly state$ = this.select();

  constructor(private errorHandler: ErrorHandler) {
    super(this.errorHandler);
  }
}
```
how can this be fixed?

BioPhoton:
Hi @d-stolyar.
THX for bearing with us! :)

Is the suggested change clear or should we add more information?
Just let me know how we can help.

d-stolyar:
> hi @d-stolyar THX for the hotfix! 
> Could you maybe add a test which covers this? This is already a regression and I'd like it to be the last one for swallowing errors in effects :)

https://github.com/rx-angular/rx-angular/pull/678#discussion_r643774841

d-stolyar:
> THX for bearing with us! :)
> 
> Is the suggested change clear or should we add more information?
> Just let me know how we can help.

https://github.com/rx-angular/rx-angular/pull/678#discussion_r643774841

libs/state/src/lib/rx-state.service.ts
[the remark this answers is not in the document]
edbzn: That's a use case I didn't think about, then I would say go for the injector usage and document why in a comment.

fpsstyle228 (approved):

BioPhoton (approved):
Lets merge it finaly! 😁🚀

hoebbelsB:
@d-stolyar can u please merge master into your branch? :)

libs/state/src/lib/rx-state.service.ts
```diff
@@ -547,7 +587,14 @@ export class RxState<T extends object> implements OnDestroy, Subscribable<T> {
     obsOrObsWithSideEffect: Observable<S>,
     sideEffectFn?: (arg: S) => void
   ): void {
-    const sideEffect = obsOrObsWithSideEffect.pipe(catchError(e => EMPTY))
+    const sideEffect = obsOrObsWithSideEffect.pipe(
+      catchError((e) => {
+        // used injector for compatibility with https://github.com/rx-angular/rx-angular/blob/master/libs/state/docs/usage.md#inherit
+        const errorHandler = inject(ErrorHandler, InjectFlags.Optional);
+        errorHandler?.handleError(e);
```
edbzn, line 594: Last detail, why considering the `ErrorHandler` as optional? It should always be defined.

```suggestion
        const errorHandler = inject(ErrorHandler);
        errorHandler.handleError(e);
```

arturovt: He has probably added it because it throws `inject() must be called from an injection context` since there's no injector on the instruction stack.

You recommendation with this one is correct:
```ts
constructor(private errorHandler: ErrorHandler) {}
```

Since the stable version is not released yet I think it's ok to have breaking changes. We can have a migration script for that.

edbzn: We can create a migration script, but honestly, this looks bad:

```ts
  constructor(private errorHandler: ErrorHandler) {
    super(this.errorHandler);
  }
  ```
  
I would prefer keeping the `InjectFlags.Optional` hack to hide this noise from the user.

Comment (arturovt) [reply libs/state/src/lib/rx-state.service.ts:594]: I've just checked that code and it seems like it's working only inside tests for now because the injector is explicitly set there. E.g. this doesn't work even with inject flags:
```ts
@Component({
  selector: 'app-root',
  template: '<button (click)="onClick()">Click me</button>',
})
export class AppComponent {
  onClick(): void {
    console.log(inject(ErrorHandler, InjectFlags.Optional));
  }
}
```

hoebbelsB: > 
> I'm wondering if it can be moved to the constructor but it also has to be wrapped into try-catch to prevent the error if there is no current injector:
> 
> ```ts
> export class ... {
>   private errorHandler: ErrorHandler | null;
> 
>   constructor() {
>     try {
>       this.errorHandler = inject(ErrorHandler);
>     } catch {
>       this.errorHandler = null;
>     }
>   }
> ```

this sounds like the only valid solution for me since the approach with extending from RxState is pretty common.

d-stolyar: How about this?
```javascript
class ... {
  private errorHandler: ErrorHandler

  constructor() {
    try {
      this.errorHandler = inject(ErrorHandler)
    } catch {
      const injector = Injector.create({
        providers: [{ provide: ErrorHandler, deps: [] }]
      });
      this.errorHandler = injector.get(ErrorHandler);
    }
  }
```

arturovt: This will create `ErrorHandler`s per each RxState instance and will have nothing to do with the global error handler (the one provided by a user).

edbzn: To conclude, we should try this approach and decide:

```ts
export class ... {
  private errorHandler: ErrorHandler | null;

  constructor() {
    try {
      this.errorHandler = inject(ErrorHandler);
    } catch {
      this.errorHandler = null;
    }
  }
```

PR #684, opened 2021-04-28
Fixes #634 - Support template typing for PushPipe

### LINKED ISSUE #634: PushPipe & RxLet do not preserve template typing for bound/piped Observables
# Description

It appears that the `push` pipe and `rxFor=` (experimental) operators do not work with Ivy template type checking. The type always breaks down to '`string`' or '`any`':

Screen Shot 2021-03-30 at 4 50 17 PM

angular compiler config in tsconfig.json:
```
  "angularCompilerOptions": {
    "enableIvy": true,
    "fullTemplateTypeCheck": true,
    "skipTemplateCodegen": false,
    "strictInjectionParameters": true,
    "strictMetadataEmit": true,
    "strictTemplates": true
  }
```

I would expect this to resolve the type:

Screen Shot 2021-03-30 at 4 52 57 PM

1. Use a `push` pipe on a bound `Observable<T>` and then try to reference properties of `T` in the template. They will all be of unknown type because `T` is unknown type. See screenshots above.
2. execute `ng run app:build:production`

See errors:
```
[ng]     Error occurs in the template of component TestPage.
[ng] Error: src/app/pages/test.page.html:10:40 - error TS2339: Property 'type' does not exist on type 'string'.
[ng] 10                     format="{{ payload.type }}">
[ng]                                           ~~~~        
```

# Environment

```
VSCode Angular Language Service plugin: 11.2.9 (latest at time of filing)
    "@rx-angular/cdk": "^1.0.0-alpha.7",
    "@rx-angular/state": "^1.4.3",
    "@rx-angular/template": "^1.0.0-beta.24",



Angular CLI: 12.0.0-next.5
Node: 14.15.0
OS: darwin x64

Angular: 12.0.0-next.6
... language-service, platform-browser, platform-browser-dynamic
... router, service-worker
Ivy Workspace: Yes

@angular-devkit/architect       0.1200.0-next.5
@angular-devkit/build-angular   0.1200.0-next.5
@angular-devkit/core            12.0.0-next.5
@angular-devkit/schematics      12.0.0-next.5
@angular/cdk                    12.0.0-next.4
@angular/cli                    12.0.0-next.5
@angular/fire                   6.1.4
@angular/pwa                    0.1200.0-next.5
@schematics/angular             12.0.0-next.5
@schematics/update              0.1200.0-next.5
rxjs                            6.6.7
typescript                      4.2.3
```

# Related to Other Issues

# Tasks to Resolve This

# Notes

lincolnthree:
This patch fixes the issue where using PushPipe in templates causes all type information to be dropped during compilation. (Probably an angular bug, but this works around the issue and allows building with `strictTemplates: true` enabled.)


libs/template/src/lib/push/push.pipe.ts
```diff
@@ -64,13 +64,13 @@ import { filter, switchMap, tap, withLatestFrom } from 'rxjs/operators';
  * @publicApi
  */
 @Pipe({ name: 'push', pure: false })
-export class PushPipe<U> implements PipeTransform, OnDestroy {
+export class PushPipe implements PipeTransform, OnDestroy {
   /** @internal */
-  private renderedValue: U | null | undefined;
+  private renderedValue: any | null | undefined;
```
hoebbelsB, line 69: can we add a comment here to explain at least why this is any? I could imagine that we will introduce some linting rule which will complain about this in the future.

libs/template/src/lib/push/push.pipe.ts
```diff
@@ -84,13 +84,13 @@ export class PushPipe<U> implements PipeTransform, OnDestroy {
     const scope = (cdRef as any).context;
     this.subscription = this.templateObserver.values$
       .pipe(
-        filter<RxNotification<U>>(
+        filter<RxNotification<any>>(
```
hoebbelsB, line 87: I think u can drop the whole type specification if we go with `any` .
```suggestion
        filter(
```

libs/template/src/lib/push/push.pipe.ts
```diff
@@ -84,13 +84,13 @@ export class PushPipe<U> implements PipeTransform, OnDestroy {
...
           (n) =>
             n.kind === RxNotificationKind.suspense ||
             n.kind === RxNotificationKind.next
         ),
-        tap<RxNotification<U>>((notification) => {
-          this.renderedValue = notification.value as U;
+        tap<RxNotification<any>>((notification) => {
+          this.renderedValue = notification.value as any;
```
hoebbelsB, line 93: same goes for this one:
```suggestion
        tap(notification => {
          this.renderedValue = notification.value;
```

hoebbelsB (changes requested):
@lincolnthree BIG THX for finding and fixing the typing issue :+1: :+1: 

I've left some minor comments on your work :)

BioPhoton:
I think this is related https://github.com/ngrx/platform/issues/2890

libs/template/src/lib/push/push.pipe.ts
lincolnthree: Without a class-level generic type option, there is no longer a way to know what value type will be passed into the `transform()` method. Happy to leave a code comment about that if you think it's necessary, but may be better just for the issue notes since it should be generally inferable?

libs/template/src/lib/push/push.pipe.ts
lincolnthree: Agreed.

libs/template/src/lib/push/push.pipe.ts
lincolnthree: Agreed.

lincolnthree:
@BioPhoton Are you sure? My PR should only affect compile-time typing, not runtime behavior.

libs/template/src/lib/push/push.pipe.ts
hoebbelsB: I understand why we change it to any. just some internal note for us as reminder :). u think its worthless?

lincolnthree: Oh I see. If that's your intent sure! Sorry, I'm not used to using comments for this purpose. I'll update it.

lincolnthree:
@hoebbelsB Updated, please review!

libs/template/src/lib/push/push.pipe.ts
hoebbelsB: love it

hoebbelsB (approved):
LGTM

lincolnthree:
@hoebbelsB I did my best. Please review carefully!

lincolnthree:
Err. Hang on.

lincolnthree:
Okay, should be good to go. LMK if you find any other issues during review.

PR #688, opened 2021-04-30
feat(cdk): decouple `zone-less` into its own self-contained library

arturovt:


BioPhoton:
Hi @arturovt THX for this nice improvement.

I thought about deprecation the old import before deleting it. What is your opinion here?
As we are alpha I see no big problem to just drop it in alpha, just was wondering what the "professional" approach would be in general.

Let me know your thoughts.

PR #700, opened 2021-05-02
perf(cdk): use existing `rxjs` schedulers

arturovt:
I removed the implementation of all schedulers and just re-used the existing constructors.

I still left interfaces to make the compiler happy, since schedulers are not exposed publicly by rxjs (tho they don't affect the bundle size).


BioPhoton:
BAM!!
Nice idea! didn't know I could do that. 

:D

PR #707, opened 2021-05-03
feat(cdk): swap arguments in the `getZoneUnPatchedApi` to benefit from autocomplete

arturovt:
I've swapped 2 arguments in the `getZoneUnPatchedApi` thus we're able to benefit from the autocomplete. It will infer all keys automatically and will provide all possible values e.g. for the `MutationObserver`, but not vice versa.

This is like the `get` lodash method, you pass an object and then the key you want to get like `get(object, 'key')`, but not vice versa `get('key', object)`:

image

Also fixed annoying warnings in tests.


BioPhoton:
Amazing finding!
Learned something!
🙏

PR #725, opened 2021-05-06
feat(cdk): move scheduler into `internals/scheduler`

arturovt:
Moved scheduler implementation into "internals" library since it's for internal use only. Also added tests for the scheduler to ensure it's really working :D


BioPhoton:
Thx!! Also, nice u added tests! Really nice!

PR #732, opened 2021-05-10
feat(template): decouple `let` into a library

arturovt:
Too many files changed because of the prettier and also had to update imports inside the `apps/demos`. No functional changes tho.


BioPhoton:
Nice! Thx also for reexporting and avoiding a breaking change before we have migrations in place! 🤩

PR #750, opened 2021-05-14
chore(ci): enable partial Ivy compilation for publishable packages

### LINKED ISSUE #742: Enable Ivy partial compilation

Ngcc will skip RxAngular packages.

# Additional Context

It's a breaking change since it targets Angular v12 and Ivy only, more info [here](https://blog.ninja-squad.com/2021/01/27/angular-linker/)

arturovt:


edbzn:
Nice! I think it should be described as a breaking change in the commit message since users will have to use Angular >= 12 and Ivy.

edbzn (commented):
Nice! I think it should be described as a breaking change in the commit message since users will have to use Angular >= 12 + Ivy.

arturovt:
> Nice! I think it should be described as a breaking change in the commit message since users will have to use Angular >= 12 + Ivy.

Merci pour review, I'll squash commits and amend the message soon!

BioPhoton:
I'm not sure if we should go ng12 + ivy only.
will have a think

BioPhoton:
I would say we wait for this one a little bit. There are quite some projects using rxa ATM that will not go v 12 soon.

Especially some of the low-end device projects.

hoebbelsB:
@arturovt @BioPhoton 
we just had a discussion about this topic in a meet. This won't be a problem for those people, just saving build times as far as I understood. correct me if i'm wrong pls

edbzn:
Using partial compilation saves compile time but relies on the Angular Linker introduced in Angular CLI version 11.1, not version 12 as I mentioned before. It's still a breaking change but should be introduced IMO.

wSedlacek:
@arturovt @BioPhoton @hoebbelsB
Angular v10 has LTS support ending on Dec 24, 2021.
Source: https://angular.io/guide/releases#support-policy-and-schedule

At that point would it make sense to introduce this change?

libs/cdk/tsconfig.lib.json
```diff
@@ -8,7 +8,8 @@
     "lib": ["dom", "es2018"]
   },
   "angularCompilerOptions": {
-    "enableIvy": false,
+    "enableIvy": true,
```
wSedlacek, line 11: `enableIvy` appears to have been removed as a compiler option on the latest versions of Angular.

https://github.com/angular/angular/commit/4ed8dc858aebfdf0ab271ec8c710e60f3d10058a#diff-0a65f1d37c6a41d9b8c6e91e484e9e58bb815f22e9b3425432c086c1b904eefa


The nx migration screen to 13.2 automatically removes this option too.

edbzn: We also have to remove umd options like `umdModuleIds`.

edbzn:
Just noticed @hoebbelsB @BioPhoton, in master, we migrated to ng-packagr 13, this version does not build to ViewEngine anymore and compiles using the partial mode by default. I would suggest cleaning this PR asap instead of reverting to ng-packagr 12, what do you think?

libs/cdk/tsconfig.lib.json
edbzn: Nevermind, it's already done :+1:

hoebbelsB:
close because obsolete with v13. we just need to release one more time with enableIvy: false

PR #752, opened 2021-05-14
feat(template): decouple `unpatch` into a lib

arturovt:


libs/template/src/index.ts
```diff
@@ -1,11 +1,15 @@
...
+export {
+  UnpatchModule as UnpatchEventsModule,
+  UnpatchDirective,
+} from '@rx-angular/template/unpatch';
 
 export { ViewportPrioModule } from './lib/experimental/viewport-prio/viewport-prio.module';
 export { ViewportPrioDirective } from './lib/experimental/viewport-prio/viewport-prio.experimental.directive';
 
 // STABLE
+// TODO(artur): we're keeping it here until we get a migration script.
```
BioPhoton, line 12:
```suggestion
// @deprecated import from template/unpatch
// TODO(artur): we're keeping it here until we get a migration script.
```

BioPhoton (approved):
Thx for the next lib spliced out  :)

PR #754, opened 2021-05-14
chore(ci): disable ES5 generation since it slows down the build

arturovt:


hoebbelsB:
this is a cool idea! the demo app isn't hosted anywhere, is it?

apps/docs/.browserslistrc
```diff
@@ -5,8 +5,4 @@
 # You can see what browsers were selected by your queries by running:
 #   npx browserslist
 
-> 0.5%
```
hoebbelsB, line 8: @Karnaukhov-kh will this be a problem for our docs page?

apps/ssr/.browserslistrc
```diff
@@ -9,9 +9,3 @@
 #   npx browserslist
 
 last 1 Chrome version
-last 1 Firefox version
```
hoebbelsB, line 12: @edbzn is this fine for the ssr setup or do we require having those settings here?

hoebbelsB (commented):
great idea, just left comments where I think we should have a second look at

apps/ssr/.browserslistrc
edbzn: It's ok.

apps/docs/.browserslistrc
arturovt: The docs page is not an Angular app.

BioPhoton:
Other opinions on dropping IE?

Karnaukhov-kh:
In the Readme we are saying that we support IE. This is true for state but not true for template, because we still have globalThis and it requires a polyfill for IE and old Edge.

Dropping IE and old Edge sounds good but maybe we need to have a separate browser support for state and template then.

arturovt:
Well, this change actually related to apps in our repository but not to libraries, `ng-packagr` doesn't use browserlistrc configuration.

edbzn:
This PR is about demos, ssr, etc.. which are not production apps, it will not affect library distribution bundles.

> In the Readme we are saying that we support IE. This is true for state but not true for template, because we still have globalThis and it requires a polyfill for IE and old Edge.

True, this is confusing.

Karnaukhov-kh:
Ok, no objections from my side then :)

PR #806, opened 2021-06-03
Add migration schematic

edbzn:
As for now this migration script target versions `< 1.0.0` and assumes next version will be `1.0.0`, let me know if you want to release with another version.

### DIFF
--- a/angular.json
+++ b/angular.json
@@ -600,6 +600,9 @@
                 "command": "npx cpx libs/template/schematics/collection.json dist/libs/template/schematics"
               },
               {
+                "command": "npx cpx libs/template/schematics/migration.json dist/libs/template/schematics"
+              },
+              {
                 "command": "npx cpx libs/template/schematics/src/**/files/** dist/libs/template/schematics/src"
               },
               {
--- a/libs/template/package.json
+++ b/libs/template/package.json
@@ -59,5 +59,8 @@
   "schematics": "./schematics/collection.json",
   "ng-add": {
     "save": "dependencies"
+  },
+  "ng-update": {
+    "migrations": "./schematics/migration.json"
   }
 }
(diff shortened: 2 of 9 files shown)


edbzn:
Note that dealing with the following scenario was pretty hard: 

```ts
import { LetModule, LetDirective, PushModule } from '@rx-angular/template';
```

The actual code outputs: 

```ts
import { LetModule } from '@rx-angular/template/let';
import { LetDirective } from '@rx-angular/template/let';
import { PushModule } from '@rx-angular/template/push';
```

Instead of:

```ts
import { LetModule, LetDirective } from '@rx-angular/template/let';
import { PushModule } from '@rx-angular/template/push';
```

arturovt:
Hey @edbzn . Looks amazing! Do you think we could re-use some schematics code to avoid copy-pasting it in our codebase?

edbzn:
@arturovt I thought about it because it's a real problem, angular-cli does not expose all the code publicly, that's why many repos including ngrx, nx and some others are copying the same code and adjust it to their need. It's a mess because they all have more or less the same code written differently. To avoid this I decided to import some Nx internals, for instance:

```ts
import { insert, RemoveChange, ReplaceChange } from '@nrwl/workspace/src/utils/ast-utils';
```

There's still room for improvement, I could use the `visitNotIgnoredFiles` from `@nrwl/devkit` instead of the actual `visitTSSourceFiles` I copied from ngrx, but it works a bit differently, so I will need to check it.

This afternoon I was thinking about creating an ast-utils library, but I think it's preferable to open an issue in Nx to move all the code related to the ast in the devkit, and expose it publicly. Actually, writing ast changes is really not DX friendly... :smile:

BioPhoton:
Yeah this is a real mess for schematics. 😩
Really should be a public API for more things.

Also had so much trouble get something nice running.

edbzn:
Opened an issue https://github.com/nrwl/nx/issues/5902.

edbzn:
Any clue why the git comparison fails in the CI? :thinking:

BioPhoton (approved):
Thx for this migration @edbzn. Major step to stable release

Issue #726, opened 2021-05-07
Improve support for non-browser environments


edbzn:
As a @rx-angular/template user I would like to have better support for non-browser environments, this is particularly relevant for testing and server environment where default strategy fails to render something.

- the documentation should mention these use cases and provides some guidelines,
- (optional, needs discussion) the library should provide some utils to improve DX (for instance [this](https://github.com/rx-angular/rx-angular/issues/714#issuecomment-832666021)),
- non-browser builds should not output irrelevant errors : 

```
This browser doesn't support requestAnimationFrame. Make sure that you load a polyfill in older browsers. https://reactjs.org/link/react-polyfills
```

I would like to have your feedback before working on this.

BioPhoton:
Sounds pretty good!

BioPhoton:
Well,  the edge cases maybe?

edbzn:
Related to #1314

hoebbelsB:
@edbzn wasn't it closed by https://github.com/rx-angular/rx-angular/pull/820 ?

edbzn:
There are points still not complete (doc and test utils) but it's tracked in another issue, let's close.

Issue #714, opened 2021-05-04
rxLet `ComponentFixture` error: "Cannot read property 'nativeElement' of null"


edbzn:
# Description

Hi there, 

I have trouble with the `LetDirective` in a testing environment, when I try to query the native element to test the DOM I face the following error (using Karma/Jasime): 

```
TypeError: Cannot read property 'nativeElement' of null
            at UserContext.<anonymous> (http://localhost:9876/_karma_webpack_/src/app/app.component.spec.ts:21:48)
            at ZoneDelegate.invoke (http://localhost:9876/_karma_webpack_/node_modules/zone.js/dist/zone-evergreen.js:364:1)
            at ProxyZoneSpec.push.QpwO.ProxyZoneSpec.onInvoke (http://localhost:9876/_karma_webpack_/node_modules/zone.js/dist/zone-testing.js:292:1)
            at ZoneDelegate.invoke (http://localhost:9876/_karma_webpack_/node_modules/zone.js/dist/zone-evergreen.js:363:1)
            at Zone.run (http://localhost:9876/_karma_webpack_/node_modules/zone.js/dist/zone-evergreen.js:123:1)
            at runInTestZone (http://localhost:9876/_karma_webpack_/node_modules/zone.js/dist/zone-testing.js:545:1)
            at UserContext.<anonymous> (http://localhost:9876/_karma_webpack_/node_modules/zone.js/dist/zone-testing.js:560:1)
```

Using Jest I get another output, maybe more verbose: 

```
 TypeError: Cannot read property 'nativeElement' of null

      19 |     fixture.detectChanges();
      20 |     expect(
    > 21 |       fixture.debugElement.query(By.css('div')).nativeElement.innerHTML
         |                                                ^
      22 |     ).toContain('hello');
      23 |   });
      24 | });

      at src/app/app.component.spec.ts:21:48
      at ZoneDelegate.Object.<anonymous>.ZoneDelegate.invoke (../../node_modules/zone.js/dist/zone.js:386:30)
      at ProxyZoneSpec.Object.<anonymous>.ProxyZoneSpec.onInvoke (../../node_modules/zone.js/dist/proxy.js:117:43)
      at ZoneDelegate.Object.<anonymous>.ZoneDelegate.invoke (../../node_modules/zone.js/dist/zone.js:385:36)
      at Zone.Object.<anonymous>.Zone.run (../../node_modules/zone.js/dist/zone.js:143:47)

```

Here is the test: 

```ts
[...]

PR #820, opened 2021-06-21
fix(cdk): setup polyfills for non browser environments in scheduler

hoebbelsB:
Addresses those two issues:
https://github.com/rx-angular/rx-angular/issues/714 & https://github.com/rx-angular/rx-angular/issues/726 .

The scheduler should now be able to run with proper fallbacks (even optimized for node.js) in non browser environments


libs/template/let/src/lib/tests/let.directive.strategy.spec.ts
```diff
@@ -73,7 +67,76 @@ describe('LetDirective when using strategy', () => {
...
 
-  it('should work with different if a strategy other than the default', () => {
+  it('should work with default strategy', done => {
+    letDirectiveTestComponent.value$ = of(5);
+    fixtureLetDirectiveTestComponent.detectChanges();
+    setTimeout(() => {
+      expect(componentNativeElement.textContent).toBe('5');
+      done();
+    }, 1);
+  });
```
edbzn, line 77: What about using `fakeAsync` + `flush` to avoid `setTimeout` callback + `done` fn for all asynchronous tests? 

```ts
  it('should work with default strategy', fakeAsync(() => {
    letDirectiveTestComponent.value$ = of(5);
    fixtureLetDirectiveTestComponent.detectChanges();
    flush();
    expect(componentNativeElement.textContent).toBe('5');
  }));
```

edbzn (commented):
I ran your branch on my machine and tests are green but there's always an error logged after tests passed :

```
Attempted to log "ERROR TypeError: Cannot read property 'createComment' of null
        at DefaultDomRenderer2.Object.<anonymous>.DefaultDomRenderer2.createComment (***/packages/platform-browser/src/dom/dom_renderer.ts:151:21)
        at DebugRenderer2.Object.<anonymous>.DebugRenderer2.createComment (***/packages/core/src/view/services.ts:736:35)
        at createElement (***/packages/core/src/view/element.ts:168:21)
        at createViewNodes (***/packages/core/src/view/view.ts:275:20)
        at createEmbeddedView (***/packages/core/src/view/view.ts:209:3)
        at callWithDebugContext (***/packages/core/src/view/services.ts:641:23)
        at Object.debugCreateEmbeddedView [as createEmbeddedView] (***/packages/core/src/view/services.ts:146:10)
        at TemplateRef_.Object.<anonymous>.TemplateRef_.createEmbeddedView (***/packages/core/src/view/refs.ts:354:34)
        at ViewContainerRef_.Object.<anonymous>.ViewContainerRef_.createEmbeddedView (***/packages/core/src/view/refs.ts:188:33)
        at createEmbeddedView (***/rx-angular/libs/cdk/src/lib/template-management/utils.ts:111:35)
        at Object.createEmbeddedView (***/rx-angular/libs/cdk/src/lib/template-management/utils.ts:139:48)
        at ***/rx-angular/libs/cdk/src/lib/template-management/template-manager.ts:86:57
```

libs/template/let/src/lib/tests/let.directive.strategy.spec.ts
arturovt: Hey Edouard. Would `await fixture.whenStable()` work in that case? I mean:
```ts
 it('should work with default strategy', async () => { 
  ...
  await ...
  expect ...
```

I'm only curious, haven't checked locally.

edbzn: Agree it also works and it's even better. 

```ts
  it('should work with default strategy', async () => {
    letDirectiveTestComponent.value$ = of(5);
    fixtureLetDirectiveTestComponent.detectChanges();
    await fixtureLetDirectiveTestComponent.whenStable();
    expect(componentNativeElement.textContent).toBe('5');
  });
```

hoebbelsB: Sorry im a test newbie, thx for help :)

hoebbelsB: @arturovt @edbzn implemented it. the error is still there. is there something wrong with my implementation?

libs/cdk/internals/scheduler/src/lib/scheduler.ts
```diff
@@ -456,20 +462,42 @@ const performWorkUntilDeadline = () => {
...
   channel.port1.onmessage = performWorkUntilDeadline;
+  schedulePerformWorkUntilDeadline = () => {
+    port.postMessage(null);
+  };
+} else {
+  // We should only fallback here in non-browser environments.
+  schedulePerformWorkUntilDeadline = () => {
+    localSetTimeout(performWorkUntilDeadline, 0);
+  };
 }
```
edbzn, line 494: @hoebbelsB I think we should unit test this, to fix the Scheduler tests I just skipped the `setImmediate` usage by setting it to `undefined` globally, but we should test the Scheduler uses the correct scheduling mechanism.

1. `setImmediate` if defined (for Node)
2. `MessageChannel` if defined
3. `setTimeout` fallback

hoebbelsB:
@edbzn thx for the feedback! I have implemented tests which should cover all cases. Can u have a look again? :)

edbzn (approved):
Nice! This is really cool, no more strategy hacks for setting up a test environment.

PR #832, opened 2021-07-06
rxFor error with @angular/cdk drag and drop

### LINKED ISSUE #831: rxFor error with @angular/cdk drag and drop
# Description

Using `rxFor` with `@angular/cdk/drag-drop` can potentially lead to an unrecoverable error state.

```bash
list-view-handler.ts:76 DOMException: Failed to execute 'insertBefore' on 'Node': The node before which the new node is to be inserted is not a child of this node.
    at HTMLUListElement.insertBefore (<anonymous>)
    at EmulatedEncapsulationDomRenderer2.insertBefore (http://localhost:4200/rx-angular/demos/vendor.js:80210:20)
    at BaseAnimationRenderer.insertBefore (http://localhost:4200/rx-angular/demos/vendor.js:79206:23)
    at nativeInsertBefore (http://localhost:4200/rx-angular/demos/polyfills.js:8514:18)
    at applyToElementOrContainer (http://localhost:4200/rx-angular/demos/polyfills.js:8028:13)
    at applyNodes (http://localhost:4200/rx-angular/demos/polyfills.js:8757:17)
    at applyView (http://localhost:4200/rx-angular/demos/polyfills.js:8764:5)
    at addViewToContainer (http://localhost:4200/rx-angular/demos/polyfills.js:8108:5)
    at ViewContainerRef.insert (http://localhost:4200/rx-angular/demos/polyfills.js:24256:13)
    at ViewContainerRef.move (http://localhost:4200/rx-angular/demos/polyfills.js:24266:21)
moveView @ list-view-handler.ts:76
(anonymous) @ list-template-manager.ts:211
(anonymous) @ onStrategy.ts:63
task.delay @ concurrent-strategies.ts:108
workLoop @ scheduler.ts:139
flushWork @ scheduler.ts:114
performWorkUntilDeadline @ scheduler.ts:439
```
```bash
core.js:6456 ERROR TypeError: Cannot read property 'context' of null
    at MapSubscriber.project (list-template-manager.ts:112)
    at MapSubscriber._next (map.js:29)
    at MapSubscriber.next (Subscriber.js:49)
    at DistinctUntilChangedSubscriber._next (distinctUntilChanged.js:50)
    at DistinctUntilChangedSubscriber.next (Subscriber.js:49)
    at SwitchMapSubscriber.notifyNext (switchMap.js:66)
    at SimpleInnerSubscriber._next (innerSubscribe.js:10)
    at SimpleInnerSubscriber.next (Subscriber.js:49)
    at BehaviorSubject.next (Subject.js:39)
    at BehaviorSubject.next (BehaviorSubject.js:30)
defaultErrorLogger @ core.js:6456
handleError @ core.js:6504
...
```

After this error happens, the view will be broken and items start to disappear. If you mirror the outcome of the operations next to an `rxFor` with an `ngFor` you can actually observe how items start to be out of sync with the state.

I am still trying to understand why this really happens, but I can try to explain my findings as detailed as possible.

It doesn't happen on very fast devices, especially if the application is very light weight. It occurs when heavy load performed needs to be performed (huge lists, heavy application in general, maybe network delay paired with poor state management) or the calculating device is very weak. If one throttles the CPU via chrome devtools, the error will occur very often.

I have prepared a demo in the rx-angular demos section under Integrations/dnd. 

**Credits go to @SanderElias for doing first investigations and providing the original source code for the [demo-application](https://stackblitz.com/edit/rxforissue?file=src%2Fapp%2Fapp.component.ts)!!!**

I will link the PR where the demo is contained

Go the demo route, open chrome devtools, throttle your CPU and start dragging like crazy on the rxFor side.

# Environment

rx-angular demo repository

# Tasks to Resolve This

The first step is to be 100% about the root cause of the issue. The error message `Failed to execute 'insertBefore' on 'Node'` [,,,] `at ViewContainerRef.move (http://localhost:4200/rx-angular/demos/polyfills.js:24266:21)` already gives us a good hint. It seems to be the case that the `TemplateHandler#moveView` operation is performed while the actual DOM Node is in some kind of detached state.

After debugging the internals of `rxFor` and `@angular/cdk` `drag-ref.ts` simultaneously I finally understood what the problem is. The `DragRef` class inside `@angular/cdk/drag-drop` internally **replaces** the actual DOM node from the parent with a `placeHolderElement`. This happens on `mousemove` and `touchmove` event.

The image below gives a visual representation of the state in which the application is while dragging:

drag-drop behavior

While in this state, it is possible that `rxFor` processes new changes made to the `Observable` data tied to it. If it is performing a `move` operation on an item which is actually not part of the list which is maintained by `rxFor`, this error occurs and `viewContainerRef` gets out of sync.

Currently, this error results in an unrecoverable state. So we are definitely in need of fixing that.

After some more hours of debugging I want to share my thoughts on this issue and list potential ways to tackle it and things i have considered but are not working.

## reproduce the issue with `ngFor`

I want to reproduce the error with `ngFor` to showcase that this error is not exclusively related to rxFor. I suspect it is the scheduling of changes and asynchronous way of processing `changeDetection` which makes this issue more visible on `rxFor` than on `ngFor`. But still, if the underlying values produce a change that lead to a move operation where a currently detached DOM node is involved, the issue should be reproducable with `ngFor` + `async` pipe as well.

## use `noop` strategy

One of the early thoughts was to use the `noop` strategy while dragging since it would interrupt the process of `rxFor`. In theory it would work out, but after digging again into `RxListManager`, reality shows that it doesn't react to strategy changes in terms of cancelling an ongoing operation. This means it is still possible that the error occurs when a `move` operation was scheduled before.

While this seems like a possible hotfix for the issue, i am not :100: sure this is the way to go. It requires users of the `rxFor` to take care of it outside with custom code to glue `@angular/cdk/drag-drop` with `@rx-angular/template/rx-for`.
On the other hand, if we can prove this issue also exists for `ngFor` than `rxFor` is superior in terms of providing at least _a_ solution to the problem.

Still, if we want to this way, we need to make a change to `RxListManager`. It currently listens to the active strategy whenever a change to the observable is passed, not when the strategy changes.

```ts
withLatestFrom(strategyHandling$.strategy$)
```

The solution would be to use the strategy inside of a `combineLatest([values$, strategy$])` and cancel the latest operation on a strategy change. @BioPhoton what do u say to that? I could come up with a Demo for it.

## detect if a move operation fails upfront

Another approach could be to detect if a move operation will fail upfront. All information we need are stored in the `EmbeddedViewRef`. At least when Ivy is used.

```ts
// view.ts: https://github.com/angular/angular/blob/c9b47f21822343286d056fd70aec882acc0ae788/packages/core/src/render3/interfaces/view.ts#L57

/**
 * Size of LView's header. Necessary to adjust for it when setting slots.
 *
 * IMPORTANT: `HEADER_OFFSET` should only be referred to the in the `ɵɵ*` instructions to translate
 * instruction index into `LView` index. All other indexes should be in the `LView` index space and
 * there should be no need to refer to `HEADER_OFFSET` anywhere else.
 */
export const HEADER_OFFSET = 20;

```

This is the current implementation of `ListTemplateManager#moveView`:

```ts
// list-view-handler.ts
// the view we want to move to the new index
const oldView = viewContainerRef.get(oldIndex);
const view = <EmbeddedViewRef<C>>viewContainerRef.move(oldView, index);
    workFactory(() => {
      updateViewContext(item, view, {
        count,
        index,
      });
      view.detectChanges();
    });
```
The error happens if the current `EmbeddedViewRef` at `index` (new index `oldView` gets moved to) is replaced by a placeholder.

The potential solution would be to check upfront if the viewRef at `index` is part of it's parent or not. If i'm not mistaken, either `HEADER_OFFSET` will give us a hint about where to find the actual affected node. We can use that information to compare their respective parentElements and see if we can safely perform a move operation or not.

Solution sketch:

```ts
const nextView = <EmbeddedViewRef<C>>viewContainerRef.get(index);
const nextViewParent = getElementParent(nextView);
const oldViewParent = getElementParent(oldView);
if (nextViewParent !== oldViewParent) {
 // we cannot perform move... what to do now?!
}
// do the move
```

This raises a lot of questions:
* how to react to this state?
* Re-schedule this operation?
* Mark it as undone and perform later?
* Cancel everything and re-start the process?

# Other issues i've stumbled upon

While debugging `rxFor` i've came accross some other potential bugs/performance issues which we should take a look at.

## Over-rendering of embedded views

In `ListTemplateManager`s `getListChanges` operation we mimic the behavior of `ngFor`, but instead of processing everything directly, we just decide the tasks which should get scheduled for later.

I've found out that there are possible scenarios where `ListTemplateManager` schedules multiple consequent operations for the same `ViewRef`. This is unintentional.

```ts
// list-view-handler.ts: https://github.com/rx-angular/rx-angular/blob/15103393fa0700d2fd795703f529fe2a7c15b7c4/libs/cdk/src/lib/template-management/list-view-handler.ts#L156

 changes.forEachIdentityChange((record) => {
    const item = record.item;
    changesArr.push(getUpdateChange(item, record.currentIndex));
    changedIdxs.add(item);
  });

```

Basically, only a check is missing here:

```ts
if (!changedIdxs.has(item)) {
      changesArr.push(getUpdateChange(item, record.currentIndex));
      changedIdxs.add(item);
    }
```

hoebbelsB:
## Description

* [x] implement strategy-switch on the fly in list-template-manager
* [x] implement demo for strategy switch with rxFor
* [x] reproduce drag n drop error with `ngFor`
* [ ] any other wishes? :)

In the end I can prove that the error is in fact not a direct problem of `rxFor` or `ngFor`, but instead it's an issue caused by `@angular/cdk/drag-drop` because it is manipulating the DOM where those directives try to process changes.

## Demo application

image

You will find 5 different cases in the demo.

* ngFor big list
* ngFor errorItems
* rxFor big list
* rxFor errorItems
* rxFor errorItems + noop-strategy
* rxFor big list + noop-strategy 

### Big lists

The **big list** demos have besides the thread-blocking (added by @SanderElias) no special implementation. Both variants (ngFor + rxFor) can eventually hit an error scenario. `ngFor` won't complain as much, it's much harder to hit the error. `rxFor` will work fine on normal usage as well, as soon as you start going crazy, you should easily see the broken view state.

### Error items

The error items test the case I already discussed in #831 . If a move operation is triggered while a user is dragging, the view will crash.
Turned out, it was very easy to implement. As soon as u start dragging the first item, the connected item-array will be shuffled.
Both directives, `ngFor` and `rxFor` experience the very same error at this stage.

`ngFor` error as video:

https://user-images.githubusercontent.com/4904455/124591991-41ab4400-de5d-11eb-93c2-109fff6a52b7.mp4

### noop strategy

As already discussed in #831 the idea is to cancel any scheduled operations when a new strategy is set to `rxFor`. Thats what I have implemented and can now showcase for both demos **big list** and **errorItems**. Both scenarios won't ever rais this error again when one switches to the `noop` strategy while a user is dragging.

fixes #831


hoebbelsB:
@BioPhoton what should I add so that we can merge it?

PR #838, opened 2021-07-10
feat(state): decouple `internals` into a lib

arturovt:
Moved out the stuff from the `state` library that is not publicly exposed for our users but is meant to be used internally. This now will be possible to split other stuff into libraries since the `internals` package is re-used between different modules.


edbzn (approved):

BioPhoton:
As we switched from separate nx libs to just sub-package folders.
Therefore we, unfortunately, have to close this PR. 

If there is something we still can reuse please let me know. 
Sry we have to drop this even if u put in  time to create it.

PR #901, opened 2021-08-13
fix(state): improve toDictionary & extract performance

Karnaukhov-kh:
10 items (11KB JSON)
image

100 items (105KB JSON)
image

500 items (526KB JSON)
image

1000 items (1049KB JSON)
image

5000 items (5276KB JSON)
image

8000 items (8443KB JSON)
image

### DIFF
--- a/libs/state/spec/transformation-helpers/array/toDictionary.spec.ts
+++ b/libs/state/spec/transformation-helpers/array/toDictionary.spec.ts
@@ -67,6 +67,12 @@ describe('toDictionary', () => {
       expect(dictionaryResult).toEqual(dictionaryByNumber);
     });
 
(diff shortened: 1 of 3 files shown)


libs/state/spec/transformation-helpers/array/toDictionary.spec.ts
```diff
@@ -67,6 +67,12 @@ describe('toDictionary', () => {
       expect(dictionaryResult).toEqual(dictionaryByNumber);
     });
 
+    it('should change the refence', () => {
```
Karnaukhov-kh, line 70:
```suggestion
    it('should change the reference', () => {
```

PR #907, opened 2021-08-16
chore: workspace & rxjs@7.x upgrade

hoebbelsB:
## Description

* upgraded angular to latest version
* upgraded nx workspace to latest version
* upgraded rxjs to 7.x => fixed all incompatibilities. some deprecations are still existent


BioPhoton (changes requested):
We should also update the dependencies in the libs to v7?
https://github.com/rx-angular/rx-angular/blob/master/libs/state/package.json#L52

hoebbelsB:
@BioPhoton are those actually valid? Why does the `cdk` package not have the peerDepedency to rxjs? Would the dependency to `@angular/core` be sufficient since it already has a peerDependency to `rxjs`? and is version `> 8.0.0` still valid?

libs/state/package.json
```diff
@@ -49,7 +49,7 @@
   },
   "peerDependencies": {
     "@angular/core": ">8.0.0",
-    "rxjs": "^6.5.5"
+    "rxjs": ">6.5.5"
```
BioPhoton, line 52:
```suggestion
```

PR #942, opened 2021-09-09
chore(cdk): split missing sub-packages

hoebbelsB:
# Description

moved and split missing sub-packages:

* render-strategies
* template
* state

one thing that I am not sure about is that the config now lives in `render-strategies`. It makes sense, since the config is all about configuring rendering-strategies. Thats why I would suggest rename the token from `RX_ANGULAR_CONFIG`  to `RX_STRATEGY_CONFIG` .


edbzn (approved):
Another way would be to deprecate the imports, remove the compat layer afterward and (optionally) provide migrations. I don't think it's too much effort to create migrations since it's mostly copy pasta.

I would recommend this way since it follows a deprecation convention which is good for the users.

hoebbelsB:
@edbzn how can we deprecate the exports? I just tried 
```ts
/** @deprecated use @rx-angular/cdk/coalescing */
export * from '@rx-angular/cdk/coalescing';
```
in the index.ts file, but it doesn't seem to have any effect

edbzn:
@hoebbelsB Here is an example of how we made it, I think we should use named exports : 

https://github.com/jscutlery/test-utils/blob/main/packages/cypress-mount/src/index.ts

hoebbelsB:
> @hoebbelsB Here is an example of how we made it, I think we should use named exports :
> 
> https://github.com/jscutlery/test-utils/blob/main/packages/cypress-mount/src/index.ts

@edbzn thanks a lot! I have marked all exports as deprecated now

PR #946, opened 2021-09-13
perf: coalesce ngZone

hoebbelsB:
# Description

When views get rendered with either `rxFor` , `rxLet`, `push` or even `RxStrategyProvider` , the task to render the update of the specific view gets scheduled into a queue order to maintain a fluent user experience. Since the scheduler is configured to maintain consistent **60fps** (or ~**10ms** (16-6ms) per run) it happens that more than one task is executed synchronously one after the other. The more tasks can fit into execution cycle (or frame), the better. That's why we should always try to keep the synchronously executed javascript code at a minimum level. 

Currently every scheduled rendering task also contains the instruction to be patched into `NgZone` . This results in an unnecessary bloat for every scheduled rendering task.

image

image
You can see how `onLeave` actually eats up more than **50%** of the total frame budget.

# Goal

The main problem to solve is that we call `NgZone#run` for each task and not for each batch of tasks executed during one `yieldInterval`. This reduces the amount scripting time executed by `NgZone` and improves the rendering performance.

image

image

# Solution

## Hybrid (recommended)

Based on the [idea](https://github.com/rx-angular/rx-angular/pull/946#pullrequestreview-758842767) of @arturovt to attach `NgZone` directly to the scheduled task, I have come up with a solution that (besides some _edge-cases_) delivers **imho** the best compromise between configurability, performance and stability (multiple NgZones).

**Attach `NgZone` to scheduled task**

Extend `ReactSchedulerTask` with a field to attach something `NgZone`like to it. All what we need is the `run` method, which is why I have just mocked it like that.

```ts
export interface SchedulerTaskZone {
  run<T>(fn: (...args: any[]) => T): T;
}

export interface ReactSchedulerTask {
  id: number;
  sortIndex: number;
  callback: Function;
  priorityLevel: PriorityLevel;
  startTime: number;
  expirationTime: number;
  isQueued?: boolean;
  ngZone?: SchedulerTaskZone;
}
```

Extend `RxRenderBehavior` to be aware of `NgZone`, since this is the place where we hand over configuration to the `scheduler` itself.

```ts
export type RxRenderBehavior = <T = unknown>(
  work: any,
  scope?: coalescingObj,
  ngZone?: NgZone
) => (o: Observable<T>) => Observable<T>;
```

Hand over `NgZone` to `scheduler`.

```ts
// concurrent-strategies.ts 
const normalStrategy: RxStrategyCredentials = {
  name: 'normal',
  work: (cdRef) => cdRef.detectChanges(),
  behavior: (work: any, scope: any, ngZone: NgZone) => {
    return (o$) =>
      o$.pipe(
        scheduleOnQueue(work, { ngZone, priority: PriorityLevel.NormalPriority, scope })
      );
  },
};
```

```ts
// scheduler.ts

interface ScheduleCallbackOptions {
  delay?: number;
  ngZone?: SchedulerTaskZone;
}

function scheduleCallback(
  priorityLevel: PriorityLevel,
  callback: VoidFunction,
  options?: ScheduleCallbackOptions
): ReactSchedulerTask {
  // some code
  const newTask: ReactSchedulerTask = {
    id: taskIdCounter++,
    callback,
    priorityLevel,
    startTime,
    expirationTime,
    sortIndex: -1,
    ngZone: options?.ngZone || defaultZone
  };
}
```

**Coalesce NgZone calls, let tasks run in _their_ NgZone**

Now that we have everything setup to interact with `NgZone` on a per task-basis, the last job is to cause a minimal amount of `NgZone#run` calls.

For this, we can slightly adjust the `workLoop` function like so:

```
const zone = currentTask.ngZone;
zone.run(() => {
  while isSame(ngZone) && hasTask { 
    executeTask();
  }
});
if (zoneChanged && hasTask) {
 return workLoop(); // restart workLoop if zone changed
}
```

With this logic in place, all tasks with the same `NgZone` will be executed during the same `run` until a new task with another `NgZone` configuration needs to be executed.

**What does it mean?**
In most cases the loop won't get restarted, since the majority of tasks should be patched by NgZone in a regular angular environment. The directive/implementations which will benefit the most from it are `*rxFor` respectively `ListTemplateManager`.

**Edge cases**
While it will 100% work as before, there are very rare cases where the performance is simply unaffected. This would be a scenario where multiple tasks each with a different, or alternating, ngZone configuration are getting processed.
In reality, most of cases will benefit from this approach by it's full potential. 

## NgZone wrap (not recommended)

The initial idea was to just wrap `NgZone#run` around `workLoop`. This approach would 100% fix all performance issues and would lead to the desired outcome, **but**...
It also means, `patchZone` is no longer a local configuration. It can only be configured once and would then be valid for all scheduled tasks. This means that the `patchZone` option on all directives is obsolete and there is no way to schedule a task outside of `NgZone`.

This also means that all `RxStrategyProvider` api's are zone-full when `patchZone` is set to true.

An idea to still have the possibility to use the scheduler without NgZone is to maintain a second queue of tasks which is not patched by NgZone. We could consider this queue as non-render-work related queue and always prioritize the tasks in the patched queue. But this is just an initial thought and I have no idea what other impacts this would lead to.

## Comparison

**Before:**
image

image

**After:**
image

image

I would be very happy about feedback to this issue @edbzn @arturovt @BioPhoton @Karnaukhov-kh !


libs/cdk/internals/scheduler/src/lib/scheduler.ts
```diff
@@ -1,6 +1,6 @@
 // see https://raw.githubusercontent.com/facebook/react/master/packages/scheduler/src/forks/SchedulerDOM.js
 
-import { ɵglobal } from '@angular/core';
+import { ɵɵdirectiveInject as inject, NgZone, ɵglobal } from '@angular/core';
```
edbzn, line 3: Minor: `ɵɵdirectiveInject as inject` is imported but unused.

arturovt (changes requested):
Hey gentlemen!

I've checked the code. There's an issue with the local variable, I mean `let ngZone`. The issue will be in a microfrontend environment, e.g. 2 Angular apps, that share the code, are bootstrapped, then the second app will override the `ngZone` variable and the first app will be using the wrong zone.

I have some thoughts... we export the `scheduleCallback` function within the scheduler code, we could extend the `ReactSchedulerTask` and add the `ngZone` property there, so each scheduled task will also have a bound `NgZone`:

```ts
function scheduleCallback(
  priorityLevel: PriorityLevel,
  callback: VoidFunction,
  ngZone: NgZone, // 👈👈
  options?: ScheduleCallbackOptions
): ReactSchedulerTask {
  const newTask: ReactSchedulerTask = {
    id: taskIdCounter++,
    ngZone, // 👈👈
    callback,
    priorityLevel,
    startTime,
    expirationTime,
    sortIndex: -1,
  };
```

So the `workLoop` function will not have to rely on the global variable:

```ts
function workLoop(hasTimeRemaining, initialTime) {
  let currentTime = initialTime;
  advanceTimers(currentTime);
  currentTask = peek(taskQueue);
  currentTask.ngZone.run(() => {
```

> P.S. these are some random thoughts and I can't check if it'll work or not. Just sharing my ideas with you!

libs/cdk/internals/scheduler/src/lib/scheduler.ts
hoebbelsB: ty

hoebbelsB:
hi @arturovt ty for your thoughts!


> Hey gentlemen!
> 
> I've checked the code. There's an issue with the local variable, I mean `let ngZone`. The issue will be in a microfrontend environment, e.g. 2 Angular apps, that share the code, are bootstrapped, then the second app will override the `ngZone` variable and the first app will be using the wrong zone.

never thought about that... without my changes this won't be a problem since NgZone is embedded in the `work` and controlled on a per directive level.

> 
> I have some thoughts... we export the `scheduleCallback` function within the scheduler code, we could extend the `ReactSchedulerTask` and add the `ngZone` property there, so each scheduled task will also have a bound `NgZone`:
> 
> ```ts
> function scheduleCallback(
>   priorityLevel: PriorityLevel,
>   callback: VoidFunction,
>   ngZone: NgZone, // 👈👈
>   options?: ScheduleCallbackOptions
> ): ReactSchedulerTask {
>   const newTask: ReactSchedulerTask = {
>     id: taskIdCounter++,
>     ngZone, // 👈👈
>     callback,
>     priorityLevel,
>     startTime,
>     expirationTime,
>     sortIndex: -1,
>   };
> ```

that is definitely more beautiful compared to the current implementation and would at least reduce the complexity of the template-managers.

> So the `workLoop` function will not have to rely on the global variable:
> 
> ```ts
> function workLoop(hasTimeRemaining, initialTime) {
>   let currentTime = initialTime;
>   advanceTimers(currentTime);
>   currentTask = peek(taskQueue);
>   👉 currentTask.ngZone.run(() => {
> ```
> 

hmm, now you will probably end up with the issue you mentioned first. `currentTask` changes during the iteration. The reference to `NgZone` could change as well while iterating. But the idea is neat and maybe we can do something with it.

The main problem I was trying to solve with this PoC is that we call `NgZone.run` for each task and not for each batch of tasks. The downside is that it's now impossible to let something run outside of zone. The idea of maintaining a second queue of tasks for unpatched work would just lead to a huge problem with priorities...

So maybe we can find a somewhat optimized hybrid solution. I try to explain it in pseudo-code:

```
let ngZone = currentTask.ngZone;

while (exists(currentTask) && !zoneChanged) {
  if (limit) {
    // currentTask hasn't expired, and we've reached the deadline
    break; 
  }
  currentTask.execute();
  currentTask = peek(taskQueue);
  zoneChanged = currentTask.ngZone !== ngZone;
}

if (zoneChanged) {
  // let workLoop run again with the currentTask and next ngZone configuration
  workLook(currentTask);
}

```

What I am trying here is to let the workloop run until there is a different NgZone setup => than run again. In best case all scheduled tasks have the same NgZone configuration and thus running in the same `.run()` environment. For scheduled tasks via `rxFor` this would 100% pay off!

let me hear what u think :)

hoebbelsB:
The question is, how do we get `NgZone` into the configured task? The only way I can think of is via `RxStrategyCredentials#behavior`. We could extend it:

```ts
export type RxRenderBehavior = <T = unknown>(
  work: any,
  scope?: coalescingObj,
  ngZone?: NgZone
) => (o: Observable<T>) => Observable<T>;

export interface RxStrategyCredentials<S = string> {
  name: S;
  work: RxRenderWork;
  behavior: RxRenderBehavior;
}
```

hoebbelsB:
okay ffs, i've just implemented it... works like a fucking charm! any thoughts???

BioPhoton:
We could think of refactoring the renderBehavior typing to and config obj instead of single params?

hoebbelsB:
> We could think of refactoring the renderBehavior typing to and config obj instead of single params?

@BioPhoton agree this would be the more elegant solution. But it would also be a breaking change. I guess we would need to deprecate the first overload and introduce the `NgZone` parameter with the new config object style.

edbzn:
cdk is still in alpha, I don't think we really need to be strict with breaking changes.

PR #968, opened 2021-10-03
fix(schematics): remove `@nrwl/workspace` dependency

edbzn:
Fix migration script for regular Angular CLI workspaces by removing Nx dependency.


hoebbelsB:
hi @edbzn i've checked it in different projects (nx & ng workspaces). the migration runs fine.

image

I could just notice some minor formatting issue. It looks like the migration would magically remove `;` :

image

edbzn:
@hoebbelsB It's fixed, and ensured using snapshot testing; originally not a fan of snapshot testing, but in this case, it catches missed bugs pretty well.

hoebbelsB:
@edbzn we just had an interesting talk about this issue with @vmasek. I guess we shouldn't make things to complicated and just remove the dependency to nrwl. Formatting in the end can be 'broken' if we can at least guarantee the code is working.
`@angular/material` e.g is putting just empty new lines, which will most probably not best formatting practice either. So it's up to the end user to format it properly.

edbzn:
@hoebbelsB Agree, but I would let the formatting as it is now, I think it's better with it and it does not throw if Prettier is not installed. In the end, I would prefer migrating to the nx-devkit which also supports regular Angular CLI projects and provides better utils. 

What do you think?

PR #987, opened 2021-10-11
test(cdk): add basic tests for coerceObservableWith

### LINKED ISSUE #969: test(cdk-coercing): implement tests for coerceObservableWith
# Description

implement tests for the `coerceObservableWith` OperatorFunction. (libs/cdk/coercing/src/lib/coerceObservableWith.ts)
The function should map any `Observable` or non-observable value to an `Observable`.
Feel free to some jsDocs as well so docs get automatically generated for the operator.

e.g.
```ts
const animals$ = ['cat', 'dog', 'pig']
animals$.pipe(
 coerceObservableWith()
).subscribe(console.log);

// output: [Observable, Observable, Observable]

```

Good example for other marble test implementations are:
* [coalesceWith](https://github.com/rx-angular/rx-angular/blob/master/libs/cdk/coalescing/spec/coalesceWith.spec.ts)
* [onStrategy](https://github.com/rx-angular/rx-angular/blob/master/libs/cdk/spec/render-strategies/onStrategy.spec.ts)

miqh:
Hello there.

I saw #969 and thought I'd try contribute some basic tests to cover the `coerceObservableWith` operator.

Happy to amend with more cases if need be (I don't have a deep knowledge of RxJS), but I think these are a good start.

fixes #969

### DIFF
--- a/libs/cdk/coercing/spec/coerceObservableWith.spec.ts
+++ b/libs/cdk/coercing/spec/coerceObservableWith.spec.ts
@@ -1,5 +1,5 @@
 // tslint:disable:nx-enforce-module-boundaries
-// eslint:disable:nx-enforce-module-boundaries
+/* eslint-disable nx-enforce-module-boundaries */
 import { coerceObservableWith } from '@rx-angular/cdk';
 import { jestMatcher, mockConsole } from '@test-helpers';
 import { of } from 'rxjs';


hoebbelsB:
@miqh can u merge in latest master branch?

hoebbelsB:
@miqh looks like you need to disable some eslint rules for the spec files:

The ci fails because of:

```
ERROR: 1:1  nx-enforce-module-boundaries  Projects should use relative imports to import from other files within the same project. Use "./path/to/file" instead of import from "@rx-angular/cdk"
ERROR: 2:1  nx-enforce-module-boundaries  buildable libraries cannot import non-buildable libraries
```

u probably need to add something like `eslint-disable @nrwl/nx-enforce-module-boundaries` as comment

hoebbelsB:
u can take a look here, where we faced the same issue: https://github.com/rx-angular/rx-angular/pull/998

hoebbelsB:
> @hoebbelsB, [e7afe70](https://github.com/rx-angular/rx-angular/commit/e7afe70577f059b82ad95298007627711a477fb3) should address the [ESLint suppression format](https://eslint.org/docs/user-guide/configuring/rules#using-configuration-comments-1).

ty, looks great!

> Also, still out of curiosity, I don't think you answered my question about what's actually required to ensure that local linting attempts mirror what happens on NxCloud.

> As I pointed out in an earlier comment, I tried running `nx run cdk:lint` even before I added the extra lint rules, but I didn't see any lint errors come up locally. Not a big deal, but it might be something other outside contributors might run into too, which is why I thought I'd draw attention to it.

`yarn nx run cdk:lint` should be fine to do what happens in CI. I can confirm locally that `yarn nx run cdk:lint` runs into an error when removing the comment for disabling `tslint`.

PR #1034, opened 2021-10-24
refactor: remove cdk compat-layer, introduce migrations

hoebbelsB:
# Description

* [x] move `template/schematics/common` to `tools/generators/utils`
* [x] implement cdk migration schematic
* [x] implement tests for cdk migration schematic


libs/cdk/schematics/migration.json
```diff
@@ -0,0 +1,10 @@
+{
+  "$schema": "../../../node_modules/@angular-devkit/schematics/collection-schema.json",
+  "schematics": {
+    "update-1-0-0": {
```
edbzn, line 4: I think we should name it `1.0.0-alpha.12`.

libs/cdk/schematics/src/migrations/update-1-0-0/index.spec.ts
```diff
@@ -0,0 +1,372 @@
...
+import * as path from 'path';
+
+describe('cdk migration 1.0.0', () => {
+  let appTree: UnitTestTree;
+
+  /* Increase Jest timeout because of long living tests. */
+  beforeAll(() => jest.setTimeout(10_000));
+
+  /* Reset default timeout to 5s. */
+  afterAll(() => jest.setTimeout(5_000));
```
edbzn, line 15: :confused: I don't know if it's worth investing time to profile the migration to remove this. What do you think?

libs/template/schematics/src/commands/ng-add/index.spec.ts
```diff
@@ -4,7 +4,7 @@ import {
 } from '@angular-devkit/schematics/testing';
 import { join } from 'path';
 
-import { readJsonInTree } from '../../common';
+import { readJsonInTree } from '../../../../../../tools/generators/common';
```
edbzn, line 7: We can not use path mapping here?

libs/cdk/schematics/src/migrations/update-1-0-0/index.spec.ts
hoebbelsB: what do you mean? to improve the tests? i won't :D

libs/template/schematics/src/commands/ng-add/index.spec.ts
hoebbelsB: no

hoebbelsB:
@edbzn i've noticed there are new test cases which are failing now. Did u already started working on how to solve them? For now we can only fix imports, but no usages? Or do I miss something?

libs/cdk/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@rx-angular/cdk",
-  "version": "1.0.0-alpha.11",
+  "version": "1.0.0-alpha.12",
```
edbzn, line 3: @hoebbelsB we should not update the lib manually on a feature branch, I revert this.

.github/workflows/build-and-test.yml
```diff
@@ -104,7 +104,8 @@ jobs:
         run: yarn nx run ssr-e2e:e2e --headless
       # don't run build in parallel, can cause dead locks
       - name: Run Affected build
-        run: yarn nx affected:build
+      # @todo: @edbzn Nx cloud fail, disable it temporary.
+        run: NX_CLOUD_DISTRIBUTED_EXECUTION=false yarn nx affected:build
```
hoebbelsB, line 108: why? :(

hoebbelsB: and if we change the github actions, can we maybe disable `NX_VERBOSE_LOGGING` as well?

apps/demos/src/app/features/experiments/structural-directives/if-visible-poc/if-visible.component.ts
```diff
@@ -1,5 +1,4 @@
-import { ChangeDetectionStrategy, Component } from '@angular/core';
-import { RxStrategyProvider } from '@rx-angular/cdk';
+import { ChangeDetectionStrategy, Component } from '@angular/core';;
```
hoebbelsB, line 1:
```suggestion
import { ChangeDetectionStrategy, Component } from '@angular/core';
```

libs/cdk/project.json
```diff
@@ -25,6 +25,45 @@
...
+      "dependsOn": [
+        {
+          "target": "build-base",
+          "projects": "self"
+        },
+        {
+          "target": "build-schematics",
+          "projects": "self"
+        }
+      ],
```
hoebbelsB, line 38: we could configure this as general taskDependency in `nx.json` what do u think?

libs/template/project.json
```diff
@@ -36,16 +42,18 @@
...
+      "dependsOn": [
+        {
+          "target": "build-lib",
+          "projects": "self"
+        },
+        {
+          "target": "build-schematics",
+          "projects": "self"
+        }
+      ],
```
hoebbelsB, line 54: we could configure this as general taskDependency in nx.json what do u think?

hoebbelsB (commented):
@edbzn big THX for jumping into this! tried in another project, works great!

libs/cdk/project.json
edbzn: There's an issue with Nx, if I change `cdk:build-base`  to `template:build-lib`  which is the same as the `template:build-lib` then it breaks the build, we cannot use global dependsOn config for now.

PR #1045, opened 2021-10-28
Create RxClass directive for push-based style changes

### LINKED ISSUE #1041: Create RxClass directive for push-based style changes
# The RxClass directive provides a way to make the style changes push-based

When using `ngClass` or `[class]` directive with an Observable value and the `async` pipe in the template of an Angular application, the `async` pipe relies on the Zone to be present - it doesn't really trigger change detection by itself. It marks the component and its children as dirty waiting for the Zone to trigger change detection. So, in case you want to create a zone-less application, the `async` pipe won't work as desired. 

# Solution

The `RxClass` directive comes with its own strategies bound to `RxClassStrategy` directive to manage change detection every time a new class value is sent from the bound Observable.

svoren258:
Closes #1041.

### DIFF
--- a/libs/template/class/src/lib/class.directive.ts
+++ b/libs/template/class/src/lib/class.directive.ts
@@ -1,0 +1,237 @@
+import {
+  Directive,
+  ElementRef,
+  Input,
+  IterableChanges,
+  IterableDiffer,
+  IterableDiffers,
+  KeyValueChanges,
+  KeyValueDiffer,
+  KeyValueDiffers,
+  OnDestroy,
+  OnInit,
+  Renderer2,
+  ɵisListLikeIterable as isListLikeIterable,
+} from '@angular/core';
+import {
+  accumulateObservables,
+  RxStrategyNames,
+  RxStrategyProvider,
+} from '@rx-angular/cdk';
+import { coerceAllFactory } from '@rx-angular/cdk/coercing';
+import {
+  BehaviorSubject,
+  isObservable,
+  Observable,
+  ObservableInput,
+  ReplaySubject,
+  Subscription,
+} from 'rxjs';
+import { map, switchAll, switchMap, withLatestFrom } from 'rxjs/operators';
+
+type NgClassValues =
+  | string[]
+  | Set<string>
+  | Record<string, boolean>
+  | string
+  | null
+  | undefined;
+
+type RxClassInput =
+  | Observable<NgClassValues>
+  | Record<string, Observable<boolean>>;
+
+@Directive({ selector: '[rxClass]' })
+export class ClassDirective implements OnInit, OnDestroy {
+  private iterableDiffer: IterableDiffer<string> | null = null;
+  private keyValueDiffer: KeyValueDiffer<string, any> | null = null;
+  private initialClasses: string[] = [];
+  private rawClass: NgClassValues = null;
+  private newInput = false;
+  private sub = new Subscription();
+
+  constructor(
+    private readonly iterableDiffers: IterableDiffers,
+    private readonly keyValueDiffers: KeyValueDiffers,
+    private readonly ngEl: ElementRef,
+    private readonly renderer: Renderer2,
+    private readonly strategyProvider: RxStrategyProvider
+  ) {}
+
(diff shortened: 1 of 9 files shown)


libs/template/class/src/lib/class.directive.ts
```diff
@@ -0,0 +1,237 @@
...
+    private readonly strategyProvider: RxStrategyProvider
+  ) {}
+
+  private rxClass = coerceAllFactory<NgClassValues>(
+    () => new ReplaySubject(1),
+    switchAll()
+  );
+
+  @Input('rxClass')
+  set rxClass$(value$: RxClassInput) {
```
BioPhoton, line 67: Let's make it also accept static values. There is a PR that intros that already for let and push.
Should be merged already into main.

libs/template/class/src/lib/class.directive.ts
```diff
@@ -0,0 +1,237 @@
...
+    this.sub.add(
+      this.rxClass.values$
+        .pipe(
+          withLatestFrom(this.strategy.values$),
+          switchMap(([value, strategy]) =>
+            this.strategyProvider.schedule(
+              () => {
+                if (this.newInput) {
+                  this.removeClasses(this.rawClass);
+                  this.applyClasses(this.initialClasses);
```
BioPhoton, line 98: Why do we need to run apply here? The Initial Classes are empty so the functions will noop.
Is this maybe related to the classes already existing in the DOM as static strings?

svoren258: I double-checked that and it looks like we don't really need to do this cleanup (most probably just legacy code I forgot to remove).

libs/template/docs/api/class-directive.md
```diff
@@ -0,0 +1,65 @@
+## ClassDirective
+
+The `rxClass` directive serves a convenient way of using observables for style (class) changes.
+
+Under the hood, it leverages a `RenderStrategy` which in turn takes care of optimizing the change detection
+of your component by passing wanted strategy to `rxClassStrategy` (the default strategy is `'immediate'`). The `ClassDirective` will apply style (class) changes to the template and manage change detection after it got an initial value.
```
hoebbelsB, line 6: the part about the default strategy is deprecated since we decided to go with `defaultStrategy`? Though I think having normal as defaultStrategy for setting classes is not the best approach. But we can target this when we refactor the configuration interface for rx-angular.

hoebbelsB (commented):
hi @svoren258 ty for the contribution, good to see you here :partying_face: 

Since we have it in all the other directives, could we maybe add an `@Output() renderCallback = new Subject()` to `RxClass` as well?

libs/template/class/src/lib/class.directive.ts
svoren258: That would be really cool. The issue here is, it has to wait for refactoring of `accumulateObservables` which doesn't accept static values at the moment and is used in the current `RxClass` implementation.

image

hoebbelsB: no, we need to create an issue and work on it.

svoren258:
> Nice PR, love it! I left a few comments. I would also suggest another thing:
> 
> * adding server-side tests to ensure it renders the classes correctly

@edbzn, thank you for the review. Do you have any references for the server-side tests? Some inspiration would be great.

edbzn:
@svoren258 There are examples in [`apps/ssr-e2e/src/integration/app.spec.ts`](https://github.com/rx-angular/rx-angular/blob/master/apps/ssr-e2e/src/integration/app.spec.ts), I can take care of it if you want.

svoren258:
> @svoren258 There are examples in [`apps/ssr-e2e/src/integration/app.spec.ts`](https://github.com/rx-angular/rx-angular/blob/master/apps/ssr-e2e/src/integration/app.spec.ts), I can take care of it if you want.

I am not 100% familiar with this type of test but it looks doable. I could try to implement that but feel free to do it if you like. :)

svoren258:
This PR continues here: https://github.com/rx-angular/rx-angular/pull/1048

PR #1048, opened 2021-11-02
Create RxClass directive for push based changes

### LINKED ISSUE #1041: Create RxClass directive for push-based style changes

svoren258:
Closes #1041. This PR is an updated copy of #1045.


libs/template/class/src/lib/class.directive.ts
```diff
@@ -0,0 +1,197 @@
...
+              { strategy, scope: this, patchZone: false }
+            )
+          )
+        )
+        .subscribe()
+    );
+  }
+
+  private applyChanges(): void {
+    console.log(this.iterableDiffer);
```
edbzn, line 132: Not sure if it was ommited.

libs/template/class/src/lib/class.directive.ts
```diff
@@ -0,0 +1,197 @@
...
+    this.sub.add(
+      this.rxClass.values$
+        .pipe(
+          withLatestFrom(this.strategy.values$),
+          switchMap(([value, strategy]) =>
+            this.strategyProvider.schedule(
+              () => {
+                this.classValueChanged = false;
+                this.rawClass = value;
+                if (this.newInput) {
```
edbzn, line 102: Why do we need these `newInput` + `classValueChanged` variables? reacting to `rxClass` observable events to update DOM values would be enough, am I missing something?

libs/template/class/src/lib/class.directive.ts
```diff
@@ -0,0 +1,197 @@
...
+  );
+
+  @Input('rxClassStrategy')
+  set strategy$(
+    strategy: RxStrategyNames<string> | Observable<RxStrategyNames<string>>
+  ) {
+    this.strategy.next(strategy);
+  }
+
+  @Output() readonly renderCallback = new Subject<void>();
```
Totati, line 90: I'd like to chime in with `rxLet` directive, and notice that it has a [`rendered` output](https://github.com/rx-angular/rx-angular/blob/master/libs/template/let/src/lib/let.directive.ts#L363). I think these outputs should have a naming convetion with a prefix or something, or be consistent as Angular can handle [outputs with the same name](https://stackblitz.com/edit/angular-ivy-x25zxw?file=src%2Fapp%2Fapp.component.html).

libs/template/class/src/lib/class.directive.ts
svoren258: The `classValueChanged` is there because of the `renderCallback` `@Output`. The `newInput` property is used to initialize the `differs` using the first value passed at the very beginning. I will rewrite that a bit so that it will be easier to read. The `differs` behaviour is kind of unusual to me as well at some point tbh 😄

libs/template/class/src/lib/class.directive.ts
```diff
@@ -0,0 +1,198 @@
...
+  );
+
+  // eslint-disable-next-line @angular-eslint/no-input-rename
+  @Input('rxClass')
+  set rxClass$(value$: RxClassInput) {
+    this.newInput = true;
+    this.rxClass.next(
+      (isObservable(value$) ? value$ : accumulateObservables(value$)).pipe(
+        map((value) => (typeof value === 'string' ? [value] : value))
+      )
```
hoebbelsB, line 72: why do we use `accumulateObservables` on a non-observable value?

libs/template/class/src/lib/test/class.directive.spec.ts
```diff
@@ -0,0 +1,166 @@
...
+  readonly setClassTwo$ = of(false);
+}
+
+describe('rxClass', () => {
+  let fixture: ComponentFixture<TestComponent>;
+
+  beforeEach(() => {
+    fixture = TestBed.configureTestingModule({
+      declarations: [TestComponent, ClassDirective],
+      schemas: [CUSTOM_ELEMENTS_SCHEMA]
```
hoebbelsB, line 74: isn't it easier to just provide the primary strategy as native here instead of setting native for each directive instance?

libs/template/docs/api/class-directive.md
```diff
@@ -0,0 +1,65 @@
...
+    <ng-container *rxLet="hero$; let hero; strategy: strategy">
+      <app-hero [hero]="hero"></app-hero>
+    </ng-container>
+  `
+})
+export class AppComponent {
+  readonly isVisible$ = new BehaviorSubject<boolean>(true);
+
+  readonly singleClass$ = new BehaviorSubject<string>('class-one');
+  readonly setOfClasses$ = new BehaviorSubject<string[]>(['class-one', 'class-two']);
```
hoebbelsB, line 59: this is an array of classes :laughing: a set is missing :)

libs/template/docs/api/class-directive.md
```diff
@@ -0,0 +1,65 @@
+## ClassDirective
+
+The `rxClass` directive serves a convenient way of using observables for style (class) changes.
+
+Under the hood, it leverages a `RenderStrategy` which in turn takes care of optimizing the change detection
+of your component by passing wanted strategy to `rxClassStrategy`. The `ClassDirective` will apply style (class) changes to the template and manage change detection after it got an initial value.
+So if the incoming `Observable` emits its value lazily (e.g. data coming from `Http`), your style/class will be
+applied lazily as well. This can very positively impact the render performance of your application.
```
hoebbelsB, line 8:
```suggestion
Under the hood, it leverages a `RenderStrategy` which in turn defines the timing when incoming changes get applied to the directives `ElementRef#nativeElement`. The key difference to the native `ngClass` directive and the main reason to use `rxClass` over `ngClass` is its runtime performance. Due to its push-based nature, the `ClassDirective` is able to apply classes defined as input to its `nativeElement` without the need of running `ChangeDetection` or waiting for `NgZone` to kick in.

For this, it needs an `Observable` as input source.

```typescript
class$ = state$.pipe(
  map(state => state.isActive ? 'active-class' : 'disabled-class' )
)
```

```html
<div [rxClass]="class$"></div>
```

libs/template/docs/api/class-directive.md
```diff
@@ -0,0 +1,65 @@
...
+### Problems with `async` and `ngClass` or `class`
+
+In Angular, a way of binding an observable to the `ngClass` or `class` directive could look like that:
+
+```html
+<div [ngClass]="{ 'visible': isVisible$ | async }"></div>
+<div [class.visible]="isVisible$ | async"></div>
+```
+
+When using `ngClass` or `class` directive with an Observable value and the `async` pipe in the template of an Angular application, the `async` pipe relies on the Zone to be present - it doesn't really trigger change detection by itself. It marks the component and its children as dirty waiting for the Zone to trigger change detection. So, in case you want to create a zone-less application, the `async` pipe won't work as desired.
```
hoebbelsB, line 19:
```suggestion
When using the native `ngClass` or `class` directive, the template of your component needs to get re-evaluated in order to apply the changes. In other words `ChangeDetection` has to run for your component in order to set a class to an `HTMLElement`.

With an Observable value you typically use something like the `async` pipe in order to reflect its values in your components template. ..TBD. ... more to continue, here is a link where the async behavior is already described. we can cross-link here: https://github.com/rx-angular/rx-angular/blob/main/libs/template/docs/performance-issues.md#binding-reactive-sources
```

libs/template/docs/api/class-directive.md
```diff
@@ -0,0 +1,65 @@
...
+
+### Features of `rxClass`
+
+
+- binding is always present. (see "Problems with `async` and `*ngClass` or `class`" section above)
+- it takes away the multiple usages of the `async` or `push` pipe
+- a unified/structured way of handling null and undefined
+- it allows to easily choose preferred rendering strategy using `rxClassStrategy` directive
```
hoebbelsB, line 27:
```suggestion
- re-rendering without `ChangeDetection`
- binding is always present. (see "Problems with `async` and `*ngClass` or `class`" section above)
- it takes away the multiple usages of the `async` or `push` pipe
- a unified/structured way of handling null and undefined
- it allows to easily choose preferred rendering strategy using `rxClassStrategy` directive
```

hoebbelsB (commented):
hi @svoren258 sorry for letting you wait for so long! I've reviewed the PR once again and put some comments here and there. 
I will continue on the docs part over the weekend

libs/template/class/src/lib/class.directive.ts
svoren258: It's actually `Record<string, Observable<boolean>>`.

libs/template/class/src/lib/class.directive.ts
svoren258: @hoebbelsB what is your proposal for this thing? I am not quite sure if any change is needed. 🤔

hoebbelsB: I have no proposal, I was just curious what would be the outcome of 'I will rewrite that a bit so that it will be easier to read' :)

libs/template/docs/api/class-directive.md
hoebbelsB: why just resolved?

libs/template/docs/api/class-directive.md
hoebbelsB: why resolved? I think the main benefit of `rxClass` is that it doesn't need changeDetection to apply it's changes. To get rid of the `async` pipe, you could use `push` or `rxLet` but you still need cd for applying class changes.

libs/template/docs/api/class-directive.md
hoebbelsB: same as above, the main benefit is that it doesn't need CD

libs/template/docs/api/class-directive.md
svoren258: I updated this part of the docs in one of my commits already. That's why I marked it as resolved.

svoren258:
@hoebbelsB thank you for the suggestions related to the docs improvement. Everything should be in place now, please double-check that just to be sure.

libs/template/docs/api/class-directive.md
```diff
@@ -0,0 +1,74 @@
...
+When using the native `ngClass` or `class` directive, the template of your component needs to get re-evaluated in order to apply the changes. In other words `ChangeDetection` has to run for your component in order to set a class to an `HTMLElement`.
+That's not the case when using `rxClass` that allows you to re-render the content asynchronously based on the chosen strategy.  The async behavior description can be found here: https://github.com/rx-angular/rx-angular/blob/main/libs/template/docs/performance-issues.md#binding-reactive-sources.
+
+### Features of `rxClass`
+
+- re-rendering without `ChangeDetection`
+- binding is always present. (see "Problems with `async` and `*ngClass` or `class`" section above)
+- it takes away the multiple usages of the `async` or `push` pipe
+- a unified/structured way of handling null and undefined
+- it allows to easily choose preferred rendering strategy using `rxClassStrategy` directive
```
hoebbelsB, line 34:
```suggestion
- re-rendering without `ChangeDetection`
- binding is always present. (see "Problems with `async` and `*ngClass` or `class`" section above)
- a unified/structured way of handling null and undefined
- it allows to easily choose preferred rendering strategy using `rxClassStrategy` directive
```

libs/template/docs/api/class-directive.md
```diff
@@ -0,0 +1,74 @@
...
+
+In Angular, a way of binding an observable to the `ngClass` or `class` directive could look like that:
+
+```html
+<div [ngClass]="{ 'visible': isVisible$ | async }"></div>
+<div [class.visible]="isVisible$ | async"></div>
+```
+
+When using the native `ngClass` or `class` directive, the template of your component needs to get re-evaluated in order to apply the changes. In other words `ChangeDetection` has to run for your component in order to set a class to an `HTMLElement`.
+That's not the case when using `rxClass` that allows you to re-render the content asynchronously based on the chosen strategy.  The async behavior description can be found here: https://github.com/rx-angular/rx-angular/blob/main/libs/template/docs/performance-issues.md#binding-reactive-sources.
```
hoebbelsB, line 26:
```suggestion
`rxClass` in comparison allows you to apply class updates without running `ChangeDetection` with a configurable render-strategy. 

You find more on this topic and why the usage of `async` pipe can harm your performance [here](https://github.com/rx-angular/rx-angular/blob/main/libs/template/docs/performance-issues.md#binding-reactive-sources).
```

hoebbelsB (approved):
hey @svoren258 one last set of changes to the docs. otherwise the PR LGTM! thx for the great work!

edbzn (approved):

galczo5:
@svoren258 According to the https://github.com/rx-angular/rx-angular/issues/1373#issuecomment-1175918000 it looks like there's some kind on queue and it's not going to be merged before stable release od cdk :(

hoebbelsB:
not needed anylonger

PR #1072, opened 2021-11-19
refactor(cdk): deprecate RX_ANGULAR_CONFIG 

BioPhoton:
Deprecate global config in favour of `RX_ANGULAR_RENDER_STRATEGIES_CONFIG`

### DIFF
--- a/apps/demos/src/app/app.module.ts
+++ b/apps/demos/src/app/app.module.ts
@@ -7,7 +7,7 @@ import { ENVIRONMENT_SETTINGS } from './shared/environment.token';
 import { environment } from '../environments/environment';
 import { HttpClientModule } from '@angular/common/http';
 import { HomeComponent } from './features/home/home.component';
-import { RX_ANGULAR_RENDER_STRATEGIES_CONFIG } from '../../../../libs/cdk/render-strategies/src/lib/config';
+import { RX_RENDER_STRATEGIES_CONFIG } from '../../../../libs/cdk/render-strategies/src/lib/config';
 
 @NgModule({
   imports: [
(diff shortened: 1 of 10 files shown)


apps/demos/src/app/app.module.ts
```diff
@@ -22,7 +22,7 @@ import { HomeComponent } from './features/home/home.component';
       useValue: environment,
     },
     {
-      provide: RX_ANGULAR_CONFIG,
+      provide: RX_ANGULAR_RENDER_STRATEGIES_CONFIG,
```
hoebbelsB, line 25: we could prefix such consts with just `RXA_` which would decrease the length of the name quite a bit: `RXA_RENDER_STRATEGIES_CONFIG`.

We could also think of not postfixing it with CONFIG => `RXA_RENDER_STRATEGIES`

what do u say?

libs/cdk/render-strategies/src/lib/config.ts
```diff
@@ -3,37 +3,66 @@ import { RX_CONCURRENT_STRATEGIES } from './concurrent-strategies';
 import { RxCustomStrategyCredentials, RxDefaultStrategyNames, RxStrategyNames } from './model';
 import { RX_NATIVE_STRATEGIES } from './native-strategies';
```
hoebbelsB, line 4: this one is actually prefixed with just `RX_` ...

libs/cdk/render-strategies/src/lib/strategy-provider.service.ts
```diff
@@ -126,15 +131,20 @@ export class RxStrategyProvider<T extends string = string> {
...
+   */
   constructor(
     @Optional()
     @Inject(RX_ANGULAR_CONFIG)
-    cfg: RxAngularConfig<T>
+    @Optional()
+    cfgOld: RxRenderStrategiesConfig<T>,
+    @Optional()
+    @Inject(RX_ANGULAR_RENDER_STRATEGIES_CONFIG)
+    cfg: RxRenderStrategiesConfig<T>
```
hoebbelsB, line 144: one of those `@Optional` is too much :D

hoebbelsB (changes requested):
looks good <3 , just some comments about naming

libs/cdk/render-strategies/src/lib/strategy-provider.service.ts
BioPhoton: thx

PR #980, opened 2021-10-10
chore: migrate to eslint

NileshPatel17:
@hoebbelsB This PR addresses #975 

Following Projects migrated:
1. cdk
2. state
3. template

Please let me know if this can be merged.

PR #1080, opened 2021-11-21
Migrate to eslint

edbzn:
- remove tslint
- remove tslint comments
- configure eslint (for state, template and cdk)
- setup boundaries rules


hoebbelsB:
hi @edbzn BIG THX! we need that desperately. Did u notice there is also another PR doing the same thing? There are just a couple of things left. Maybe we can connect those two together? https://github.com/rx-angular/rx-angular/pull/980

BioPhoton:
Yes lets see to get that PR in. It looks good already got feedback and is a nice contribution. ❤

edbzn:
@BioPhoton @hoebbelsB all rules have been added, I put some `@todo` in the code because some rules fail for the directive (rxFor and ViewportPrio). There's an issue with the Nx cache and build dependencies, I had to disable it for the build step. Will report the problem to Nx.

edbzn:
Close in favor of #1086

PR #1081, opened 2021-11-21
chore(cdk): enable strict mode

### LINKED ISSUE #990: enable strict mode for `@rx-angular/cdk`
Please enable strict mode for the rx-angular project based on [angulars guide](https://angular.io/guide/strict-mode) on how to configure it properly.

niklaas:
Fixes #990 .

Seems pretty straightforward. The only thing I haven't finally got my head around ist the following error:

```
libs/cdk/coalescing/src/lib/coalescingManager.ts:69:9 - error TS2345: Argument of type '([prop, value]: [KeyOf<P>, ValueOf<P>]) => void' is not assignable to parameter of type '(value: unknown, index: number, array: unknown[]) => void'.
  Types of parameters '__0' and 'value' are incompatible.
    Type 'unknown' is not assignable to type '[keyof P, ValueOf<P>]'.

69         ([prop, value]: [KeyOf<P>, ValueOf<P>]): void => {
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/libs/cdk/coalescing/src/lib/coalesceWith.ts
+++ b/libs/cdk/coalescing/src/lib/coalesceWith.ts
@@ -58,7 +58,7 @@ export function coalesceWith<T>(
       outerObserver: Subscriber<T>,
       rootSubscription: Subscription
     ): Observer<T> {
-      let actionSubscription: Unsubscribable;
+      let actionSubscription: Unsubscribable | undefined;
       let latestValue: T | undefined;
 
       const tryEmitLatestValue = () => {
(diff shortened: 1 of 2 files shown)


BioPhoton:
hi @niklaas is there any progress with the last missing thing?
If not, would you think it makes sense to merge it as is and open a second one for the last problem?

niklaas:
> hi @niklaas is there any progress with the last missing thing? If not, would you think it makes sense to merge it as is and open a second one for the last problem?

No, I haven't given this further attention after my initial push.

Let me take another look at the weekend and send you an update then. 🙌🏼

BioPhoton:
Hi @niklaas do you think we can finalize this? Is there something where we can support?

PR #1082, opened 2021-11-22
refactor: rx-angular/state sub-packages

hoebbelsB:
# Description

just like `template` and `cdk`, i've refactored rx-angular/state into sub-packages.

* `core/utils` => `@rx-angular/cdk/utils` (new)
* `rxjs` & `cdk` => `@rx-angular/cdk/state`
* `transformation-helpers` => `@rx-angular/cdk/transformations` (new)
* `RxState` => `@rx-angular/state/state` (new)
* moved spec
* moved docs
* implemented migration
* implemented tests
* removed exports from `@rx-angular/state`

list of all packages after this PR:

**state**
`@rx-angular/state`

**template**
`@rx-angular/template/let`
`@rx-angular/template/push`
`@rx-angular/template/experimental/view-prio`
`@rx-angular/template/experimental/for`

**cdk**
`@rx-angular/cdk/coalescing`
`@rx-angular/cdk/coercing`
`@rx-angular/cdk/internals/scheduler`
`@rx-angular/cdk/notifications`
`@rx-angular/cdk/render-strategies`
`@rx-angular/cdk/state`
`@rx-angular/cdk/template`
`@rx-angular/cdk/transformations`
`@rx-angular/cdk/utils`
`@rx-angular/cdk/zone-configurations`
`@rx-angular/cdk/zone-less`


BioPhoton:
> transformation-helpers => @rx-angular/cdk/transformations (new)
This is a breaking change. Could we keep them there somehow?

hoebbelsB:
> This is a breaking change. Could we keep them there somehow?

basically all of our changes lately where breaking. migrations are in place. what's the issue exactly?

hoebbelsB:
just had a quick chat with @edbzn about the current naming situation. We both agree on that `@rx-angular/state/state` is kind of redundant not feeling good.
Why not keep the main entrypoint `@rx-angular/state` exposing `RxState`. We will still be able to have sub-entry points for `effects` or whatever comes up to our mind.

So state would like this:
`@rx-angular/state` => `RxState`
`@rx-angular/state/effects` => `RxEffects`
`@rx-angular/state/x` => `RxX`  :trollface:  

@BioPhoton what do you think?

hoebbelsB:
so, after discussing that on the team-meeting we come to the following conclusion.
We don't introduce a breaking change now, we keep everything exported from `@rx-angular/state` as is.
To achieve it, we will copy/paste all the migrated code back to the package and mark it as `deprecated`. No change will happen to the deprecated code any longer, we plan to remove it asap.

On the otherhand we ship all the new code in `@rx-angular/cdk/*` sub-entrypoints.

edbzn (approved):

PR #1095, opened 2021-11-24
docs(state): add Marmicode Tasting Session resource

edbzn:


BioPhoton:
The best resource ever!! 

We should make it more present!

edbzn:
@BioPhoton What about displaying the video in the web doc instead of a simple link?

PR #1097, opened 2021-11-24
docs(template): improve API doc

edbzn:
**WIP**

- Add rxFor doc
- Improve main readme
- Move some documentation files


apps/demos/src/app/features/template/rx-let/let-template-binding/let-template-binding.component.spec.ts
```diff
@@ -0,0 +1,111 @@
...
+
+import { LetTemplateBindingComponent } from './let-template-binding.component';
+import { LetTemplateBindingModule } from './let-template-binding.module';
+
+const RX_ANGULAR_TEST_PROVIDER: StaticProvider = {
+  provide: RX_RENDER_STRATEGIES_CONFIG,
+  useValue: {
+    primaryStrategy: 'native',
+  },
+};
```
edbzn, line 17: @rx-angular/core-team do we want to provide this out of the box in the `cdk`?

hoebbelsB: yes please, we should provide a testing package

libs/template/docs/api/let-directive.md
```diff
@@ -103,6 +103,47 @@ Internally, `*rxLet` is using a simple "view memoization" - it caches all anchor
...
+  ...
+  providers: [RX_ANGULAR_TEST_PROVIDER],
+}).compileComponents();
+```
+
+This way, `*rxLet` will use the same rendering strategy used by the Angular's built-in `async` pipe.
+
+### Handling the lazy rendering problem
+
+Keep in mind that no matter which strategy is used, `*rxLet` always renders the view asynchronously. For this reason, you want to be sure the asynchronous change detection is done before querying any DOM element rendered by the directive. You can achieve this using the `ComponentFixture#whenStable()` function.
```
hoebbelsB, line 134: this should not be true. The provided strategy itself is accountable for the let directive rendering sync or async. e.g. `native` has no built in scheduling or other asynchronous calls, thus it will instantly call `cdRef#markForCheck`.

maybe for testing we should even provide a very new strategy which is sync and calls `detectChanges` instead. This would probably result in the desired behavior

hoebbelsB (changes requested):
hey @edbzn awesome PR, thx for that!
I have left some comments in the testing section which we should discuss imo. Also, i'm thinking about moving the testing content to the main page.
We can easily generalize everything you've stated so far since it will be true for any strategy-based component in rx-angular.

apps/demos/src/app/features/template/rx-let/let-template-binding/let-template-binding.component.spec.ts
hoebbelsB: @edbzn we probably will have a testing package soon: https://github.com/rx-angular/rx-angular/pull/1305 :)

hoebbelsB:
close in favor of https://github.com/rx-angular/rx-angular/pull/1420

PR #1107, opened 2021-12-02
Resolve null error in cdk-template-notifier.ts

### LINKED ISSUE #1106: TypeError: Cannot read properties of null (reading 'then')
# Description

```
Uncaught TypeError: Cannot read properties of null (reading 'then')
    at MapSubscriber.project (cdk-notifications.js:196)
    at MapSubscriber._next (map.js:29)
    at MapSubscriber.next (Subscriber.js:49)
    at FilterSubscriber._next (filter.js:33)
    at FilterSubscriber.next (Subscriber.js:49)
    at MapSubscriber._next (map.js:35)
    at MapSubscriber.next (Subscriber.js:49)
    at DistinctUntilChangedSubscriber._next (distinctUntilChanged.js:50)
    at DistinctUntilChangedSubscriber.next (Subscriber.js:49)
    at ReplaySubject.next (Subject.js:39)
    at ReplaySubject.nextInfiniteTimeWindow (ReplaySubject.js:31)
    at Object.next (cdk-notifications.js:225)
    at LetDirective.set rxLet [as rxLet] (template-let.js:191)
    at Object.ngOnChangesSetInput [as setInput] (core.mjs:1517)
    at setInputsForProperty (core.mjs:10976)
```

Should not receive a null error if `observable$` is undefined.

Uncertain at this time, but a `null` error is clearly possible with this code design:

[create-template-notifier.ts:151](https://github.com/rx-angular/rx-angular/blob/3f9986e036934159e256c259c45593ade7385fbf/libs/cdk/notifications/src/lib/create-template-notifier.ts#L151)
```
function createTemplateNotifier(withSuspenseTpl) {
   ...
    map((observable$) => {
        const isNull = observable$ === null;
        const isPromiseOrObs = typeof observable$.then === 'function' || isObservable(observable$); // <-- NULL ERROR HERE IF observable$ is null
```

Ironically, there is a null check immediately before the affected line, but it is not used to protect the following conditional evaluation.

# Environment

rx-angular/cdk@1.0.0-alpha.11
rx-angular/template@1.0.0-beta.29

```
Ionic:

   Ionic CLI                     : 6.18.1 (/usr/local/lib/node_modules/@ionic/cli)
   Ionic Framework               : @ionic/angular 5.9.1
   @angular-devkit/build-angular : 13.1.0-next.2
   @angular-devkit/schematics    : 13.1.0-next.2
   @angular/cli                  : 13.1.0-next.2
   @ionic/angular-toolkit        : 5.0.3

Capacitor:

   Capacitor CLI      : 3.3.1
   @capacitor/android : 3.3.1
   @capacitor/core    : 3.3.1
   @capacitor/ios     : 3.3.1

Utility:

   cordova-res : 0.15.3
   native-run  : 1.5.0

System:

   NodeJS : v14.15.0 (/usr/local/bin/node)
   npm    : 6.14.9
   OS     : macOS Big Sur


```

# Related to Other Issues

# Tasks to Resolve This

Implement null-safe handling in `cdk-notifications`:

# Notes

lincolnthree:
This fixes #1106.


hoebbelsB (commented):
hey @lincolnthree i'm sorry that it has taken so long to finally review this. very good catch, thank you so much!
I don't want to bother you, but I have one question: can we add a test case which would fail with the previous version? Just to be 100% sure :)

edbzn:
@lincolnthree I'm trying to reproduce the error to add a test case against, but I'm not able to reproduce it. How do you get the error, do you have an example to give?

lincolnthree:
Hey @edbzn Thanks so much for doing that. To be honest, I don't remember :( I believe it happened when null was passed in where an observable is expected. But that's obvious from the pull request. I just don't remember how it was wired together to reproduce the issue, I'm guessing it could be occurring when using something like push pipe with a null observable:

```
 transform<U>(
    potentialObservable: ObservableInput<U> | null | undefined, <--- NULL OBSERVABLE RECEIVED HERE
    config:
      | PushInput<U, S>
      | RxStrategyNames<S>
      | Observable<RxStrategyNames<S>>
      | undefined,
    renderCallback?: NextObserver<U>
  ): U | null | undefined {
    this._renderCallback = renderCallback;
    if (config) {
      if (isRxComponentInput(config)) {
        this.strategyHandler.next(config.strategy as string);
        this._renderCallback = config.renderCallback;
        // set fallback if patchZone is not set
        this.setPatchZone(config.patchZone);
      } else {
        this.strategyHandler.next(config as string);
      }
    }
    this.templateObserver.next(potentialObservable); <--- NULL OBSERVABLE PASSED HERE
    return this.renderedValue as U;
  }
```

edbzn:
@lincolnthree I'm not able to reproduce the issue, tried multiple things but no way. I always get an observable in `create-template-notifier.ts` and not a `null` value causing the error.

lincolnthree:
Yeah I clearly see this code in the npm package:

```
191         const isNull = observable$ === null;^M
192         const isPromiseOrObs = typeof observable$.then === 'function' || isObservable(observable$);^
```

It shouldn't be possible to be null there, but somehow it happened.

IMO. It would be best to make this change just for future refactoring safety. But I do not see a way to test it other than to test that it doesn't happen (which it currently does not appear to.)

lincolnthree:
Wait wait wait... what does `mapFirst` do?

lincolnthree:
YES! That's it. Only the first value is null-protected by the mapFirst operator. Any subsequent null values will cause this to fail.

edbzn:
Thanks for the help @lincolnthree!

lincolnthree:
You're welcome. Glad we got it figured out :)

PR #1124, opened 2021-12-15
New ESLint plugin

matejchalk:
This PR adds an `@rx-angular/eslint-plugin` package with 10 custom ESLint rules (for reactive, performant and Zone-less Angular apps).

The rules in question were previously implemented by @Karnaukhov-kh in TSLint (this original implementation isn't open-source). I then reimplemented them in ESLint (since TSLint is deprecated), added unit tests, extended @Karnaukhov-kh's docs with examples and generally configured the ESLint plugin for public use (tested integration with `npm link`).

### DIFF
--- a/libs/cdk/render-strategies/src/lib/onStrategy.ts
+++ b/libs/cdk/render-strategies/src/lib/onStrategy.ts
@@ -1,7 +1,7 @@
 import { NgZone } from '@angular/core';
 import { RxCoalescingOptions } from '@rx-angular/cdk/coalescing';
-import { switchMap, take } from 'rxjs/operators';
-import { Observable, of, throwError } from 'rxjs';
+import { Observable, throwError } from 'rxjs';
+import { catchError, map, take } from 'rxjs/operators';
 import { RxRenderWork, RxStrategyCredentials } from './model';
 
 /**
@@ -22,22 +22,16 @@ export function onStrategy<T>(
   ) => void,
   options: RxCoalescingOptions & { ngZone?: NgZone } = {}
 ): Observable<T> {
-  let error: Error;
   return new Observable<T>((subscriber) => {
     subscriber.next(value);
   }).pipe(
     strategy.behavior({
-      work: () => {
-        try {
-          workFactory(value, strategy.work, options);
-        } catch (e) {
-          error = e;
-        }
-      },
+      work: () => workFactory(value, strategy.work, options),
       scope: (options.scope as Record<string, unknown>) || {},
       ngZone: options.ngZone,
     }),
-    switchMap(() => (error ? throwError([error, value]) : of(value))),
+    catchError((error) => throwError(() => [error, value])),
+    map(() => value),
     take(1)
   );
 }


libs/eslint-plugin/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+```bash
+npm install --save-dev @rx-angular/eslint-plugin
+```
+
+To use the recommended ruleset, extend the configuration in your ESLint config (e.g. `.eslintrc.json`):
+
+```json
+{
+  "extends": ["@rx-angular/recommended"]
```
hoebbelsB, line 23: i love that one :)

libs/eslint-plugin/src/lib/rules/prefer-no-lodash-clone-deep.ts
```diff
@@ -0,0 +1,32 @@
...
+  name: path.parse(__filename).name,
+  meta: {
+    docs: {
+      recommended: 'error',
+      description: 'Detects all usages of cloneDeep from Lodash.',
+    },
+    type: 'problem',
+    messages: {
+      [MESSAGE_ID]:
+        'Use with caution - cloneDeep is a costly operation that may cause performance problems.',
```
hoebbelsB, line 18: we could give a hint here and point to `structuredClone` https://web.dev/structured-clone/

hoebbelsB (approved):
@matejchalk amazing PR, big big thx for the contribution! I hope we can soon move forward with our release-schedule!

libs/eslint-plugin/src/lib/rules/prefer-no-lodash-clone-deep.ts
matejchalk: Good suggestion :+1: I've added it now.

edbzn:
Hey, I love this PR! 

Here are some struggles I faced in the usage of rxa that I think we can avoid by adding proper eslint rules. 

1- The first thing I face over and over is the inconsistent mixing of reactive and imperative usage. This is a real struggle for me and I think we can prevent this by adding a rule. Here is an example, of that inconsistent mixing:

```ts
     this.state.connect(
            'something',
            this.service.something$.pipe(
                map((something) =>
                   ({ ...something, ...this.state.get('somethingElse') }) // <- please no
                ),
            ),
      );
```

The rule would be no imperative usage within `state#hold` and `state#connect` methods.

2- Another useful rule that I can see is the recommended place to put both `state#hold` and `state#connect`, I have seen some usage of these methods in another place than the constructor (eg in a method called in the template). Which IMO is bad for several reasons (not reactive).

```ts
export class MyComponent {
    doSomething() {
        this.state.connect('something', this.service.doSomething()) // <- please no
    }
}
```

3- For this one, I'm not sure if it's possible, but I often see mixed usage of state properties and component properties which leads to inconsistent component state. Eg: 

```ts
export class MyComponent {
    isLoading = false; // <- please no
    
    constructor(private state: RxState<{ loaded: boolean }>) {}
}
```

What do you think about those?

matejchalk:
Closed this by accident :facepalm: I added 2 of @edbzn's suggested rules, you can check them out in the replacement PR #1394.

PR #1169, opened 2022-01-25
docs(template): fix for "Render Strategies" link

galczo5:


galczo5:
@BioPhoton I think I need your acceptance to continue with the rest of checks.

PR #1176, opened 2022-01-31
added a link to the readme

web-dave:
I've added a link to my livestrem recording.
Michael was so kind and explained rx-state to me and my stream.
Most importaten: It's the only german contente about rx-state so far.


BioPhoton:
Hi! Thanks a lot for the content!

Could you drag and drop an image into the readme here in Github? This way we can avoid the load of inline encoded images and keep the file maintainable.

It should be possible to just use the suggested change feature and that's it.

Let me know if I can help!

BioPhoton:
still see the base64 strig, but no problem, I can fix it late.
 
 THANKS a lot for the contribution and the content you created! 
 😁😁😁😁😁😁😍😍😍😘😘😘

PR #1183, opened 2022-02-02
fix(cdk): drop `ObservableNotification` to support rxjs 6

edbzn:


hoebbelsB (approved):
can confirm that cdk and template are now working with rxjs 6!

PR #1185, opened 2022-02-02
chore: upgrade dependencies

hoebbelsB:
# Description

upgraded dependencies:
* nx
* ng
* jest
* typescript
* cypress
* ....

@edbzn @BioPhoton should we do
```ts
"angularCompilerOptions": {
    "enableIvy": true
  },
```

???


edbzn (approved):
I think we should introduce the partial compilation mode:

```json
{
  "angularCompilerOptions": {
    "enableIvy": true,
    "compilationMode": "partial"
  }
}
```

That's the recommended way since Angular v11.1

hoebbelsB:
didn't it change with v13? not sure :D

edbzn:
@hoebbelsB still the case in v13, we should stick to the partial compilation mode.

PR #1191, opened 2022-02-03
fix: drop `@nrwl/tao` deep import

edbzn:


hoebbelsB:
@edbzn since beta.0 is already published, should we target the new migration towards beta.1?

edbzn:
@hoebbelsB yes if we want to trigger the migration for people that already ran the update.

hoebbelsB:
@edbzn can you refactor it so it runs with beta.1? we can merge after thats done

PR #1197, opened 2022-02-06
fix(template): don't override primary strategy 

hoebbelsB:
# Description

somehow the let directive is resetting the primary strategy of the global `StrategyProvider` on init.
I guess this is a leftover of times where each letdirective provided its very own `StrategyProvider`.

This PR removes the piece of code, because now each let directive has impact on a global scope which is not what we actually want


BioPhoton:
Nice catch!

edbzn (approved):

PR #1198, opened 2022-02-06
fix(template): Rxfor template typings

### LINKED ISSUE #1195: Wrong type for each looped item in `rxFor`  and mismatching type for `rxFor` directive
# Description

`rxFor` returns incorrect type for each looped item. See attached screenshot

`rxFor` type error:
rxfor-type-error

source observable:
source

looped item:
loop-item

Sample build error in stackblitz repro:
from stackblitz

Sample repro in StackBlitz: https://stackblitz.com/edit/angular-egwkfk
Repo created by StackBlitz: https://github.com/exequiel09/rxfor-test

# Environment

Angular CLI: 13.1.1
Node: 16.13.2
OS: macOS 12.2

TypeScript config:
enableIvy: true
strict: true

hoebbelsB:
# Description

closes #1195 

I've taken another look at the implementation of the template typings in rxFor. After comparing them against the original `ngFor`, i've noticed some differences.

`rxFor` now should be able to derive the correct typings in the template even without defining a `trackBy` function.

## ViewContext

First of all, the class `RxDefaultListViewContext` is missing a property which reflects the type of the actual `NgIterable`.
That's why I've introduced a new class `RxForViewContext` extending from the default one.

```ts
export class RxForViewContext<
  T,
  U extends NgIterable<T> = NgIterable<T>,
  K = keyof T
> extends RxDefaultListViewContext<T, U, K> {
  constructor(
    item: T,
    public rxForOf: U,
    customProps?: { count: number; index: number }
  ) {
    super(item, customProps);
  }
}

/** @internal */
  static ngTemplateContextGuard<
    T,
    U extends NgIterable<T> = NgIterable<T>,
    K = keyof T
  >(dir: RxFor<T, U>, ctx: any): ctx is RxForViewContext<T, U, K> {
    return true;
  }

```

## Input Typing

The input typing also was missing a hint to the `NgIterable`

```ts
/** @internal */
  @Input()
  set rxForOf(
    potentialObservable:
      | Observable<(U & NgIterable<T>) | undefined | null>
      | (U & NgIterable<T>)
      | null
      | undefined
  ) {
    this.observables$.next(potentialObservable);
  }
```

## Selector

I think defining the selector as `[rxFor][rxForOf]` finally solved the issue @exequiel09 was facing `Type '""' is not assignable to...`.
It also made it possible to get rid of the duplicated `@Input` property for `rxFor`.

## Template Reference Variables

based on the base implementation, i've introduced a new `@Input()` again reflecting the templateRef created for each attached element.

```ts

/**
   * @internal
   * A reference to the template that is created for each item in the iterable.
   * @see [template reference variable](guide/template-reference-variables)
   * (inspired by @angular/common `ng_for_of.ts`)
   */
  @Input()
  set rxForTemplate(value: TemplateRef<RxForViewContext<T, U>>) {
    if (value) {
      this.templateRef = value;
    }
  }

```

### DIFF
--- a/apps/demos/src/app/features/template/rx-context/rx-context.module.ts
+++ b/apps/demos/src/app/features/template/rx-context/rx-context.module.ts
@@ -1,4 +1,6 @@
 import { NgModule } from '@angular/core';
+import { ForModule } from '@rx-angular/template/experimental/for';
+import { LetModule } from '@rx-angular/template/let';
 import { VisualizerModule } from '../../../shared/debug-helper/visualizer';
 import { StrategySelectModule } from '../../../shared/debug-helper/strategy-select';
 import { ValueProvidersModule } from '../../../shared/debug-helper/value-provider';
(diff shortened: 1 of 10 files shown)


exequiel09:
Wow that was fast! 🚀

exequiel09:
Thanks @hoebbelsB @BioPhoton . I was able to confirm that type errors are gone. Can build production builds and it even works on SSR 🎊

Issue #1261, opened 2022-03-08
@rx-angular/template depends on @angular/platform-browser v12


jochenjonc:
# Description

When doing an `npm install` on an Angular 13 project with the @rx-angular/template package, it cannot resolve the @angular/platform-browser@"12.2.16" dependency.

```
npm ERR! Could not resolve dependency:
npm ERR! peer @angular/platform-browser@"12.2.16" from @rx-angular/template@1.0.0-beta.32
npm ERR! node_modules/@rx-angular/template
npm ERR!   @rx-angular/template@"1.0.0-beta.32" from the root project
```

# Steps to Reproduce the Issue

New Angular 13 project and then install @rx-angular/template

zandorstorm:
npm install --save @rx-angular/template @rx-angular/cdk --force

it worked for me

PR #1184, opened 2022-02-02
docs: add angular version compatibility table

hoebbelsB:
# Description

added a table which displays which rxa versions support which angular version

PR #1289, opened 2022-03-25
fix: add missing peerDependencies

edbzn:
Resolves https://github.com/rx-angular/rx-angular/issues/1261. Under the hood, ng-packagr adds multiple peer dependencies to the distribution package.json with an incorrect version range. This fix ensures versions are well defined to deal with npm 7.

### DIFF
--- a/libs/cdk/package.json
+++ b/libs/cdk/package.json
@@ -48,7 +48,8 @@
     "url": "https://github.com/rx-angular/rx-angular.git"
   },
   "peerDependencies": {
-    "@angular/core": ">10.1.1"
+    "@angular/core": ">=12.0.0",
+    "rxjs": ">=6.5.5"
   },
   "dependencies": {
     "tslib": "^2.0.0",
--- a/libs/state/package.json
+++ b/libs/state/package.json
@@ -48,7 +48,9 @@
     "url": "https://github.com/rx-angular/rx-angular.git"
   },
   "peerDependencies": {
-    "@angular/core": ">8.0.0"
+    "@angular/core": ">=12.0.0",
+    "@angular/platform-browser": ">=12.0.0",
+    "rxjs": ">=6.5.5"
   },
   "dependencies": {
     "tslib": "^2.0.0",
--- a/libs/template/package.json
+++ b/libs/template/package.json
@@ -49,8 +49,10 @@
     "url": "https://github.com/rx-angular/rx-angular.git"
   },
   "peerDependencies": {
-    "@angular/core": ">10.1.1",
-    "@rx-angular/cdk": ">=1.0.0-beta.2"
+    "@angular/core": ">=12.0.0",
+    "@angular/platform-browser": ">=12.0.0",
+    "@rx-angular/cdk": ">=1.0.0-beta.2",
+    "rxjs": ">=6.5.5"
   },
   "dependencies": {
     "tslib": "^2.0.0",


BioPhoton:
It is a good idea to provide different versions for the peer depts?

- "rxjs": ">=6.0.0" & "rxjs": ">=6.5.0"
- "@angular/core": ">8.0.0" &  "@angular/core": "10.1.1"
- "@angular/cdk": ">8.0.0" & "@angular/cdk": ">10.1.1"
- "@angular/platform-browser": ">8.0.0" & "@angular/platform-browser": ">10.1.1"

edbzn:
@BioPhoton I aligned peer dependencies @angular/* versions to `>=8.0.0` and RxJS to `>=6.4.0`.

mackelito:
When will this fix be released?

libs/cdk/package.json
```diff
@@ -48,7 +48,8 @@
     "url": "https://github.com/rx-angular/rx-angular.git"
   },
   "peerDependencies": {
-    "@angular/core": ">10.1.1"
+    "@angular/core": ">=8.0.0",
```
hoebbelsB, line 51: do we really support angular 8?

libs/state/package.json
```diff
@@ -48,7 +48,10 @@
     "url": "https://github.com/rx-angular/rx-angular.git"
   },
   "peerDependencies": {
-    "@angular/core": ">8.0.0"
+    "@angular/core": ">=8.0.0",
+    "@angular/cdk": ">=8.0.0",
```
hoebbelsB, line 52: angular/cdk? I don't think thats true

libs/template/package.json
```diff
@@ -49,8 +49,11 @@
     "url": "https://github.com/rx-angular/rx-angular.git"
   },
   "peerDependencies": {
-    "@angular/core": ">10.1.1",
-    "@rx-angular/cdk": ">=1.0.0-beta.2"
+    "@angular/core": ">=8.0.0",
+    "@rx-angular/cdk": ">=1.0.0-beta.2",
+    "@angular/cdk": ">=8.0.0",
```
hoebbelsB, line 54: this is true, but I would rather remove the dependency to angular/cdk. We only use it once in the `rxIf`. Can u remove that dependency and adjust the package.json?

hoebbelsB (changes requested):
hi @edbzn thx for the PR! Left some comments with feedback :)

libs/template/package.json
edbzn: We can't remove `@angular/cdk` because ng-packagr will put it in the distribution package.json at build time with a wrong version range, I think the best here is to use the optional peer dependency flag.

libs/cdk/package.json
edbzn: Not tested, but I guess so. It's not very clear which version we support.

libs/template/package.json
hoebbelsB: sorry, my comment led to a misunderstanding. the plan is to remove the usage of `angular/cdk` in rxIf, so that we can remove the dependency from the package.json as well :) it's just the one function which we can replace easily

kroeder:
Is there anything I can do to complete this PR? Without it, this package is the reason why people need to do `npm install --force` or else they run into a dep graph error in npm 8

> LGTM, I would love to have a testrun with an angular application with v8 though ^^

@hoebbelsB  Do you want a smoke test in this repo using Angular 8 or just a public test and a "works" statement? 🙂

edbzn:
Agree with @kroeder, we should stick to the Angular versioning support policy. I updated the peer dependencies to `>=11.0.0`.

hoebbelsB (changes requested):
I have just double-checked and created a fresh angular 11 project. Unfortunately, the rx-angular packages are not compatible.

just an example error:

```bash
Error: ./node_modules/@rx-angular/template/fesm2015/template-unpatch.js 133:10-28
"export 'ɵɵFactoryTarget' (imported as 'i0') was not found in '@angular/core'
```

It is working with > 12. Also, looking at this [gist](https://gist.github.com/LayZeeDK/c822cc812f75bb07b7c55d07ba2719b3), we should go with rxjs >= 6.5.5, since this is what angular will force the user anyway to use.

hoebbelsB:
I have aligned the compatibility table to what i've posted in the PR as well: https://github.com/rx-angular/rx-angular/pull/1184

we should merge both of those PR's asap :)

hoebbelsB:
this PR will probably resolve https://github.com/rx-angular/rx-angular/issues/1295 as well, so I've linked it

hoebbelsB (approved):
LGTM

kroeder:
Is it possible in the near future to create a release including this PR? The last one was in March and npm 8 still throws errors for any command you execute (`npm dedup`, `npm i`, ...) 🙏

PR #1310, opened 2022-04-11
fix(template): drop @angular/cdk dep for rxIf

edbzn:
Related to #1289, drop the `@angular/cdk` dependency. Let me know if you think the function `coerceBooleanProperty` should be exposed in the `@rx-angular/cdk/coercion` instead of being inlined in the `RxIf` directive. As it's only used here I prefer to not augment the API surface.


hoebbelsB (approved):
love it, thanks!

PR #1313, opened 2022-04-25
fix(cdk): requestPaint should force a paint

Gabweb:
# Description

requestPaint is exposed publicly by CDK as part of the internal scheduler. However, it is without any effect - it seems to be broken since 20c4197784fa83092a97e6c76060ab07e80e18fd. This PR fixes the functionality of requestPaint. Calling requestPaint will force the scheduler to 'yield to host'. 

### Example use-case for requestPaint:
Having two chunks of the UI one small one (~5ms, e.g. a side-bar) and a large one (~100ms, e.g. the main-content). Usually both chunks would be scheduled into the some frame and blocking the main thread. As an improvement one could call requestPaint in the smaller chunk and forcing the browser to render this part earlier.

### DIFF
--- a/libs/cdk/internals/scheduler/src/lib/scheduler.spec.ts
+++ b/libs/cdk/internals/scheduler/src/lib/scheduler.spec.ts
@@ -17,6 +17,7 @@ describe('Scheduler', () => {
 
   const NormalPriority = 3;
   let scheduleCallback: typeof import('./scheduler').scheduleCallback;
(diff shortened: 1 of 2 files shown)


hoebbelsB (approved):
hi @Gabweb nice to see you here :partying_face:!

Also thank you very much for the fix, test and explanation. Tbh, this API wasn't exposed on purpose :D. It probably should be hidden behind a `RenderStrategy`

PR #1344, opened 2022-05-18
feat(cdk): add rxScheduleTask function

Karnaukhov-kh:
Introduces `rxScheduleTask` function to schedule tasks with our scheduler.


libs/cdk/render-strategies/docs/rx-schedule-task.md
```diff
@@ -0,0 +1,111 @@
...
+  this.stateService.set(data);
+
+  rxScheduleTask(
+    () => localStorage.setItem('state', JSON.stringify(state)),
+    'idle'
+  );
+}
+```
+
+### Usage with options strategy
```
Karnaukhov-kh, line 70:
```suggestion
### Usage with options
```

libs/cdk/render-strategies/docs/rx-schedule-task.md
```diff
@@ -0,0 +1,111 @@
...
+...
+
+let saveToLocalStorageCallback;
+
+function updateStateAndBackup<T>(data: T) {
+  this.stateService.set(data);
+
+  if (saveToLocalStorageCallback) {
+    saveToLocalStorage();
+  }
```
hoebbelsB, line 100: shouldn't this call the `saveToLocalStorageCallback` ?

libs/cdk/render-strategies/src/lib/rx-schedule-task.ts
```diff
@@ -0,0 +1,47 @@
...
+) => {
+  const task = scheduleCallback(
+    strategiesPrio[strategy],
+    () => work(),
+    options
+  );
+  return () => {
+    cancelCallback(task);
+  };
+};
```
hoebbelsB, line 47: why don't we provide an observable version of this standalone function as well?

libs/cdk/render-strategies/src/lib/rx-schedule-task.ts
```diff
@@ -0,0 +1,47 @@
...
+ * This function is used to schedule a task with a certain priority.
+ * It is useful for tasks that can be done asynchronously.
+ *
+ * ```ts
+ * const task = rxScheduleTask(() => localStorage.setItem(state, JSON.stringify(state));
+ * ```
+ */
+export const rxScheduleTask = (
+  work: (...args: any[]) => void,
+  strategy: keyof StrategiesPriorityRecord = 'normal',
```
hoebbelsB, line 33: How would a user use the function with a custom strategy? the typing wouldn't allow it currently. Or am I missing something?

libs/cdk/render-strategies/docs/rx-schedule-task.md
```diff
@@ -0,0 +1,111 @@
...
+...
+
+function updateStateAndBackup<T>(data: T) {
+  this.stateService.set(data);
+
+  rxScheduleTask(
+    () => localStorage.setItem('state', JSON.stringify(state)),
+    'idle',
+    { delay: 200, zone: this.ngZone }
+  );
```
edbzn, line 83: One thing I try to avoid is having two optional params because if you want for instance to customize the delay you have to pass the strategy. What about moving the strategy param in the configuration object as well? 

```suggestion
  rxScheduleTask(
    () => localStorage.setItem('state', JSON.stringify(state)),
    { delay: 200, zone: this.ngZone, strategy: 'idle' }
  );
```

So you can only pass options you want to customize and the order does not matter.

hoebbelsB (commented):
hi @Karnaukhov-kh very nice PR, i like the idea a lot! Left some thoughts in the comments

edbzn (commented):
Nice work, I would add some unit tests to see if the `scheduleCallback` is correctly called. I think that `render-strategies` is not the right place to expose that fn, but I can't figure out a better place.

Karnaukhov-kh:
Thank you for feedback @hoebbelsB @edbzn, I'll answer and apply changes soon :).

libs/cdk/render-strategies/docs/rx-schedule-task.md
Karnaukhov-kh: Yes, it should, thx

libs/cdk/render-strategies/docs/rx-schedule-task.md
Karnaukhov-kh: Totally agree with your suggestion, fixed, thanks :)

libs/cdk/render-strategies/src/lib/rx-schedule-task.ts
Karnaukhov-kh: I guess we should not use custom strategy here, because only thing we need from strategy is priority. And scheduler not working with custom priority levels.

That's why we need this, to provide users familiar API instead of `PriorityLevel` enum.

```
const strategiesPrio: StrategiesPriorityRecord = {
  immediate: PriorityLevel.ImmediatePriority,
  userBlocking: PriorityLevel.UserBlockingPriority,
  normal: PriorityLevel.NormalPriority,
  low: PriorityLevel.LowPriority,
  idle: PriorityLevel.IdlePriority,
};
```

libs/cdk/render-strategies/src/lib/rx-schedule-task.ts
Karnaukhov-kh: Isn't `scheduleOnQueue` already an operator that uses it, but with coalescing?

libs/cdk/render-strategies/src/lib/rx-schedule-task.ts
```diff
@@ -0,0 +1,52 @@
...
+    () => work(),
+    {
+      delay: options?.delay,
+      ngZone: options?.ngZone,
+    }
+  );
+  return () => {
+    cancelCallback(task);
+  };
+};
```
edbzn, line 52: What about the following syntax to initialize options? 

```suggestion
export const rxScheduleTask = (
  work: (...args: any[]) => void,
  {
    strategy = defaultStrategy,
    delay,
    ngZone,
  }: {
    strategy?: keyof StrategiesPriorityRecord;
    delay?: number;
    ngZone?: NgZone;
  } = {}
) => {
  const task = scheduleCallback(strategiesPrio[strategy], () => work(), {
    delay,
    ngZone,
  });
  return () => {
    cancelCallback(task);
  };
};
```

Karnaukhov-kh:
Closing in favor of #1633

PR #1368, opened 2022-06-09
perf(template): PushPipe - don't over-render sync initial values

hoebbelsB:
# Make the PushPipe great again :rocket: 

for a looooooooooooooooooooooooooooooooooong time now I wanted to tackle this nasty bug which basically makes the push-pipe quite useless since it overrenders each and every component attaching sync values to it.

This PR tweaks the changedetection behavior of the `PushPipe`. It will now detect if the passed observable has an initial value or not. In case there is a sync value, the PushPipe will skip the first change-detection run, since the renderedValue will be evaluated anyway during this part of the components lifecycle.
Every continuously emitted value will cause the regular rendering flow, though.

# Testing

I've implemented unit tests for different sets of asynchronous and sync values to prove the pushpipe is rendering without the need of CD when a sync value was passed + it runs CD & renders when the value is async


edbzn (approved):

Karnaukhov-kh (approved):

edbzn:
@hoebbelsB is it a breaking change due to the fact that some apps can rely on those extra CD cycles? Otherwise, we can merge I believe.

PR #1374, opened 2022-07-18
feat(template): implement rxLet template triggers

hoebbelsB:
# Description

template-triggers

Contextual Template States

[Contextual Template States](https://docs.google.com/presentation/d/1z3A0GjDJpu3_Lipmcqn9BL1LLTKXaMgYBKt1UKUs5rY/edit?usp=sharing)

implemented template triggers for rxLet directive

* [x] demo
* [x] docs (JSdocs)
* [x] unit tests 

notable changes:

**template-notifier**

`create-template-notifier` no longer has responsibility for the suspense template. Instead the `Suspense` notification handling was improved. `create-template-notifier` will now emit suspense on the following occasions:

* new observable input with a lazy or no value (if value is sync, no suspense notification is send)
  * EMPTY | NEVER included
* static undefined

**template-manager**

`template-manager` now takes over the responsibility of managing the suspense state, if it should be processed or not. If no `SuspenseTemplate` is provided, the first suspense notification will be ignored as this would violate with rxLets lazy template nature. All consequent suspense notifications though will cause the `$suspense` context variable to toggle.


BioPhoton:
A really nice change of responsibility!!!
Motivated to write technical docs for template Manager 😅

libs/cdk/template/src/lib/template-manager.ts
```diff
@@ -138,79 +142,116 @@ export function createTemplateManager<
   const getContext = notificationKindToViewContext(
     templateSettings.customContext || ((v) => {})
   );
+  let withSuspense = false;
 
   return {
-    addTemplateRef: templates.add,
+    addTemplateRef: (name: N, templateRef: TemplateRef<C>) => {
+      // @ts-ignore
```
edbzn, line 149: Do you remember why we need this here?

libs/template/let/src/lib/let.directive.ts
```diff
@@ -13,34 +13,38 @@ import {
...
   Observable,
   ObservableInput,
   ReplaySubject,
   Subject,
   Subscription,
+  filter,
 } from 'rxjs';
-import { mergeAll } from 'rxjs/operators';
+
+/* eslint-disable @angular-eslint/no-input-rename */
```
edbzn, line 47: We should consider turning off this rule.

edbzn (approved):
To be honest, I'm not sure to fully understand what happens in the template manager, but there are tests and I can understand them so it looks fine!

libs/cdk/template/src/lib/template-manager.ts
hoebbelsB: no, I didn't understand and gave up after 5 minutes :D

libs/cdk/template/src/lib/template-manager.ts
```diff
@@ -138,79 +142,116 @@ export function createTemplateManager<
...
+          const templateName = notificationToTemplateName[contextKind](
             value,
             templates
           );
           const template = templates.get(templateName);
-          return { template, templateName, notification };
+          return { template, templateName, notification, contextKind };
         }),
-        filter(({ template }) => !!template),
+        // filter(({ template }) => !!template),
```
Karnaukhov-kh, line 195: Are we sure it's not needed? If yes, remove it please

Karnaukhov-kh (approved):

edbzn:
I think I screwed up the build, let me fix that

PR #1380, opened 2022-07-25
feat(template): introduce rxChunk

### LINKED ISSUE #1802: Feature rxChunk

# Solution

hoebbelsB:
# ChunkDirective

This PR introduces the `*rxChunk` structural directive.

the  `*rxChunk` directive serves as a convenient way for dividing template work into chunks.
Applied to an element, it will schedule a task with the given RxRenderStrategy in order to postpone the template creation of this element.

rx-chunk

closes https://github.com/rx-angular/rx-angular/issues/1802

### PR Checklist

- [x] implementation
- [x] demo
- [x] unit tests
- [x] docs (jsdocs)

### Features of `*rxChunk`

* lightweight alternative to `*rxLet="[]"`
* renderCallback
* no value binding
* no context variables
* zone agnostic
* suspense template

long task

distribute work

### DIFF
--- a/libs/template/chunk/src/lib/chunk.directive.ts
+++ b/libs/template/chunk/src/lib/chunk.directive.ts
@@ -1,0 +1,178 @@
+import {
+  Directive,
+  inject,
+  Input,
+  NgZone,
+  OnDestroy,
+  OnInit,
+  TemplateRef,
+  ViewContainerRef,
+} from '@angular/core';
+import { RxStrategyProvider } from '@rx-angular/cdk/render-strategies';
+import { NextObserver, Subscription } from 'rxjs';
+
+/**
+ * @Directive ChunkDirective
+ *
+ * @description
+ *
+ * The `*rxChunk` directive serves as a convenient way for dividing template work into
+ * chunks. Applied to an element, it will schedule a task with the given RxRenderStrategy
+ * in order to postpone the template creation of this element.
+ *
+ * ### Features of `*rxChunk`
+ *
+ * - lightweight alternative to `*rxLet="[]"`
+ * - renderCallback
+ * - no value binding
+ * - no context variables
+ * - zone agnostic
+ * - suspense template
+ *
+ * @docsCategory ChunkDirective
+ * @docsPage ChunkDirective
+ * @publicApi
+ */
+@Directive({ selector: '[rxChunk]', standalone: true })
+export class RxChunk implements OnInit, OnDestroy {
+  #strategyProvider = inject<RxStrategyProvider<string>>(RxStrategyProvider);
+
+  /**
+   * @description
+   * The rendering strategy with which the chunk directive should schedule the view
+   * creation. If no strategy is defined, `*rxChunk` will use the configured
+   * `primaryStrategy` as fallback instead.
+   *
+   * @example
+   * \@Component({
+   *   selector: 'app-root',
+   *   template: `
+   *     <div *rxChunk="strategy">
+   *       chunked template
+   *     </div>
+   *   `
+   * })
+   * export class AppComponent {
+   *   strategy = 'low';
+   * }
+   *
+   * @param { RxStrategyNames<string> } strategy
+   */
+  @Input('rxChunk') strategy = this.#strategyProvider.primaryStrategy;
+
+  /**
+   * @description
+   * Setting the `patchZone` to a falsy value will cause the `*rxChunk` directive
+   * to create the `EmbeddedView` outside of `NgZone`.
+   *
+   * @example
+   * \@Component({
+   *   selector: 'app-root',
+   *   template: `
+   *     <div *rxChunk="patchZone: false">
+   *       chunked template out of NgZone
+   *     </div>
+   *   `
+   * })
+   * export class AppComponent { }
+   *
+   * @param { boolean } patchZone
+   */
+  @Input('rxChunkPatchZone') patchZone =
+    this.#strategyProvider.config.patchZone;
+
+  /**
+   * @description
+   * A template to show while the chunk directive is waiting for scheduled task
+   * with the given `RenderStrategy`.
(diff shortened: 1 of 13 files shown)


libs/template/chunk/src/lib/chunk.directive.ts
```diff
@@ -0,0 +1,174 @@
...
+   *       // the div is now accessible
+   *     })
+   *   }
+   * }
+   *
+   *
+   * @param {NextObserver<void>} renderCallback
+   */
+  @Input('rxChunkRenderCallback')
+  set renderCallback(renderCallback: NextObserver<void>) {
```
Karnaukhov-kh, line 131: Is it possible to use not only subjects but also just a function? I know it's more like a general question for render callbacks but would it worth the try?

libs/template/chunk/src/lib/chunk.directive.ts
```diff
@@ -0,0 +1,174 @@
...
+   *     </div>
+   *   `
+   * })
+   * export class AppComponent {
+   *   strategy = 'low';
+   * }
+   *
+   * @param { RxStrategyNames<string> } strategy
+   */
+  @Input('rxChunk') strategy = this.strategyProvider.primaryStrategy;
```
Karnaukhov-kh, line 58: Is it planned to work with not concurrent strategies?

libs/template/chunk/src/lib/tests/chunk.directive.spec.ts
```diff
@@ -0,0 +1,82 @@
...
+  });
+
+  describe.each([
+    [undefined] /* <- Invalid strategy should fallback. */,
+    [''] /* <- Same here. */,
+    ['invalid'] /* <- Same here. */,
+    ['immediate'],
+    ['userBlocking'],
+    ['normal'],
+    ['low'],
```
Karnaukhov-kh, line 51: If we plan to work with not concurrent strats, should we test it?

Karnaukhov-kh (approved):
Hi @hoebbelsB really like the idea. Looks super interesting!

libs/template/chunk/src/lib/tests/chunk.directive.spec.ts
BioPhoton: Yes we should test normal too

libs/template/chunk/src/lib/chunk.directive.ts
hoebbelsB: sure, but why would u? it basically makes the `rxChunk` do nothing useful anymore :D

libs/template/chunk/src/lib/chunk.directive.ts
hoebbelsB: I like the idea, I would even more like to just have an `@Output`. But it looks like it's not supported, at least for the `*` syntax. https://github.com/angular/angular/issues/12121

edbzn (approved):

libs/template/chunk/src/lib/tests/chunk.directive.spec.ts
hoebbelsB: done

libs/template/chunk/src/lib/chunk.directive.ts
Karnaukhov-kh: Yes, that's the thing. Can we type this input to RxConcurrentStrategyNames?

edbzn:
@hoebbelsB shouldn't we wait for the first stable release before introducing new elements?

BioPhoton:
As not all strategies chunk let's call it rxTemplate?

libs/template/chunk/src/lib/chunk.directive.ts
BioPhoton: We have to support all Strategien. Other strategies then concurrent will have an effect definitely. U could e.g. use 'noop' strategy when out of viewport

ChristopherPHolder: There is a need to support all strategies. If you want to change the rendering strategy conditionally, then this is necessary. 

And it makes sense in many cases to switch the strategies based on conditions like first render and other things which may affect CLS

ChristopherPHolder:
Hi, the current workaround for this is to do *rxLet="[]" 

Which from anyone who is not aware of the pattern makes 0 sense. 

Is there anything blocking the issue? 

Is it only the name that's blocking? 

`*rxChunk` - `*rxTemplate` - `*rxRender`

AdrianRomanski:
@Karnaukhov-kh @hoebbelsB  Everything is looking solid :) 

***rxChunk** is best name in my opinion - referring to Christopher last comment. 

That feature have a great marketing potential, to be announced or either be shown in a blog post. 
Lets go and merge it :D

ChristopherPHolder:
@Karnaukhov-kh @hoebbelsB @AdrianRomanski 

Just noticed that this MR has been open since before standalone 😂

I would be happy to clean that up but i guess its kind of clear that there is not plans to merge this?

libs/template/chunk/src/lib/chunk.directive.ts
```diff
@@ -0,0 +1,178 @@
...
+   *     <div *rxChunk="patchZone: false">
+   *       chunked template out of NgZone
+   *     </div>
+   *   `
+   * })
+   * export class AppComponent { }
+   *
+   * @param { boolean } patchZone
+   */
+  @Input('rxChunkPatchZone') patchZone =
```
eneajaho, line 81: Do we still want this one to be at input level, instead of having it at the provider level?

hoebbelsB: well, it's anyway on provider level. I guess we can get rid of it

hoebbelsB: we could get rid of it, yeah

Issue #1386, opened 2022-08-05
Replace usage of `ɵmarkDirty`


edbzn:
# Problem Solved By The Feature

The private `ɵmarkDirty` API was removed from the Angular core and will land in v14.2.0 so we need to refactor our global strategy.

# Solution

I think we can use `ApplicationRef#tick` instead.

# Additional Context

This PR https://github.com/angular/angular/pull/46806 removes the private export.

hoebbelsB:
interesting idea to use `ApplicationRef#tick`. But we need to make sure to coalesce it properly, otherwise we run into heavy over-renderings. Also we don't have access to AppRef inside the strategycredentials.
I would rather suggest ditching the `global` strategy and migrate it to `native`. I guess it isn't in use anyway :shrug:

edbzn:
If we have no access to the ApplicationRef in the strategy credentials, then I agree it would be easier to drop it. Do we want to create a migration script for that? or just drop it? I'm in favor of dropping it.

hoebbelsB:
we can also just point global to native strategy and deprecate it

muuvmuuv:
With the latest version, I get this, which might be related:

```
./node_modules/.pnpm/@rx-angular+cdk@1.0.0-beta.3_vurzrkwbihck2hvr3ls5yaqlwm/node_modules/@rx-angular/cdk/fesm2015/cdk-render-strategies.js:148:24-34 - Error: export 'ɵmarkDirty' (imported as 'ɵmarkDirty') was not found in '@angular/core' (possible exports: ANALYZE_FOR_ENTRY_COMPONENTS, ANIMATION_MODULE_TYPE, APP_BOOTSTRAP_LISTENER, .........)
```


    "@rx-angular/cdk": "1.0.0-beta.3"
    "@angular/common": "^14.2.0",
    "@angular/core": "^14.2.0",

twittwer:
Do somebody know of some workaround until this is fixed?
We aren't using the global strategy but cannot start anymore after the Angular 14.2 update.

edbzn:
This has been fixed and released in `@rx-angular/cdk@1.0.0-rc.0`.

PR #1389, opened 2022-08-06
feat(cdk): deprecate global StrategyCredentials and replace with native

hoebbelsB:
# Description

closing #1386 

* drop `markDirty`
* replace global strategy with native
* check if in NgZone in native strategy (otherwise we will overrender brutally)


libs/cdk/render-strategies/src/lib/native-strategies.ts
```diff
@@ -57,7 +48,18 @@ const nativeCredentials: RxStrategyCredentials = {
...
+            ? ngZone.run(() => work())
+            : work()
+        )
+      ),
+};
+
+const globalCredentials: RxStrategyCredentials = {
+  name: 'global',
+  ...nativeCredentials,
 };
```
edbzn, line 63: Do we want to deprecate it and maybe drop it from the doc?

hoebbelsB: done, please have a look!

edbzn (approved):

Karnaukhov-kh (approved):

PR #1390, opened 2022-08-06
chore: regenerate stale lockfile

LayZeeDK:
When regenerating the Yarn lockfile, tests in the `state` and `cdk` projects fail.


hoebbelsB:
thanks @LayZeeDK, I will check it out soon!

PR #1391, opened 2022-08-08
CI update

edbzn:
- update CI node engine to v16
- bump workflow actions
- remove unused release action
- update node requirement
- update yarn cache


package.json
```diff
@@ -125,6 +125,10 @@
     "ts-jest": "27.0.7",
     "ts-node": "10.4.0",
     "typescript": "4.3.5"
+  },
+  "engines": {
+    "node": ">=16.0.0",
+    "yarn": ">=1.9.2"
```
hoebbelsB, line 131: is this stable with angular 12? according to @LayZeeDK [list](https://gist.github.com/LayZeeDK/c822cc812f75bb07b7c55d07ba2719b3) angular 12 doesn't support node 16 officially?

edbzn: It seems to not be a problem as they updated the engine to be less restrictive, see https://github.com/angular/angular/issues/42076

LayZeeDK: Neither Angular 12.x or Nx 12.x has official support for Node.js 16.

When we read the comment description of https://github.com/iRealNirmal/angular/commit/7f502037ad8bb3bf5e972b92da1ab72e95c3309a, it states that:

>While this change will no longer cause yarn or npm to fail these engine's check, this does
not reflect a change in the officially supported versions of node for Angular.  Angular
continues to maintain support for **Active LTS and Maintenance LTS versions of node**.

.github/actions/setup/action.yml
```diff
@@ -4,35 +4,25 @@ inputs:
...
-    - uses: actions/cache@v2
-      id: yarn-cache
-      with:
-        path: |
-          node_modules
-          ~/.cache/Cypress
-        key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
-        restore-keys: |
-          ${{ runner.os }}-yarn-
+        cache: 'yarn'
```
LayZeeDK, line 23: question: are we intentionally removing the Cypress cache?

.github/actions/setup/action.yml
```diff
@@ -4,35 +4,25 @@ inputs:
   node-version:
     description: Node.js version
     required: false
-    default: '14'
+    default: '16'
```
LayZeeDK, line 7: issue(blocking): as discussed in https://github.com/rx-angular/rx-angular/pull/1391#pullrequestreview-1064696394, we should hold back a full Node.js 16.x migration until we have migrated to at least Angular 13.x **and** Nx 13.x.

libs/cdk/package.json
```diff
@@ -6,11 +6,6 @@
   "publishConfig": {
     "access": "public"
   },
-  "engines": {
-    "node": ">=10.12.0",
-    "npm": ">=5.3.0",
-    "yarn": ">=1.9.2"
-  },
```
LayZeeDK, line 13: question: are we intentionally removing engine version ranges from our library packages?

libs/state/package.json
```diff
@@ -6,11 +6,6 @@
   "publishConfig": {
     "access": "public"
   },
-  "engines": {
-    "node": ">=10.12.0",
-    "npm": ">=5.3.0",
-    "yarn": ">=1.9.2"
-  },
```
LayZeeDK, line 13: question: are we intentionally removing engine version ranges from our library packages?

libs/template/package.json
```diff
@@ -6,11 +6,6 @@
   "publishConfig": {
     "access": "public"
   },
-  "engines": {
-    "node": ">=10.12.0",
-    "npm": ">=5.3.0",
-    "yarn": ">=1.9.2"
-  },
```
LayZeeDK, line 13: question: are we intentionally removing engine version ranges from our library packages?

LayZeeDK (changes requested):
Discussions to address.

package.json
BioPhoton: Sad!

libs/template/package.json
BioPhoton: IMHO we should keep it / add it again.

package.json
edbzn: @hoebbelsB does something prevent us to migrate to Nx 13 and Angular 13? From what I can recall esm modules break jest setup, isn't it?

LayZeeDK: > From what I can recall esm modules break jest setup, isn't it?

Jest might need some additional setting and I think it's less of a problem in recent major versions of Jest. Not something we can't figure out though.

PR #1394, opened 2022-08-11
New ESLint plugin

matejchalk:
Replaces previous PR #1124 (accidentally closed).


Karnaukhov-kh:
A-M-A-Z-I-N-G

libs/eslint-plugin/docs/rules/no-zone-critical-rxjs-creation-apis.md
```diff
@@ -0,0 +1,51 @@
...
+<br />
+
+<details>
+<summary>✅ Examples of correct code for this rule</summary>
+
+```ts
+import { interval } from '@rx-angular/cdk/zone-less';
+
+const source = interval(1000);
+const subscribe = source.subscribe((val) => console.log(val));
```
Karnaukhov-kh, line 48: This one I guess not relevant anymore since we removed zone-less/rxjs and having it as separate deprecated package. Maybe something with getZoneUnpatchedApi will be a better fit

libs/eslint-plugin/docs/rules/no-zone-critical-rxjs-operators.md
```diff
@@ -0,0 +1,70 @@
...
+import { mergeMap, takeUntil } from 'rxjs/operators';
+import { delay } from '@rx-angular/cdk/zone-less';
+
+const mousedown$ = fromEvent(document, 'mousedown');
+const mouseup$ = fromEvent(document, 'mouseup');
+
+mousedown$
+  .pipe(mergeMap((event) => of(event).pipe(delay(700), takeUntil(mouseup$))))
+  .subscribe((event) => console.log('Long Press!', event));
+```
```
Karnaukhov-kh, line 68: Same as creation APIs

libs/eslint-plugin/docs/rules/no-zone-critical-rxjs-schedulers.md
```diff
@@ -0,0 +1,48 @@
...
+<details>
+<summary>✅ Examples of correct code for this rule</summary>
+
+```ts
+import { interval } from 'rxjs';
+import { observeOn } from 'rxjs/operators';
+import { asyncScheduler } from '@rx-angular/cdk/zone-less';
+
+interval(10).pipe(observeOn(asyncScheduler)).subscribe(console.log);
+```
```
Karnaukhov-kh, line 46: Same, zone-less rxjs deprecated

edbzn:
Dope! :fire:

libs/eslint-plugin/docs/rules/no-zone-critical-rxjs-creation-apis.md
matejchalk: Oh right, so the import would have to be from `@rx-angular/cdk/zone-less/rxjs` now? But if that package is deprecated anyway (\*), then what is now the zone-less way to use these RxJS creators/operators/schedulers? I'm struggling to think of a "correct" example otherwise. Does it still make sense to have these zone-less RxJS rules?

---

(\*) BTW if zone-less RxJS is deprecated, then the docs are very misleading. E.g. [zone-less docs still have an RxJS section](https://github.com/rx-angular/rx-angular/tree/main/libs/cdk/zone-less/docs#usage) and [the standalone package docs mention nothing about deprecation either](https://github.com/rx-angular/rx-angular/tree/main/libs/cdk/zone-less/rxjs). I could only find JSDoc `@deprecated will be removed with 1.0.0` comments in the code itself, but that's only because I went looking for it after your comment :sweat_smile:

Karnaukhov-kh: We moved package [here](https://github.com/rx-angular/rxjs-zone-less). Guess we missed docs part. Yeah, as for example no clue honestly. Maybe we don't need this rules anymore. @hoebbelsB @BioPhoton @edbzn what do you think? Should we keep this rules just to "frighten" :D people? Or maybe we can come up with some example?

hoebbelsB: I think the rule is still valid. We need to update the docs and the rule can just point to the other package?

matejchalk: Ok, I updated the examples.

Karnaukhov-kh (approved):

Issue #311, opened 2020-09-12
Docs for select operator are not clear in taking operators


BioPhoton:
https://twitter.com/iAbu94/status/1304843704952262656?s=20

BioPhoton:
```typescript
select(
   myOperator(),
   // filter undefined
   // distinct until changed
   // shareReplay
)
```

Issue #263, opened 2020-08-26
Proposal for a better version of the select operator


BioPhoton:
# Problem solved by the feature

ATM we have 2 main operators/ways to shortcut state selections. `select` and `selectSlice`.

## `select` 
`select` is a standalone operator as well as a method on `RxState`. It plucks a (deeply) nested property or takes a series of operators the derive state.

A string-based version that takes a series of keys, reflecting a deeply nested value in an object: 
```typescript
 select<K1 extends keyof T, K2 extends keyof T[K1]>(
    k1: K1,
    k2: K2,
   // more property names
  ): Observable<T[K1][K2][...]>;
```
And a reactive version that takes a series of functions that adhere to the signature of an RxJS operator (`(Observable<T>) => Observable<T|O>`)

```typescript
select<A = T, B = A>(
    op1: OperatorFunction<T, A>,
    op2: OperatorFunction<A, B>,
    // more operators
  ): Observable<D>;
```

Pros:
- Both versions turn the stream into a "stateful stream", a stream that emits only distinct values and replays the last one on subscribe.
- The string version is super nice to pluck a single property from the state and get a single distinct replayed observable of it.
- The operator version can compose any RxJS operator to derive state and apply behavior. 

Cons:
- The string version does not allow to select a subset of the shallow properties of an object. (only nested properties)
- No way to finetune distinct values

## `selectSlice` 
`selectSlice` is a standalone operator that takes a set of properties and an optional set of compare functions to select a state slice. 

The shallow property names:

```typescript
selectSlice<T extends object, K extends keyof T>(
  keys: K[]
): OperatorFunction<T, PickSlice<T, K>> 
```

And an optional compare map:

```typescript
selectSlice<T extends object, K extends keyof T>(
  keys: K[],
  keyCompareMap: KeyCompareMap<{ [P in K]: T[P] }>
): OperatorFunction<T, PickSlice<T, K>> 
```

Pros:
- Perfect for deriving a slice of the state (a subset of the shallow properties).
- Easy to tweak what distinct values from another. (`KeyCompareMap`)

Cons:
- The resulting stream is not "stateful"
- No options to pass operators

The problem here is that `selectSlice` is often used within select like this:

```typescript
state.select(
    selectSlice(['prop1', 'prop2', 'prop3']),
    map(s => ({
        ...s,
        num: Object.keys(prop2).length
    })),
    // optionally used with transformationHelpers
[...]

PR #1396, opened 2022-08-14
Add overloads to select for more concise state projections

matejchalk:
This PR adds two overloads to RxState's `select` method (and operator), which I think would make state projections more concise.

What I use `select` for most frequently is transforming one or more state properties reactively into the shape I need for my template. In other words, to project state using the declarative `f(state) = UI` approach. I often need a custom projection function for this (=> `map` operator), and also want the automatic optimization of only emitting changes to state props it depends on (=> `select` operator if single key, or `selectSlice` operator if multiple keys). This means I often write code like this:

```ts
isLoaded$ = this.state.select(
  select('loading'),
  map(loading => !loading),
);

filteredItems$ = this.state.select(
  selectSlice(['items', 'textFilter']),
  map(({ items, textFilter }) => items.filter(item => item.title.includes(textFilter))),
);
```

Due to how common projecting state is, I feel like the syntax for this could be shorter. So in this PR, I'd like to add two overloads to `select` that would enable this code to be more concise:

```ts
isLoaded$ = this.state.select('loading', loading => !loading);

filteredItems$ = this.state.select(
  ['items', 'textFilter'],
  ({ items, textFilter }) => items.filter(item => item.title.includes(textFilter)),
);
```

I'd say the pros of adding these overloads are:
- makes state projections more concise and therefore more declarative,
- non-breaking changes to `select`'s API,
- performance optimizations under the hood,
- feels similar to NgRx's `createSelector` (projector function + memoization),
- less reason to use `state.select(...).pipe(...)` pattern I've noticed with others (possibly related to #311).

I guess the main con would be that `select` already has many overloads, so this might complicate the interface. (There's an older discussion in #263 about changes to `select` or `selectSlice` which seems relevant.)

### DIFF
--- a/libs/state/selections/spec/select.spec.ts
+++ b/libs/state/selections/spec/select.spec.ts
@@ -128,6 +128,24 @@ describe('select', () => {
     });
   });
 
+  it('should accept array of strings keyof T and one map function', () => {
+    testScheduler.run(({ cold, expectObservable }) => {
+      const primitiveState: PrimitiveState = {
+        bol: true,
+        str: 'string',
+        num: 42,
+      };
+      const source: Observable<PrimitiveState> = cold('a|', {
+        a: primitiveState,
+      });
+      expectObservable(
+        source.pipe(select(['num', 'str'], ({ num, str }) => `${str} (${num})`))
+      ).toBe('a|', {
+        a: 'string (42)',
+      });
+    });
+  });
+
   it('should accept one operator', () => {
     testScheduler.run(({ cold, expectObservable }) => {
       const source: Observable<PrimitiveState> = cold('a|', {
--- a/libs/state/selections/spec/select.spec.ts
+++ b/libs/state/selections/spec/select.spec.ts
@@ -1,6 +1,4 @@
 // eslint-disable-next-line @nrwl/nx/enforce-module-boundaries
-import { select } from '@rx-angular/state/selections';
-// eslint-disable-next-line @nrwl/nx/enforce-module-boundaries
 import {
   initialNestedState,
   initialPrimitiveState,
@@ -11,6 +9,7 @@ import {
 import { EMPTY, NEVER, Observable, of } from 'rxjs';
 import { map } from 'rxjs/operators';
 import { TestScheduler } from 'rxjs/testing';
+import { select } from '../src/lib/operators/select';
 
 let testScheduler: TestScheduler;
 
--- a/libs/state/selections/spec/select.spec.ts
+++ b/libs/state/selections/spec/select.spec.ts
@@ -1,16 +1,16 @@
 // eslint-disable-next-line @nrwl/nx/enforce-module-boundaries
+import { select } from '@rx-angular/state/selections';
+// eslint-disable-next-line @nrwl/nx/enforce-module-boundaries
 import {
-  jestMatcher,
   initialNestedState,
   initialPrimitiveState,
+  jestMatcher,
   NestedState,
   PrimitiveState,
 } from '@test-helpers';
 import { EMPTY, NEVER, Observable, of } from 'rxjs';
 import { map } from 'rxjs/operators';
 import { TestScheduler } from 'rxjs/testing';
-// eslint-disable-next-line @nrwl/nx/enforce-module-boundaries
-import { select } from '@rx-angular/state/selections';
 
 let testScheduler: TestScheduler;
 
@@ -109,6 +109,22 @@ describe('select', () => {
       expectObservable(
         source.pipe(select('obj', 'key1', 'key11', 'key111'))
       ).toBe('a|', { a: 'test' });
+    });
+  });
+
+  it('should accept one string keyof T and one map function', () => {
+    testScheduler.run(({ cold, expectObservable }) => {
+      const primitiveState: PrimitiveState = {
+        bol: true,
+        str: 'string',
+        num: 42,
+      };
+      const source: Observable<PrimitiveState> = cold('a|', {
+        a: primitiveState,
+      });
+      expectObservable(source.pipe(select('num', (x) => x / 2))).toBe('a|', {
+        a: 21,
+      });
     });
   });
 
--- a/libs/state/selections/spec/select.spec.ts
+++ b/libs/state/selections/spec/select.spec.ts
@@ -145,6 +145,43 @@ describe('select', () => {
     });
   });
 
+  it('should accept array of strings keyof T, map function and key compare map', () => {
+    testScheduler.run(({ cold, expectObservable }) => {
+      const state: PrimitiveState & NestedState = {
+        bol: true,
+        str: 'string',
+        num: 42,
+        obj: {
+          key1: {
+            key11: {
+              key111: 'foo',
+            },
+          },
+        },
+      };
+      const source: Observable<PrimitiveState & NestedState> = cold('abcde|', {
+        a: state,
+        b: { ...state, bol: false },
+        c: { ...state, num: 69 },
+        d: { ...state, num: 69, obj: { key1: { key11: { key111: 'foo' } } } },
+        e: { ...state, num: 69, obj: { key1: { key11: { key111: 'bar' } } } },
+      });
+      expectObservable(
+        source.pipe(
+          select(
+            ['num', 'obj'],
+            ({ num, obj }) => `${num}: ${obj.key1.key11.key111}`,
+            { obj: (a, b) => a.key1.key11.key111 === b.key1.key11.key111 }
+          )
+        )
+      ).toBe('a-c-e|', {
+        a: '42: foo',
+        c: '69: foo',
+        e: '69: bar',
+      });
+    });
+  });
+
   it('should accept one operator', () => {
     testScheduler.run(({ cold, expectObservable }) => {
       const source: Observable<PrimitiveState> = cold('a|', {
--- a/libs/state/selections/src/lib/operators/select.ts
+++ b/libs/state/selections/src/lib/operators/select.ts
@@ -1,6 +1,10 @@
 import { MonoTypeOperatorFunction, Observable, OperatorFunction } from 'rxjs';
-import { pluck } from 'rxjs/operators';
-import { isOperateFnArrayGuard, isStringArrayGuard } from '../utils/guards';
+import { map, pluck } from 'rxjs/operators';
+import {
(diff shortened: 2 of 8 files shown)


libs/state/selections/src/lib/operators/select.ts
```diff
@@ -150,11 +202,25 @@ export function select<T,
...
+    | string[]
+    | [string, (val: any) => any]
+    | [string[], (slice: any) => any]
 ): OperatorFunction<T, any> {
   return (state$: Observable<T>) => {
     if (!opOrMapFn || opOrMapFn.length === 0) {
       return state$.pipe(stateful());
+    } else if (isStringAndFunctionTupleGuard(opOrMapFn)) {
+      return state$.pipe(
+        stateful(pluck(opOrMapFn[0])),
```
Karnaukhov-kh, line 216: I would not go with pluck here since it is deprecated.

libs/state/src/lib/rx-state.service.ts
```diff
@@ -547,10 +583,24 @@ export class RxState<T extends object> implements OnDestroy, Subscribable<T> {
...
+      | OperatorFunction<T, R>[]
+      | string[]
+      | [string, (val: any) => R]
+      | [string[], (slice: any) => R]
   ): Observable<T | R> {
     if (!opOrMapFn || opOrMapFn.length === 0) {
       return this.accumulator.state$.pipe(stateful());
+    } else if (isStringAndFunctionTupleGuard(opOrMapFn)) {
+      return this.accumulator.state$.pipe(
+        stateful(pluck(opOrMapFn[0])),
```
Karnaukhov-kh, line 596: Same, better to avoid pluck.

Karnaukhov-kh (commented):
Great idea @matejchalk. Strongly believe that this is an amazing addition to the selection possibilities. Can you please extend documentation [here](https://github.com/rx-angular/rx-angular/blob/main/libs/state/docs/api/rx-state.md#signature-7) and [here](https://github.com/rx-angular/rx-angular/blob/main/libs/state/selections/docs/operators/select.md#select) with the new usage example and the ones we missed  😅?

P.S. @hoebbelsB top level seelctSlice is real :D.

libs/state/selections/src/lib/operators/select.ts
```diff
@@ -150,11 +202,25 @@ export function select<T,
  * @internal
  */
 export function select<T>(
-  ...opOrMapFn: OperatorFunction<T, any>[] | string[]
+  ...opOrMapFn:
+    | OperatorFunction<T, any>[]
+    | string[]
+    | [string, (val: any) => any]
+    | [string[], (slice: any) => any]
```
edbzn, line 209: What about using `unknown` instead of `any` here?

libs/state/selections/src/lib/operators/select.ts
```diff
@@ -18,6 +25,43 @@ import { stateful } from './stateful';
...
+ *     ({ query, results }) => `${results.length} results found for "${query}"`
+ *   )
+ * );
+ *
+ * @return Observable<R>
+ */
+export function select<T extends object, K extends keyof T, R>(
+  keys: K[],
+  fn: (slice: PickSlice<T, K>) => R
+): OperatorFunction<T, R>;
```
edbzn, line 47: Not directly related to your PR but a general thought on the codebase, we shouldn't name generics with a single letter as it becomes hard to read and not really meaningful.

libs/state/selections/spec/select.spec.ts
```diff
@@ -1,16 +1,16 @@
 // eslint-disable-next-line @nrwl/nx/enforce-module-boundaries
+import { select } from '@rx-angular/state/selections';
+// eslint-disable-next-line @nrwl/nx/enforce-module-boundaries
```
edbzn, line 3: I think we can use a relative import here and remove the eslint disabled rule.

edbzn (commented):
We discussed that recently with @LayZeeDK, I'm wondering if this might be helpful for auto-refactoring tools, or if the first keys argument is still problematic.

libs/state/selections/src/lib/operators/select.ts
hoebbelsB: i would prefer it as well if we are unable to provide a meaningful type

libs/state/selections/src/lib/operators/select.ts
```diff
@@ -150,11 +202,25 @@ export function select<T,
...
+    | [string, (val: any) => any]
+    | [string[], (slice: any) => any]
 ): OperatorFunction<T, any> {
   return (state$: Observable<T>) => {
     if (!opOrMapFn || opOrMapFn.length === 0) {
       return state$.pipe(stateful());
+    } else if (isStringAndFunctionTupleGuard(opOrMapFn)) {
+      return state$.pipe(
+        stateful(pluck(opOrMapFn[0])),
+        stateful(map(opOrMapFn[1]))
```
hoebbelsB, line 217: can we combine those two somehow? stateful itself applies quite some operators and I guess we do not need to duplicate all of them

libs/state/src/lib/rx-state.service.ts
```diff
@@ -547,10 +583,24 @@ export class RxState<T extends object> implements OnDestroy, Subscribable<T> {
    * @internal
    */
   select<R>(
```
hoebbelsB, line 585: is there a reason why we just don't re-use `select` here? Just came to my mind

libs/state/src/lib/rx-state.service.ts
```diff
@@ -547,10 +583,24 @@ export class RxState<T extends object> implements OnDestroy, Subscribable<T> {
...
+      | string[]
+      | [string, (val: any) => R]
+      | [string[], (slice: any) => R]
   ): Observable<T | R> {
     if (!opOrMapFn || opOrMapFn.length === 0) {
       return this.accumulator.state$.pipe(stateful());
+    } else if (isStringAndFunctionTupleGuard(opOrMapFn)) {
+      return this.accumulator.state$.pipe(
+        stateful(pluck(opOrMapFn[0])),
+        stateful(map(opOrMapFn[1]))
```
hoebbelsB, line 597: same question about duplicated usage of `stateful`

libs/state/selections/src/lib/operators/select.ts
matejchalk: Didn't know that, thanks :+1:

libs/state/selections/src/lib/operators/select.ts
matejchalk: Good point :+1: Fixed it.

libs/state/src/lib/rx-state.service.ts
matejchalk: You mean why doesn't the `RxState#select` method just apply the `select` operator with the same arguments? That's a very good question, I don't know why the implementations are duplicated :smile: Was there some reason for it?

libs/state/selections/spec/select.spec.ts
matejchalk: Good catch, fixed.

libs/state/selections/src/lib/operators/select.ts
matejchalk: Done :+1:

matejchalk:
> hey @matejchalk I love that PR, thank you very much! We wanted top-level `selectSlice` for a long time now, it's finally happening heart
> 
> I've left some comments with feedback for you!
> 
> Also, do you think it's possible and beneficial to also the `keyCompareMap` argument to `select` in case you are selecting multiple keys?
> 
> Technically one could also just use `distinctUntilSomeChanged`, but I wanted to at least share my thought about this.

Thanks, glad you like it :blush: 

Regarding the `keyCompareMap`, I've actually never ended up using it with `selectSlice` so far, so I'm not sure how common it is. I guess it could be an optional extra argument, probably making it 2nd out of 3 parameters would be more intuitive than 3rd (`select(keys, keyCompareMap, projectFn)` vs `select(keys, projectFn, keyCompareMap)`)? :thinking: But if it wouldn't often be used like that, then it might not be worth complicating the interface, since you can always apply the RxJS operators :man_shrugging:

libs/state/src/lib/rx-state.service.ts
hoebbelsB: that's what i'm asking myself ^^ is it possible to switch it? It's maybe the case because it used lived in the cdk package

hoebbelsB:
> Regarding the `keyCompareMap`, I've actually never ended up using it with `selectSlice` so far, so I'm not sure how common it is. I guess it could be an optional extra argument, probably making it 2nd out of 3 parameters would be more intuitive than 3rd (`select(keys, keyCompareMap, projectFn)` vs `select(keys, projectFn, keyCompareMap)`)? thinking But if it wouldn't often be used like that, then it might not be worth complicating the interface, since you can always apply the RxJS operators man_shrugging

If we want to introduce it, i'd suggest to add it as 3rd argument since the projectFn will be used def. more often.

@Karnaukhov-kh @edbzn @BioPhoton any opinion on adding `KeyCompareMap` also as argument for the new select overload?

libs/state/src/lib/rx-state.service.ts
matejchalk: OK, I've refactored it :heavy_check_mark:

hoebbelsB: great job!!!

matejchalk:
> If we want to introduce it, i'd suggest to add it as 3rd argument since the projectFn will be used def. more often.

@hoebbelsB Ok, I've added it.

> Can you please extend documentation [here](https://github.com/rx-angular/rx-angular/blob/main/libs/state/docs/api/rx-state.md#signature-7) and [here](https://github.com/rx-angular/rx-angular/blob/main/libs/state/selections/docs/operators/select.md#select) with the new usage example and the ones we missed sweat_smile?

@Karnaukhov-kh [Done](https://github.com/flowup/rx-angular/blob/extend-select/libs/state/docs/api/rx-state.md#signature-9) and [done](https://github.com/flowup/rx-angular/blob/extend-select/libs/state/selections/docs/operators/select.md#select) :heavy_check_mark:

hoebbelsB (approved):
@matejchalk u are a hero! Will merge as soon as I'm back from Prague!

Issue #1335, opened 2022-05-16
RxIf can output undefined


Karnaukhov-kh:
# Description

Main idea of `ngIf` directive is to make sure that value is present. With `rxIf` we should have the same "mindset". Currently since we're using `createTemplateNotifier` that potentially can produce falsy values it is not the case.

Also `suspense` for `rxIf` might be unnecessary because same effect can be achieved with `else` template.

# Example

We have a new angular application generated with default settings.

Let's say we have a card component with card input that is not defined by default.

```typescript
export class CardComponent {
  @Input() card?: Card;
}
```

Normal case will be to have next check in the template template

![image](https://user-images.githubusercontent.com/15088626/168630477-675b90c2-fef3-4290-b7bd-f031bf50867b.png)

```html
<div  class="card"  *ngIf="card">
  <div class="card-header">{{ card.cost }}</div>
  <h1 class="card-name">{{ card.name }}</h1>
  <p class="card-text">{{card.text}}</p>
</div>
```

However with `rxIf` at the moment value still can be undefined and we need to make optional chaining everywhere.

![image](https://user-images.githubusercontent.com/15088626/168630550-18624bfd-0e17-43e9-b0f6-75b00f802cb0.png)

```html
<div  class="card"  *rxIf="card">
  <div class="card-header">{{ card?.cost }}</div>
  <h1 class="card-name">{{ card?.name }}</h1>
  <p class="card-text">{{card?.text}}</p>
</div>
```

# Environment

Ivy Workspace: Yes
Editor: VSCode

Package                           Version
-----------------------------------------------------------
@angular-devkit/architect       0.1303.5
@angular-devkit/build-angular   13.3.5
@angular-devkit/core            13.3.5
@angular-devkit/schematics      13.3.5
@angular/cdk                    13.3.7
@angular/cli                    13.3.5
@schematics/angular             13.3.5
rxjs                            7.5.5
typescript                      4.6.4

Karnaukhov-kh:
@hoebbelsB can't reproduce it anymore, looks good.

PR #1400, opened 2022-08-28
feat(template): Make rxIf stable

hoebbelsB:
# Description

* [x] adjust `rxIf` API to be in line with ngIf
* [x] implemented rxIf test-suite based on original [ngIfs test-suite](https://github.com/angular/angular/blob/main/packages/common/test/directives/ng_if_spec.ts)
* [x] adjust rxIf template handling in order to properly display values
* [x] find a solution for not implemented ability to switch templates dynamically
  * `should support removing the then/else templates`
  * `should support dynamic else`
* [x] implement test cases for suspense, complete, error
* [x] implement test cases for observable and strategy handling
* [x] create JS docs for `rxIf` 
* [x] move `rxIf` to `@rx-angular/template/if`


edbzn (approved):

libs/template/experimental/if/src/lib/if.directive.ts
```diff
@@ -48,21 +48,20 @@ export class RxIf<U> implements OnInit, OnDestroy, OnChanges {
     rxIfTemplateNames
   >;
 
-  @Input()
-  set rxIf(potentialObservable: Observable<U> | U | null | undefined) {
-    this.observablesHandler.next(coerceObservable(potentialObservable));
-  }
+  private lastValidValue: U | null;
```
edbzn, line 51: General thought, what about using the private instance fields syntax? So we make sure these fields are not accessible at runtime.

```suggestion
  #lastValidValue: U | null;
```

libs/template/experimental/if/src/lib/tests/if.directive.spec.ts
```diff
@@ -0,0 +1,384 @@
+import { ComponentFixture, TestBed, waitForAsync } from '@angular/core/testing';
+import { By } from '@angular/platform-browser';
+import { RX_RENDER_STRATEGIES_CONFIG } from '@rx-angular/cdk/render-strategies';
+import { IfModule } from '../if.module';
+import { createTestComponent, TestComponent } from './fixtures';
+
+{
```
edbzn, line 7: These braces are superfluous.

libs/template/experimental/if/src/lib/tests/if.directive.observable.spec.ts
```diff
@@ -0,0 +1,483 @@
+import { ComponentFixture, TestBed, waitForAsync } from '@angular/core/testing';
+import { By } from '@angular/platform-browser';
+import { RX_RENDER_STRATEGIES_CONFIG } from '@rx-angular/cdk/render-strategies';
+import { BehaviorSubject, of, startWith, throwError, tap } from 'rxjs';
+import { IfModule } from '../if.module';
+import { createTestComponent, TestComponent } from './fixtures';
+
+{
```
edbzn, line 8: These braces are superfluous.

edbzn (approved):

edbzn:
Question, does it fix the following issue https://github.com/rx-angular/rx-angular/issues/1335?

libs/template/experimental/if/src/lib/if.directive.ts
hoebbelsB: If we go that route I think we should do it in a separate PR for all of our directives/components

Karnaukhov-kh (approved):

edbzn:
Tested in my todo mvc app, it's working well!

PR #1406, opened 2022-09-12
feat(template): introduce experimental virtual-scrolling sub-package

hoebbelsB:
Introducing `@rx-angular/template/experimental/virtual-scrolling` submodule :partying_face: :fire: :fire: :fire: 

Heavily inspired by the article ["Complexities of an infinite scroller
" ](https://developer.chrome.com/blog/infinite-scroller/) of @surma i've tried to port his [Proof of Concept implementation ](https://github.com/GoogleChromeLabs/ui-element-samples/tree/gh-pages/infinite-scroller) to angular.

The public API is quite similar to the on of [`@angular/cdk/scrolling`](https://material.angular.io/cdk/scrolling/api#CdkVirtualForOf) `CdkVirtualForOf` et al.

Some gifs showcasing the implementation

## Fixed Size

fixed-size

## Autosized

auto-size

## Autosized scrolling behavior compared to `@angular/cdk/scrolling`

autosize-compare-to-cdk

## Docs

The `@rx-angular/template/experimental/virtual-scrolling` package can be seen as a
high performant implementation of its counterpart from the official `@angular/cdk`.
The API is heavily inspired by `@angular/cdk` and is divided into multiple
core components which have to be glued together:

* `RxVirtualViewRepeater`, implemented by `RxVirtualFor`
* `RxVirtualScrollViewport`, implemented by `RxVirtualScrollViewportComponent`
* `RxVirtualScrollStrategy`, implemented by `AutosizeVirtualScrollStrategy` and `FixedSizeVirtualScrollStrategy`

## RxVirtualFor

The `*rxVirtualFor` structural directive provides a convenient and performant
way for rendering huge lists of items. It brings all the benefits `rxFor` does, and additionally provides virtual scrolling.
Instead of rendering every item provided, rxVirtualFor only renders what is currently visible to the user, thus providing
excellent runtime performance for huge sets of data. The technique to render items is comparable to the one used by twitter and 
explained in very much detail by @DasSurma in his blog post about the
[complexities of infinite scrollers](https://developer.chrome.com/blog/infinite-scroller/).   

"Each recycling of a DOM element would normally relayout the entire runway which would bring us well below our target
of 60 frames per second. To avoid this, we are taking the burden of layout onto ourselves and use absolutely positioned
elements with transforms." (@DasSurma)

### API
The API is a combination of `@rx-angular/template/for` &
`@angular/cdk` `*cdkVirtualFor`.
* trackBy: `(index: number, item: T) => any` | `keyof T`
* strategy: `string` | `Observable<string>`
* parent: `boolean`;
* renderCallback: `Subject<T[]>`
* viewCache: `number`
* (Injected) scrollStrategy: `RxVirtualScrollStrategy<T, U>`
* provides itself as RxVirtualViewRepeater for RxVirtualViewPortComponent to operate

### Features
* Push based architecture
* Comprehensive set of context variables
* Opt-out of `NgZone` with `patchZone`
* Notify when rendering of child templates is finished (`renderCallback`)
* Super efficient layouting with css transformations
* Define a viewCache in order to re-use views instead of re-creating them
* Configurable RxVirtualScrollStrategy<T, U> providing the core logic to calculate the viewRange and position DOM
  Nodes

### Context Variables

The following context variables are available for each template:

- $implicit: `T` // the default variable accessed by `let val`
- item$: `Observable<T>` // the same value as $implicit, but as `Observable`
- index: `number` // current index of the item
- count: `number` // count of all items in the list
- first: `boolean` // true if the item is the first in the list
- last: `boolean` // true if the item is the last in the list
- even: `boolean` // true if the item has on even index (index % 2 === 0)
- odd: `boolean` // the opposite of even
- index$: `Observable<number>` // index as `Observable`
- count$: `Observable<number>` // count as `Observable`
- first$: `Observable<boolean>` // first as `Observable`
- last$: `Observable<boolean>` // last as `Observable`
- even$: `Observable<boolean>` // even as `Observable`
- odd$: `Observable<boolean>` // odd as `Observable`
- select: `(keys: (keyof T)[], distinctByMap) => Observable<Partial<T>>`
  // returns a selection function which
  // accepts an array of properties to pluck out of every list item. The function returns the selected properties of
  // the current list item as distinct `Observable` key-value-pair. See the example below:

This example showcases the `select` view-context function used for deeply nested lists.

 ```html
<rx-virtual-scroll-viewport>
  <div
   autosized
   *rxVirtualFor="let hero of heroes$; trackBy: trackItem; let select = select;">
    <div>
      <strong>{{ hero.name }}</strong></br>
      Defeated enemies:
    </div>
     <span *rxFor="let enemy of select(['defeatedEnemies']); trackBy: trackEnemy;">
       {{ enemy.name }}
     </span>
  </div>
</rx-virtual-scroll-viewport>
 ```

### Using the context variables

```html
<rx-virtual-scroll-viewport>
 <div
    *rxVirtualFor="
      let item of observableItems$;
      let count = count;
      let index = index;
      let first = first;
      let last = last;
      let even = even;
      let odd = odd;
      trackBy: trackItem;
    "
  >
    <div>{{ count }}</div>
    <div>{{ index }}</div>
    <div>{{ item }}</div>
    <div>{{ first }}</div>
    <div>{{ last }}</div>
    <div>{{ even }}</div>
    <div>{{ odd }}</div>
  </div>
</rx-virtual-scroll-viewport>
```

## RxVirtualScrollViewportComponent

Container component comparable to CdkVirtualScrollViewport acting as viewport for `*rxVirtualFor` to operate on. 
Its main purpose is to implement the `RxVirtualScrollViewport` interface as well as maintaining the scroll runways'
height in order to give the provided `RxVirtualScrollStrategy` room to position items. Furthermore, it will gather and forward
all events to the consumer of `rxVirtualFor`.

## AutosizeVirtualScrollStrategy

The `AutosizeVirtualScrollStrategy` provides a twitter-like virtual-scrolling implementation. It is able to render and position
items based on their individual size. It is comparable to `@angular/cdk/experimental` `AutosizeVirtualScrollStrategy`, but with
a high performant layouting technique and more features. On top of this the `AutosizeVirtualScrollStrategy` is leveraging the
native `ResizeObserver` in order to detect size changes for each individual view rendered to the DOM and properly re-position
accordingly. In order to provide top-notch runtime performance the `AutosizeVirtualScrollStrategy` builds up caches that
prevent DOM interactions whenever possible. Once a view was visited, its properties will be stored instead of re-read from the DOM
again as this can potentially lead to unwanted forced reflows.

## FixedSizeVirtualScrollStrategy

The `FixedSizeVirtualScrollStrategy` provides a  way of rendering items of
a given size. It is comparable to `@angular/cdk` `FixedSizeVirtualScrollStrategy`,
but with a high performant layouting technique.


libs/template/experimental/virtual-scrolling/src/lib/observe-element-size.ts
```diff
@@ -0,0 +1,32 @@
...
+): Observable<T>;
+export function observeElementSize<T>(
+  element: Element,
+  config?: {
+    options?: ResizeObserverOptions;
+    extract?: (entries: ResizeObserverEntry[]) => T;
+  }
+): Observable<T | DOMRectReadOnly> {
+  const extractProp: (entries: ResizeObserverEntry[]) => T | DOMRectReadOnly =
+    config?.extract ? config?.extract : (entries) => entries[0].contentRect;
```
matejchalk, line 24: Could also use null coalescing here:

```suggestion
    config?.extract ?? (entries) => entries[0].contentRect;
```

libs/template/experimental/virtual-scrolling/src/lib/scroll-strategies/autosized-virtual-scroll-strategy.ts
```diff
@@ -0,0 +1,689 @@
...
+  /** @internal */
+  private _contentSize = 0;
+  /** @internal */
+  private set contentSize(size: number) {
+    this._contentSize = size;
+    this._contentSize$.next(size);
+  }
+
+  /** @internal */
+  private readonly _renderedRange$ = new ReplaySubject<ListRange>(1);
```
matejchalk, line 183: Why not use a `BehaviorSubject`? Then you could have synchronous reads via `this._renderedRange$.value` and wouldn't need the extra `_renderedRange` attribute (same goes for `scrolledIndex`). Or am I missing something?

libs/template/experimental/virtual-scrolling/src/lib/scroll-strategies/autosized-virtual-scroll-strategy.ts
```diff
@@ -0,0 +1,689 @@
...
+    );
+    const onScroll$ = this.viewport.elementScrolled$.pipe(
+      map(() => this.viewport.getScrollTop()),
+      distinctUntilChanged(),
+      startWith(0),
+      tap((_scrollTop) => {
+        this.scrollTop = _scrollTop;
+      })
+    );
+    merge(
```
matejchalk, line 339: It looks you're only passing one Observable into `merge`, so I guess it can be omitted.

matejchalk (approved):
Wow, the autosized virtualing scrolling is so much smoother than in the Angular CDK when dragging the scrollbar :+1: Very impressive :clap: 

I must admit a lot of the code is beyond my comprehension, but I left a few minor comments.

libs/template/experimental/virtual-scrolling/src/lib/scroll-strategies/fixed-size-virtual-scroll-strategy.ts
```diff
@@ -0,0 +1,278 @@
...
+  private _setViewPosition(
+    view: EmbeddedViewRef<RxVirtualForViewContext<T, U>>,
+    scrollTop: number
+  ): void {
+    const element = this.getElement(view);
+    element.style.position = 'absolute';
+    element.style.transform = `translateY(${scrollTop}px)`;
+    if (this.enterStyles) {
+      Object.keys(this.enterStyles).forEach((style) => {
+        element.style[style] = this.enterStyles[style];
```
Gabweb, line 264: enterStyles are not working as suggested by the JsDoc at the top. Both the CSS styling (opacity: 0) as well as the enterStyles (opacity: 1) are set in the same frame. Therefore, the CSS transition does not trigger any animation (tested in Chrome 107). Could be solved be executing this code-block in a raf. :) 

Nevertheless, nice work!

libs/template/experimental/virtual-scrolling/src/lib/scroll-strategies/autosized-virtual-scroll-strategy.ts
hoebbelsB: `renderedRange$` is part of the `RxVirtualScrollStrategy` interface, so it's publicly accessible. I didn't want to make a subject public, but just the observable.

hoebbelsB:
close in favor of https://github.com/rx-angular/rx-angular/pull/1539

PR #1445, opened 2022-10-18
chore: fix docs build in CI

LayZeeDK:
The docs build has been flaky and not logged much to output.

- Update GitHub action versions
- Match major versions of Nx and Nx Cloud
- Use `setup-node` for package caching
- Include Node.js version in cache key
- Increase Node.js max memory allocation to 6GB
- Separate docs build from deployment
- Build docs in PR runs
- Remove duplicate checkout steps
- Remove unnecessary nx-set-shas steps
- Remove unused job steps from previous setups
- Remove unused `docs` workflow
- Add GitHub Actions concurrency strategy
- Add Nx workspace lint step
- Add format check step
- Add 4th Nx Cloud DTE agent
- Optimize Nx Cloud DTE agent stop timing
- Optimize task distribution strategy for GitHub Free


.github/actions/setup/action.yml
```diff
@@ -8,31 +8,28 @@ inputs:
...
           ~/.cache/Cypress
-        key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+        key: ${{ runner.os }}-node-${{ inputs.node-version }}-yarn-${{ hashFiles('**/yarn.lock') }}
         restore-keys: |
-          ${{ runner.os }}-yarn-
+          ${{ runner.os }}-node-${{ inputs.node-version }}-
 
     - name: Install dependencies
-      if: steps.yarn-cache.outputs.cache-hit != 'true'
       shell: bash
```
hoebbelsB, line 33: I don't get why this should be better. Now the installation & post-install hooks need to happen every time.

.github/workflows/build-and-test.yml
```diff
@@ -22,20 +25,32 @@ jobs:
...
 
     steps:
-      - uses: actions/checkout@v2
-      - uses: ./.github/actions/setup
+      - name: Checkout all commits
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Derive appropriate SHAs for base and head for `nx affected` commands
+        uses: nrwl/nx-set-shas@v3
```
hoebbelsB, line 49: why can't we put those two into the setup step as well? looks like it is duplicated

.github/workflows/build-and-test.yml
```diff
@@ -50,70 +65,71 @@ jobs:
...
 
     # Deploy to the github-pages environment
     environment:
       name: github-pages
       url: www.rx-angular.io
       #url: ${{ steps.deployment.outputs.page_url }}
 
-    env:
-      NX_CLOUD_DISTRIBUTED_EXECUTION: false
-      NX_DISTRIBUTED_TASK_EXECUTION: false
```
hoebbelsB, line 75: I think this is still needed as we stop all agents in the build step, not in the build-docs step.

.github/workflows/build-and-test.yml
```diff
@@ -22,20 +25,32 @@ jobs:
...
+      - name: Derive appropriate SHAs for base and head for `nx affected` commands
+        uses: nrwl/nx-set-shas@v3
+      - name: Set up dependencies
+        uses: ./.github/actions/setup
 
-      - run: yarn nx-cloud start-ci-run
+      - run: npx nx-cloud start-ci-run
 
       - name: Run Affected lint
         run: yarn nx affected:lint --parallel --maxParallel=3
```
hoebbelsB, line 56: can we let it run with `--quiet`? I think we can also increase the `maxParallel` arg for the linting & testing steps.

image

.github/workflows/build-and-test.yml
```diff
@@ -50,70 +65,71 @@ jobs:
       - name: Run Affected build
         run: yarn nx affected:build --exclude=docs --parallel=false
 
-      - run: yarn nx-cloud stop-all-agents
+      - run: npx nx-cloud stop-all-agents
```
hoebbelsB, line 68: this will be problematic for the build-docs step. We could introduce another job which just stops all agents. It needs to wait for `build` and `build-docs` to pass

hoebbelsB (changes requested):
hey @LayZeeDK thank u very much for this PR! I have left a couple of questions and suggestions :) looking forward to your feedback

.github/actions/setup/action.yml
LayZeeDK: One issue before was that the Node.js version wasn't taken into account. This causes problems because the `docs` project uses Node.js v16 while the rest use Node.js v14. Natively compiled modules are different between different operating systems and different versions of Node.js so they shouldn't be reused across operating systems or major Node.js versions.

In general, caching the compiled output in `node_modules` rather than the global package cache isn't recommended as some tools such as NGCC write to `node_modules`, sometimes in non-deterministic ways which can cause the fields to be corrupted, go stale, or out of sync.

There are other issues with using `node_modules` as some versions of some package managers delete files in `node_modules` as part of determining dependency graphs and linking dependencies.

Reproducibility and determinism should be preferred over speed if we have to choose between the two. Our docs CI builds have been flaky, likely because we have been caching the `node_modules` directory instead of the Yarn cache directory and sharing it across Node.js versions 14 and 16.

Ideally, we would just use the `cache: yarn` parameter of `actions/setup-node`, in fact I suggest we do that to decrease maintenance but we would still have to cache the Cypress cache directory manually by using `actions/cache`.

.github/workflows/build-and-test.yml
LayZeeDK: We can't use a local action without checking out the source code first which is why we had duplicate `actions/checkout` steps. We could keep `nrwl/nx-set-shas` but it isn't required for jobs that don't use `nx affected` commands.

.github/workflows/build-and-test.yml
LayZeeDK: > I think we can also increase the maxParallel arg for the linting & testing steps.

Hosted GitHub runners only have 2 vCPUs so 2 parallel tasks would be a better fit as this workflow isn't using Nx's reusable workflows for DTE.

LayZeeDK: > can we let it run with --quiet?

It would be better to have 0 warnings than ignore them. I usually add `--max-warnings=0` when linting to have lint warnings break the build.

The reason I switched from `yarn nx-cloud` to `npx nx-cloud` is that it was hiding errors caused by a mismatched version of `@nrwl/nx-cloud`.

.github/workflows/build-and-test.yml
LayZeeDK: Until we can upgrade to Nx >=13.10 to support Node.js v16 across all projects, we have to keep the `docs` build in a job separate from the `nx affected` job for the other projects.

> We could introduce another job which just stops all agents. It needs to wait for build and build-docs to pass

That's a good suggestion 👍

LayZeeDK: Aah, now I understand this custom Nx Cloud DTE workflow. We don't need DTE to build just the `docs` project. We can still use Nx Cloud's distributed cache or build the `docs` project on Nx Cloud cache miss.

LayZeeDK: I have disabled DTE for the `build-docs` job as it can't run on the Node.js 14 agents which is the reason it has a separate job. It can still use Nx Cloud's distributed cache though.

LayZeeDK:
@hoebbelsB 
I have tweaked the CI workflow according to your comments.

I have also created an equivalent workflow that uses Nx's reusable DTE workflows in `ci-cd.yml`. Consider whether we're better off using what's open source and maintained by Nrwl.

.github/workflows/build-and-test.yml
hoebbelsB: Ofc it would be better. But finding errors in that mess is currently the bigger problem I'd say. As for me, I'd still go with quiet for now

.github/workflows/build-and-test.yml
```diff
@@ -5,115 +5,185 @@ on:
...
-
+      - name: Checkout all commits
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Set up dependencies
+        uses: ./.github/actions/setup
+
+      - name: Start Nx Cloud DTE
+        run: yarn nx-cloud start-ci-run --stop-agents-after="e2e" --agent-count=4
```
hoebbelsB, line 80: why do we need to specify --stop-agents here? Isn't it possible we run into a race condition? Let's say e2e comes from cache and finishes instantly, but some build is still missing. We anyway stop the agents manually, right?

hoebbelsB: also, do we need to specify the --agent-count here? never seen this before. can we have this as a variable at least?

.github/workflows/build-and-test.yml
```diff
@@ -5,115 +5,185 @@ on:
...
+      - name: Lint Nx workspace
+        env:
+          NX_CLOUD_DISTRIBUTED_EXECUTION: false
+          NX_DISTRIBUTED_TASK_EXECUTION: false
+        run: yarn nx workspace-lint
+      - name: Check formatting
+        env:
+          NX_CLOUD_DISTRIBUTED_EXECUTION: false
+          NX_DISTRIBUTED_TASK_EXECUTION: false
+        run: yarn nx format:check
```
hoebbelsB, line 91: why are those running outside nx cloud?

.github/workflows/build-and-test.yml
```diff
@@ -5,115 +5,185 @@ on:
...
+      # lint: 2 tasks assigned at a time, 1 task per vCPU
       - name: Run Affected lint
-        run: yarn nx affected:lint --parallel --maxParallel=3
+        run: yarn nx affected:lint --parallel --max-parallel=2 --quiet
 
+      # test: 1 task assigned at a time with 2 parallel processes, 1 process per vCPU
       - name: Run Affected test
-        run: yarn nx affected:test --parallel --maxParallel=3
+        id: test
+        run: yarn nx affected:test --parallel --max-parallel=1 --max-workers=2 --ci --code-coverage
```
hoebbelsB, line 103: I don't get why max-parallel=2 when we are running 4 agents. Also, why do we need to specify workers for test and cannot use parallel execution there?

.github/workflows/build-and-test.yml
```diff
@@ -5,115 +5,185 @@ on:
...
-
-      # don't run build in parallel, can cause dead locks
+      # build: 2 tasks assigned at a time, 1 task per vCPU
       - name: Run Affected build
-        run: yarn nx affected:build --exclude=docs --parallel=false
+        run: yarn nx affected:build --exclude=docs --parallel --max-parallel=2
+
+      # e2e: 1 tasks assigned at a time, 1 process per agent
+      - name: Run Affected e2e
+        run: yarn nx affected:e2e --parallel=false --max-parallel=1
```
hoebbelsB, line 111: isn't max-parallel=1 === parallel=false?

hoebbelsB (commented):
hi @LayZeeDK again, big thx for all your efforts! I've reviewed everything now and left some questions for you.

If you ask me, i'd go with the self-maintained version right now. It's less mental overhead for me. But I have to admit that the code of the nrwl-style pipeline is quite nice. Maybe don't throw it away, but just comment it out somehow? 

Also, i've noticed the docs build is breaking. Can we do something about it?

.github/workflows/build-and-test.yml
LayZeeDK: > why do we need to specify --stop-agents here? Isn't it possible we run into a race condition? Let's say e2e comes from cache and finishes instantly, but some build is still missing. We anyway stop the agents manually, right?

We don't need `--stop-agents-after` because we are always calling `nx-cloud stop-all-agents` in a job step. I will remove this.

LayZeeDK: > also, do we need to specify the --agent-count here? never seen this before. can we have this as a variable at least?

In the workflow generated by `@nrwl/workspace:ci-workflow`, this is added. I'm think I read about it in the Nx docs but seems that it has since been removed. It's equivalent to setting the [`NX_CLOUD_DISTRIBUTED_EXECUTION_AGENT_COUNT`](https://nx.dev/nx-cloud/reference/env-vars#nx_cloud_distributed_execution_agent_count) environment variable and can prevent timing and queuing issues.

.github/workflows/build-and-test.yml
LayZeeDK: This is [what Nrwl's `nx-cloud-main.yml` workflow does](https://github.com/nrwl/ci/blob/af5d61669406b7c8713d286692d0add4c8ca924b/.github/workflows/nx-cloud-main.yml#L295).

.github/workflows/build-and-test.yml
LayZeeDK: > I don't get why max-parallel=2 when we are running 4 agents

When using Nx Cloud DTE, `--max-parallel` specifies how many tasks of this type are (at the most) distributed to **each** agent. This is what is reflected in the comment about this line. Each default hosted GitHub runner (agent) has 2 vCPUs. so it can effectively process 2 single-process tasks at the same time.

LayZeeDK: > Also, why do we need to specify workers for test and cannot use parallel execution there?

On hosted GitHub runners with 2 vCPUs, Jest runs better when only processing one task but with `--max-workers=2`. Jest is better at distributing the load of one test project between 2 logical cores than running 2 test projects in parallel on 2 logical cores.

.github/workflows/build-and-test.yml
LayZeeDK: I'm not sure. I'm used to more recent versions of Nx where `--max-parallel=<n>` is deprecated in favor of `--parallel=<n>`.

LayZeeDK: I guess since `parallel` is set to `false` in `nx.json`, tasks are not run in parallel by default.

.github/workflows/build-and-test.yml
LayZeeDK: These commands are not running without Nx Cloud caching though, only without DTE agents. I don't think the `workspace-lint` and `format:check` commands are cacheable operations though.

PR #1448, opened 2022-10-21
progress-viewport-prio

BioPhoton:


libs/template/experimental/viewport-prio/src/lib/viewport-strategy.directive.ts
```diff
@@ -0,0 +1,155 @@
...
+  ngOnInit() {
+    const letStrategyName$ = this.strategyProvider.primaryStrategy$.pipe(
+      map(({ name }) => name)
+    );
+
+    let lastNextNotification: RxNotification<any> = undefined;
+    let lastRenderedValue: any = undefined;
+
+    this.subscription.add(
+      this.letDirective.values$.subscribe((v) => {
```
Karnaukhov-kh, line 96: Let directive is optional. On line 120 you're checking if let directive is present, shouldn't you do it here also?

libs/template/experimental/viewport-prio/src/lib/viewport-strategy.directive.ts
```diff
@@ -0,0 +1,155 @@
...
+  set viewportPrio(prio: string | Observable<string>) {
+    this.viewportPrioObservables.next(prio);
+  }
+
+  constructor(
+    private readonly el: ElementRef<HTMLElement>,
+    private strategyProvider: RxStrategyProvider,
+    @Inject(LetDirective)
+    @Optional()
+    private letDirective: LetDirective<any> | null
```
Karnaukhov-kh, line 83: Is it functional without let directive?

libs/template/experimental/viewport-prio/src/lib/viewport-strategy.directive.ts
```diff
@@ -0,0 +1,155 @@
...
+        .subscribe(([visibility, strategyName]) => {
+          // @Todo add comments
+          if (this.letDirective !== null) {
+            this.letDirective.strategy = strategyName as string;
+          }
+
+          // @Todo add comments
+          if (visibility && lastRenderedValue !== lastNextNotification.value) {
+            // render actual state on viewport enter
+            this.letDirective.templateNotification$.next(lastNextNotification);
```
Karnaukhov-kh, line 127: Same, do you need to check for letDirective presence?

Karnaukhov-kh (commented):
Hi @BioPhoton, viewport-prio will definitely be a very promising tool in our arsenal. I have couple of questions about it.

- Is it doing anything without let directive?
- Is it supposed to work *only* with let directive? (What about if,for,chunk)
- If it's an extension specifically for let directive, should we make it part of let directive instead of standalone feature?

PR #1460, opened 2022-11-08
fix(cdk): list-template-manager should always emit latest value

hoebbelsB:
## Bug

The list template manager emits an empty array after it received a value and detected no changes to render. This leads to unexpected value emissions through `rxFor`s rendercallback.

## Fix

The list template manager should always emit the latest set of data, even if no change was detected on new value input.
This is e.g. useful in scenarios where you rely on the renderCallback in order to hide a loading spinner after an asynchronous operation.
I'm thinking about a search bar triggering an http request and returning the same data as before. `rxFor` wouldn't render any new change, but the renderCallback should still indicate the change got processed.


edbzn (approved):

edbzn:
I started to investigate to find out why coverage reports are broken.

Issue #431, opened 2020-11-17
Harmonize Naming for Template RC


BioPhoton:
# Problem Solved By The Feature

As we will ship quite a lot of features we need to come up with a naming strategy. IMHO we should align as much as possible to the Angular naming as people will understand it easier.

We should consider consistent naming of:

- Type
- Interface
- Constant
- Enum
- Functions
- RxJS
  - RxJS Operator
  - RxJS Creation
- Token
- Service/Provider
- Pipe
- Directive
  - Structural Directive
  - Attribute Directive
- Module

# Solution

Angular applies no prefix to Services, Modules, Functions, etc. this is a good idea as it shortens all names.
I suggest the following naming and the related reasoning:

The items
- **Type**
- **Interface**
- **Constant**
- **Enum**
- **Functions**
should have no prefix and should just be named intuitively and distinguishable.

RxJS related things stick to the upcoming naming conventions or at least align closely.

Example:

- **RxJS**
```typescript
  mergeWith(observables)
```
RxAngular
```typescript
  coalesceWith(durationSelector, scope)
```

- **Token**
- Service/Provider
should also not be prefixed and just be intuitively and distinguishable named. 

- **Pipe**
same

- **Directive**
  - **Structural Directive**
  - **Attribute Directive**

The only prefix (`ng`), present in Angular is for directives. 
e.g. `*ngIf` or `ngClass`. This is already existing in names for the structural directive `rxLet`.

As the directives have quite a lot of features ATM and more upcoming this will be the biggest consideration.
The prefix `ng`, as mentioned, can just be replaced by `rx`. 

The tricky part here is the intuitive naming of:
- NotificationKind
- Observer callbacks
- Bindings
- Class names
- Local variables
  - Reactive-Context variables
  - Template Triggers
  - Other local variables in e.g. `rxFor`
  - Local methods in e.g. `rxFor`
- Template slots
- Child directives e.g. `rxSwitchCase`
- Local Syntax e.g. `let of` 

Example:

```typescript
// RenderAware
type NotificationKind = 'suspense' | 'next' | 'error' | 'complete';

// Let Directive - value observer -> no more derivation of Observer possible
interface RxTemplateObserver<T> {
  suspense?: () => void;
  next?: (v: T) => void;
  error?: (v: any) => void;
  complete?: () => void;
}

// Let Directive - template names
type TemplateName = 'suspense' | 'next'  | 'error' | 'complete';

// TemplateManager - view context
[...]

PR #1473, opened 2022-11-24
refactor(template): harmonize LetDirective triggers

edbzn:
Harmonize LetDirective inputs as described in https://github.com/rx-angular/rx-angular/issues/431.


hoebbelsB (approved):

hoebbelsB:
just as a note, shouldn't the title not be: `refactor(template)` instead of `fix(template)` ? :D

PR #1474, opened 2022-11-24
refactor(template): remove root barrel exports

edbzn:


libs/template/src/index.ts
```diff
@@ -1,14 +1 @@
...
-} from '@rx-angular/template/unpatch';
-
-// STABLE
-// TODO(artur): we're keeping it here until we get a migration script.
-export { LetModule, LetDirective } from '@rx-angular/template/let';
-
-export { PushModule, PushPipe } from '@rx-angular/template/push';
-
-export { ForModule, RxFor, RxForViewContext } from '@rx-angular/template/for';
+export {};
```
hoebbelsB, line 1: why do we have to keep that file anyway?

edbzn: Because an entry file is required for ng-packagr in order to build, if the file is empty then I get the following error:

```
Internal error: failed to get symbol for entrypoint
```

hoebbelsB:
@edbzn didn't we implement a migration to handle this already? Should we re-enable it for this change?

PR #1475, opened 2022-11-24
refactor(template): harmonize RxIf triggers

edbzn:
Harmonize RxIf inputs as described in https://github.com/rx-angular/rx-angular/issues/431.


hoebbelsB:
can we do it for let.directive as well?

hoebbelsB (approved):

PR #1477, opened 2022-11-24
chore: fix template build warnings

edbzn:


tsconfig.base.json
```diff
@@ -5,7 +5,7 @@
     "sourceMap": true,
     "declaration": false,
     "moduleResolution": "node",
-    "emitDecoratorMetadata": true,
```
hoebbelsB, line 8: isn't emitDecoratorMetadata: true angular default setting? Why do we turn it off?

edbzn: I removed this change, the correct way to fix these warnings 

```
WARNING: "ChangeDetectorRef", "NgZone", "TemplateRef" and "ViewContainerRef" are imported from external module "@angular/core" but never used in "dist/libs/template/esm2015/if/lib/if.directive.js" and "dist/libs/template/esm2015/if/lib/if.module.js".
```

is to upgrade to ng-packagr.

hoebbelsB: I don't get it. Do we still have the warning now? Can we remove it by upgrading ng-packagr? If so, why didn't u? :D

edbzn: @hoebbelsB This PR fixes multiple warnings, but not all, some of them are related to ng-packagr, and we have to update the whole @nrwl/workspace to update ng-packagr, so I'll make another PR for the workspace. Last time I tried I had different issues, so I'll need to work on it further. This PR can be merged as it.

PR #1487, opened 2022-12-23
fix(docs): add wrapper to home table to fix overflow

ChristopherPHolder:
The table on the landing page of the docs has a table that is rendered at larger with a mobile device. 

This causes an overflow-style issue:

Screenshot 2022-12-23 at 11 35 19

It was fixed by wrapping it with a div tag and applying an overflow scroll.


edbzn:
I think we can remove the compatibility table from root readme, it's not the first information I need to see and the table is already in every project readme.

ChristopherPHolder:
> I think we can remove the compatibility table from the landing page and root readme, it's not the first information we need to see and the table is already in every project. Wdyt?

I agree that is should not be in the landing page, but I don't know what is the write place to have it. 

Do you have any suggestions on where is should go? 

However, this is also a topic for another issue. 

This PR is just a quick UI fix.

PR #1497, opened 2023-01-11
docs(state): remove coverage status badge

mikelgo:
I guess this badge is by accident still there (in all other packages this has been removed)


edbzn:
Note that I configured codecov in the repository, it's accessible [here](https://app.codecov.io/gh/rx-angular/rx-angular), we can get coverage badges from it.

PR #1509, opened 2023-01-30
feat: upgrade Angular to v13

edbzn:
Upgrade Angular to support v13.


package.json
```diff
@@ -34,7 +34,7 @@
   },
   "lint-staged": {
     "**/!(images)/**/*.{ts,js,html,scss,md,mdx,json}": [
-      "yarn affected:lint -- --uncommitted --fix --parallel"
+      "nx affected:lint --uncommitted --fix --parallel"
```
hoebbelsB, line 37: won't this mean it will always use the users globally installed version of `nx` instead of the one living in node_modules?

hoebbelsB:
```suggestion
      "yarn affected:lint --uncommitted --fix --parallel 4"
```

Also, I have the feeling `affected:lint --uncommitted` is doing way way more than needed. We should probably rely only on eslint, without affected. But that's a topic for another PR.

hoebbelsB: I assume everyone has 4 threads available btw ^^

apps/ssr/src/test-setup.ts
```diff
@@ -1 +1,14 @@
...
+  BrowserDynamicTestingModule,
+  platformBrowserDynamicTesting,
+} from '@angular/platform-browser-dynamic/testing';
+
+getTestBed().resetTestEnvironment();
+getTestBed().initTestEnvironment(
+  BrowserDynamicTestingModule,
+  platformBrowserDynamicTesting(),
+  { teardown: { destroyAfterEach: false } }
+);
```
LayZeeDK, line 14: suggestion: I know this is the default Nx migration but it would be better if our unit tests could run with the default TestBed setup where `destroyAfterEach` is set to `true` by removing all the added lines
```suggestion
```

apps/demos/src/test-setup.ts
```diff
@@ -1 +1,14 @@
...
+  BrowserDynamicTestingModule,
+  platformBrowserDynamicTesting,
+} from '@angular/platform-browser-dynamic/testing';
+
+getTestBed().resetTestEnvironment();
+getTestBed().initTestEnvironment(
+  BrowserDynamicTestingModule,
+  platformBrowserDynamicTesting(),
+  { teardown: { destroyAfterEach: false } }
+);
```
LayZeeDK, line 14: suggestion: I know this is the default Nx migration but it would be better if our unit tests could run with the default TestBed setup where `destroyAfterEach` is set to `true` by removing all the added lines
```suggestion
```

apps/tour-of-heroes-ngxs/src/test-setup.ts
```diff
@@ -1 +1,14 @@
...
+  BrowserDynamicTestingModule,
+  platformBrowserDynamicTesting,
+} from '@angular/platform-browser-dynamic/testing';
+
+getTestBed().resetTestEnvironment();
+getTestBed().initTestEnvironment(
+  BrowserDynamicTestingModule,
+  platformBrowserDynamicTesting(),
+  { teardown: { destroyAfterEach: false } }
+);
```
LayZeeDK, line 14: suggestion: I know this is the default Nx migration but it would be better if our unit tests could run with the default TestBed setup where `destroyAfterEach` is set to `true` by removing all the added lines
```suggestion
```

apps/vanilla-case-studies/src/test-setup.ts
```diff
@@ -1 +1,14 @@
...
+  BrowserDynamicTestingModule,
+  platformBrowserDynamicTesting,
+} from '@angular/platform-browser-dynamic/testing';
+
+getTestBed().resetTestEnvironment();
+getTestBed().initTestEnvironment(
+  BrowserDynamicTestingModule,
+  platformBrowserDynamicTesting(),
+  { teardown: { destroyAfterEach: false } }
+);
```
LayZeeDK, line 14: suggestion: I know this is the default Nx migration but it would be better if our unit tests could run with the default TestBed setup where `destroyAfterEach` is set to `true` by removing all the added lines
```suggestion
```

libs/cdk/package.json
```diff
@@ -48,7 +48,7 @@
     "url": "https://github.com/rx-angular/rx-angular.git"
   },
   "peerDependencies": {
-    "@angular/core": ">=12.0.0",
+    "@angular/core": ">=13.0.0",
```
LayZeeDK, line 51: issue: Only mark Angular 13.x as officially supported
```suggestion
    "@angular/core": "^13.0.0",
```

libs/cdk/src/test-setup.ts
```diff
@@ -1,2 +1,17 @@
...
+  BrowserDynamicTestingModule,
+  platformBrowserDynamicTesting,
+} from '@angular/platform-browser-dynamic/testing';
+
+getTestBed().resetTestEnvironment();
+getTestBed().initTestEnvironment(
+  BrowserDynamicTestingModule,
+  platformBrowserDynamicTesting(),
+  { teardown: { destroyAfterEach: false } }
+);
```
LayZeeDK, line 17: suggestion: I know this is the default Nx migration but it would be better if our unit tests could run with the default TestBed setup where `destroyAfterEach` is set to `true` by removing all the added lines
```suggestion
```

libs/cdk/template/spec/template-manager.spec.ts
```diff
@@ -1,4 +1,3 @@
-import 'jest-preset-angular/setup-jest'; // TODO: move this into test-setup when zone-config.spec is in its own lib
```
LayZeeDK, line 1: praise: Good catch!

libs/eslint-plugin/project.json
```diff
@@ -14,12 +13,12 @@
...
       "outputs": ["coverage/libs/eslint-plugin"],
       "options": {
-        "jestConfig": "libs/eslint-plugin/jest.config.js",
+        "jestConfig": "libs/eslint-plugin/jest.config.ts",
         "passWithNoTests": true
       }
     },
     "build": {
-      "executor": "@nrwl/node:package",
+      "executor": "@nrwl/js:tsc",
```
LayZeeDK, line 21: question: Is this executor recommended for Node.js packages?

libs/state/package.json
```diff
@@ -48,8 +48,8 @@
     "url": "https://github.com/rx-angular/rx-angular.git"
   },
   "peerDependencies": {
-    "@angular/core": ">=12.0.0",
-    "@angular/platform-browser": ">=12.0.0",
+    "@angular/core": ">=13.0.0",
+    "@angular/platform-browser": ">=13.0.0",
```
LayZeeDK, line 52: issue: Only mark Angular 13.x as officially supported
```suggestion
    "@angular/core": "^13.0.0",
    "@angular/platform-browser": "^13.0.0",
```

libs/state/test-setup.ts
```diff
@@ -2,3 +2,16 @@ import 'jest-preset-angular/setup-jest';
...
+  BrowserDynamicTestingModule,
+  platformBrowserDynamicTesting,
+} from '@angular/platform-browser-dynamic/testing';
+
+getTestBed().resetTestEnvironment();
+getTestBed().initTestEnvironment(
+  BrowserDynamicTestingModule,
+  platformBrowserDynamicTesting(),
+  { teardown: { destroyAfterEach: false } }
+);
```
LayZeeDK, line 17: suggestion: I know this is the default Nx migration but it would be better if our unit tests could run with the default TestBed setup where `destroyAfterEach` is set to `true` by removing all the added lines
```suggestion
```

libs/template/for/ng-package.json
```diff
@@ -2,11 +2,6 @@
   "$schema": "../../../node_modules/ng-packagr/package.schema.json",
   "lib": {
     "entryFile": "src/index.ts",
-    "flatModuleFile": "template-for",
-    "umdModuleIds": {
-      "@rx-angular/cdk/coercing": "rx-angular.cdk.coercing",
-      "@rx-angular/cdk/template": "rx-angular.cdk.template",
-      "@rx-angular/cdk/render-strategies": "rx-angular.cdk.render-strategies"
-    }
```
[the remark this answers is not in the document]
LayZeeDK: They were used by older versions of the Angular Package Format (APF) which included UMD bundles.

libs/template/package.json
```diff
@@ -49,9 +49,9 @@
     "url": "https://github.com/rx-angular/rx-angular.git"
   },
   "peerDependencies": {
-    "@angular/core": ">=12.0.0",
-    "@angular/platform-browser": ">=12.0.0",
-    "@rx-angular/cdk": ">=1.0.0",
+    "@angular/core": ">=13.0.0",
+    "@angular/platform-browser": ">=13.0.0",
```
LayZeeDK, line 53: issue: Only mark Angular 13.x as officially supported
```suggestion
    "@angular/core": "^13.0.0",
    "@angular/platform-browser": "^13.0.0",
```

libs/template/src/test-setup.ts
```diff
@@ -2,3 +2,16 @@ import 'jest-preset-angular/setup-jest';
...
+  BrowserDynamicTestingModule,
+  platformBrowserDynamicTesting,
+} from '@angular/platform-browser-dynamic/testing';
+
+getTestBed().resetTestEnvironment();
+getTestBed().initTestEnvironment(
+  BrowserDynamicTestingModule,
+  platformBrowserDynamicTesting(),
+  { teardown: { destroyAfterEach: false } }
+);
```
LayZeeDK, line 17: suggestion: I know this is the default Nx migration but it would be better if our unit tests could run with the default TestBed setup where `destroyAfterEach` is set to `true` by removing all the added lines
```suggestion
```

nx.json
```diff
@@ -41,7 +41,7 @@
         "accessToken": "OTg2OGFkNmMtNzA5Zi00MjBiLWFhMmQtOGYwNTQ1MjM1ZjQ3fHJlYWQtd3JpdGU=",
         "canTrackAnalytics": false,
         "showUsageWarnings": true,
-        "parallel": false
+        "parallel": 1
```
LayZeeDK, line 44: question: Do we specifically need serial task running everywhere, including locally?

suggestion: Remove this setting to use the default parallelization (3 tasks in parallel)
```suggestion
```
However, we might want or need to change our CI jobs to do this.

LayZeeDK (changes requested):

libs/cdk/package.json
hoebbelsB: It's not really telling the truth. You can easily use this version with any version above 13. But on the other hand, please do what feels right. I just want to get this thing in!! :D

nx.json
hoebbelsB: agree! I guess you already made sure the CI is precisely specifying everything, right?

libs/template/package.json
edbzn: IMO we should manage compat range this way: 

RxAngular 2.0.0 :arrow_right: @angular/core >=13.0.0
RxAngular 14.0.0 :arrow_right: @angular/core ^14.0.0

nx.json
edbzn: :+1: but let's do this in another PR

aaa123eee: additional issue created https://github.com/rx-angular/rx-angular/issues/1527

libs/cdk/package.json
aaa123eee: as far as angular repo itself uses this format we can go in the same way

https://github.com/angular/angular/blob/main/packages/router/package.json#L30

`json
"peerDependencies": {
  "rxjs": "^6.5.3 || ^7.4.0"
}
`

libs/state/package.json
```diff
@@ -48,9 +48,9 @@
     "url": "https://github.com/rx-angular/rx-angular.git"
   },
   "peerDependencies": {
-    "@angular/core": ">=13.0.0",
-    "@angular/platform-browser": ">=13.0.0",
```
hoebbelsB: btw. why do we have platform-browser as a dependency at all?

libs/eslint-plugin/project.json
aaa123eee: It looks like yes, this is the default executor after generating the`@nrwl/node:library`

edbzn:
Don't merge, I'll rebase the branch because we need to change the compatibility range in the commit message with the breaking change to reflect `^13.0.0` instead of `>=13.0.0`.

PR #1512, opened 2023-02-08
Update effects.mdx

radvil:
Should make it clear so newbie like me can understand whether to just define one or both of these


Karnaukhov-kh:
Hi @radvil. Thank you for these contributions :)

PR #1514, opened 2023-02-08
docs: improve rx-state landing

Karnaukhov-kh:
- Normalized structure
- API overview
- Recommended addons
- Shorter snippets and additional materials list

### DIFF
--- a/libs/state/README.md
+++ b/libs/state/README.md
@@ -68,8 +68,8 @@ export class RxStateInheritanceClass extends RxState<{ foo: string }> {
 
 ## API overview
 
-With @rx-angular/state, you can easily manage your component state with a range of powerful methods.
-Detailed API documentation available [here](https://rx-angular.io/docs/state/api).
+With `@rx-angular/state`, you can easily manage your component state with a range of powerful methods.
+You find a detailed API documentation [here](https://rx-angular.io/docs/state/api).
 
 #### [.connect()](https://www.rx-angular.io/docs/state/api/rx-state#connect)
 
@@ -135,7 +135,7 @@ Optimize state selections and data transfer, ensure only the necessary data is t
 
 | Angular                | RxJS                 | @rx-angular/state |
 | ---------------------- | -------------------- | ----------------- |
-| `14`                   | `^7.4.0`             | `> 1.4.6`         |
+| `14+`                  | `^7.4.0`             | `> 1.4.6`         |
 | `^12.0.0` or `^13.0.0` | `^6.5.5` or `^7.4.0` | `> 1.4.6`         |
 | `^11.0.0`              | `^6.5.5`             | `<= 1.4.6`        |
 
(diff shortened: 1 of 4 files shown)


libs/state/README.md
```diff
@@ -1,79 +1,70 @@
...
-## Intro Video
+### Introduction Video
 
 [![intro-video_rx-angular--state-rx-state](https://user-images.githubusercontent.com/10064416/147395467-876ec499-645f-4f84-bde9-9bffaac22c62.PNG)](https://www.youtube.com/watch?v=CcQYj4V2IKw)
 
-## Description
+### API overview
+
+With @rx-angular/state, you can easily manage your component state with a range of powerful methods.
+Detailed API documentation available [here](https://rx-angular.io/docs/state/api).
```
hoebbelsB, line 24:
```suggestion
With `@rx-angular/state`, you can easily manage your component state with a range of powerful methods.
You find a detailed API documentation [here](https://rx-angular.io/docs/state/api).
```

libs/state/README.md
```diff
@@ -1,79 +1,70 @@
...
-- Slim and intuitive API
-- Automated subscription handling
-- Intuitive way for handling ViewModels
-- Connect any Observable source to the state
-- Partial state updates
-- Reactive state selection
-- Lazy state (no BehaviorSubject)
-- Foundation for zone-less Angular applications
+Customize your state accumulation function with `.setAccumulator()`.
+Whether you want to update the accumulation logic for deep updates or solve immutability problems, this method is a powerful tool for fine-tuning your state management.
```
hoebbelsB, line 56: Two things about this part. Why did u create custom texts for the landing page instead of just re-using what is stated right now in the docs as description text for each api?

I also do not find it so useful without any usage information at this stage. I anyway have to click on the link to see how it is used.

libs/state/README.md
```diff
@@ -83,59 +74,62 @@ nx migrate @rx-angular/state
...
+```ts
+export class RxStateInheritanceClass extends RxState<{ foo: string }> {
+  value$ = this.select();
 
-## Tutorials
+  constructor() {
+    super();
+  }
+}
+```
```
hoebbelsB, line 101: The API section maybe does make more sense if you move the `Usage` part on top of it. I think it should be the very first thing after the description.

libs/state/README.md
```diff
@@ -145,6 +139,28 @@ nx migrate @rx-angular/state
 | `^12.0.0` or `^13.0.0` | `^6.5.5` or `^7.4.0` | `> 1.4.6`         |
 | `^11.0.0`              | `^6.5.5`             | `<= 1.4.6`        |
```
hoebbelsB, line 140: `14` is not true, it's `14` & `15`

hoebbelsB (changes requested):
hey @Karnaukhov-kh big THX for this PR!! I have reviewed it and added some suggestions and questions for you. Happy to hear about feedback

libs/state/README.md
Karnaukhov-kh: Agree

hoebbelsB (approved):
LGTM, ty!

PR #1519, opened 2023-02-24
Upgrade to angular v14

hoebbelsB:
* Upgraded the workspace to latest nx & ng 14 lts
* Fixed broken typings due to typescript version update
* updated peerDependencies

Needs to wait for #1509 has been landed and released


libs/cdk/coercing/spec/coerceDistinctObservableWith.spec.ts
```diff
@@ -320,18 +344,23 @@ describe('coerceDistinctWith', () => {
...
-          c:               '-----',
-          d:               '-----',
+          b: '-----',
+          c: '-----',
+          d: '-----',
         };
 
         const values = {
-          a: 'hello dear contributor',
+          a: cold('(a)', { a: 'hello dear contributor' }),
```
hoebbelsB, line 356: @BioPhoton this test was failing for a good reason and I don't understand why it passed before this upgrade.

`coerceDistinctWith` internally uses `coerceObservableWith` which will turn any non-observable value into `of`.

`of` completes immediately, so `exhaustAll()` emits values.

Do we want to change that behavior? For now, i've adjusted the spec

aaa123eee:
Is the output the same for you? I've compared v14 library build output with v13 and it is missing secondary entries

v13 (package.json exists)
Screenshot 2023-02-27 at 16 30 39

v14 (no package.json)
Screenshot 2023-02-27 at 16 18 15

libs/state/package.json
```diff
@@ -48,7 +48,7 @@
     "url": "https://github.com/rx-angular/rx-angular.git"
   },
   "peerDependencies": {
-    "@angular/core": ">=13.0.0",
+    "@angular/core": ">=14.0.0",
```
edbzn, line 51:
```suggestion
    "@angular/core": "^14.0.0",
```

libs/template/package.json
```diff
@@ -49,8 +49,8 @@
     "url": "https://github.com/rx-angular/rx-angular.git"
   },
   "peerDependencies": {
-    "@angular/core": ">=13.0.0",
-    "@rx-angular/cdk": "^2.0.0",
+    "@angular/core": ">=14.0.0",
```
edbzn, line 52:
```suggestion
    "@angular/core": "^14.0.0",
```

edbzn (commented):
Breaking change commits should contain a body message with `^14.0.0` instead of `>= 14`.

libs/cdk/coercing/spec/coerceDistinctObservableWith.spec.ts
edbzn: Assuming we keep the spec adjusted and move forward.

PR #1523, opened 2023-02-26
Draft: signal prototypes

hoebbelsB:
# Description

https://hackmd.io/gf5D1kC0RVKfDrJZXY-V0Q

## State

* [x] connect accepts signals
* [x] get computed from state
* [x] read signal from state
* [x] create effect from state

```ts
// component state
private state = inject<RxState<{ todos: Todo[]; foo: string}>>(RxState);
```

**Connect a signal**

```ts
const todos: Signal<Todo[]> = signal(todoData);
this.state.connect('todos', todos);
```

**Computed from state**

```ts
// takes a function that gets each property of the state as `Signal<T[K]>`.
// in this case, todos will be `Signal<Todo[]>`

const count: Signal<number> = this.state.computed(({ todos }) => {
    return todos().length;
});
```

**Signal from state**

```ts
const todosSignal: Signal<Todo[]> = this.state.signal('todos');
```

**Effect from state**

```ts
const _effect: Effect = this.state.effect(({ todos }) => {
   if (todos().length > 20) {
      console.log('something was added');
    }
});
```

### To be discussed

* [ ] proxy object for easier access

## Template

* [ ] rxLet accepts signal
* [ ] rxFor accepts signal
* [ ] rxIf accepts signal
* [ ] push pipe accepts signal

### To be discussed

* [ ] introduce `transition` like API (as in react concurrent mode) to have a scheduled signal using `RxRenderStrategies` internally?

### DIFF
--- a/apps/demos/src/app/features/state/signal-state/signal-let.directive.ts
+++ b/apps/demos/src/app/features/state/signal-state/signal-let.directive.ts
@@ -1,0 +1,54 @@
+import {
+  Directive,
+  effect,
+  EmbeddedViewRef,
+  Input,
+  OnDestroy,
+  OnInit,
+  Signal,
+  TemplateRef,
+  ViewContainerRef,
+} from '@angular/core';
+import { RxStrategyProvider } from '@rx-angular/cdk/render-strategies';
+import { Subject } from 'rxjs';
+
+@Directive({ selector: '[signalLet]', standalone: true })
+export class SignalLetDirective<T> implements OnInit, OnDestroy {
+  static ngTemplateGuard_rxLet: 'binding';
+
+  @Input() signalLet: Signal<T>;
+  constructor(
+    private viewContainerRef: ViewContainerRef,
+    private template: TemplateRef<{ $implicit: T }>,
+    private strategyProvider: RxStrategyProvider
+  ) {}
+
+  ngOnInit() {
+    let embeddedView: EmbeddedViewRef<{ $implicit: T }>;
+    const scheduleCd$ = new Subject<T>();
+    effect(() => {
+      const value = this.signalLet();
+      scheduleCd$.next(value);
+    });
+    scheduleCd$
+      .pipe(
+        this.strategyProvider.scheduleWith((value) => {
+          if (!embeddedView) {
+            embeddedView = this.viewContainerRef.createEmbeddedView(
+              this.template,
+              { $implicit: value }
+            );
+            embeddedView.detach();
+          } else {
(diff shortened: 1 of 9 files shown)


apps/demos/src/app/features/state/signal-state/signal-let.directive.ts
```diff
@@ -0,0 +1,54 @@
...
+  OnDestroy,
+  OnInit,
+  Signal,
+  TemplateRef,
+  ViewContainerRef,
+} from '@angular/core';
+import { RxStrategyProvider } from '@rx-angular/cdk/render-strategies';
+import { Subject } from 'rxjs';
+
+@Directive({ selector: '[signalLet]', standalone: true })
```
eneajaho, line 15: At this point, I'd just call it let directive as signals are going to be the recommended approach going forward

libs/template/let/src/lib/let.directive.ts
```diff
@@ -616,3 +630,14 @@ export class LetDirective<U> implements OnInit, OnDestroy, OnChanges {
     this.templateManager.nextStrategy(this.strategyHandler.values$);
   }
 }
+
+function toObservable<T>(input: Signal<T>): Observable<T> {
```
eneajaho, line 634: Do we want to remove this in favor of Angular's toObservable?

libs/state/src/lib/rx-state.service.ts
```diff
@@ -634,6 +684,62 @@ export class RxState<T extends object> implements OnDestroy, Subscribable<T> {
...
+  private createSignalStoreProxy() {
+    const state$ = this.$;
+    const subs = this.subscription;
+    const currentState = (): T => this.get() || ({} as T);
+    const signalStore: SignalStore<T> = {} as SignalStore<T>;
+    this.signalStoreProxy = new Proxy<SignalStore<T>>(signalStore, {
+      get<K extends keyof T>(
+        target: SignalStore<T>,
+        p: K | string | symbol
+      ): Signal<T[K]> {
```
eneajaho, line 700: Here we say `Signal<T[K]>` but in the `initialValue` we have `signal(val ?? null)`

Do we want to have it as `Signal<T[K] | null>` ?

apps/demos/src/app/features/state/signal-state/signal-let.directive.ts
BioPhoton: I would reconsider the ideas in the template in general. The recent releases change a lot and we should adopt to the "modern way".

PR #1530, opened 2023-03-07
chore: move transformations docs

aaa123eee:
- mv transformations docs from `docs/cdk/api/transformation-helpers` to `docs/cdk/transformations`

To test locally run `nx run docs:serve` and open `http://localhost:3000/docs/cdk/transformations`


apps/docs/docs/cdk/transformations/transformations.mdx
```diff
@@ -8,19 +8,46 @@ hide_title: true
...
+
+**Helpers that converting one structure to another**
+
+```
+Array: dictionaryToArray
+Object: toDictionary
+```
+
+If the original structure not matching the expected type empty target structure will be returned.
+If the original structure is null or undefined helper will return null or undefined.
```
edbzn, line 50: What about adding more general information, but not implementation details (like what it returns) here:

- Manage objects and collections with an immutable guarantee
- Reduce boilerplate to perform state changes
- Type-safe and tree-shakeable functions
- Ensure serializable objects and arrays

What do you think?

apps/docs/docs/cdk/transformations/transformations.mdx
```diff
@@ -8,19 +8,46 @@ hide_title: true
 # @rx-angular/cdk/transformations
 
 [![npm](https://img.shields.io/npm/v/%40rx-angular%2Fcdk.svg)](https://www.npmjs.com/package/%40rx-angular%2Fcdk)
-![rx-angular CI](https://github.com/rx-angular/rx-angular/workflows/rx-angular%20CI/badge.svg?branch=master)
+![rx-angular CI](https://github.com/rx-angular/rx-angular/workflows/rx-angular%20CI/badge.svg?branch=main)
 
 > A set of functions to manipulate the state in an efficient way.
 
-`@rx-angular/cdk/transformations` provides a set of functions to perform state transformations in a performant way.
+- `@rx-angular/cdk/transformations` provides a set of tools that simplifies the management of data structures in an immutable way.
```
edbzn, line 15:
```suggestion
- `@rx-angular/cdk/transformations` provides a set of tools that simplifies the management of data structures in a performant and immutable way.
```

apps/docs/docs/cdk/transformations/transformations.mdx
aaa123eee: nice, I can put it as `Key features`

like here https://www.rx-angular.io/docs/cdk/coalescing#key-features

aaa123eee: added `Key features` list instead of `Principles`

edbzn (approved):

PR #1537, opened 2023-04-12
Fix angular peerDependency

hoebbelsB:
Tried to install the latest rx-angular version in the angular movies app. It failed horribly as our peerDependency to angular/core disallows versions beyond v14. This makes no sense to me, as the build is compatible with newer versions.

Had to to `npm install --force` in order to test the library.

image


LayZeeDK:
As you know, we haven't run tests with Angular 15 so we don't officially support it with the current version. The solution is to use a package manager install flag such as --legacy-peer-deps or wait for the next major release of RxAngular.

PR #1542, opened 2023-04-19
test(template): implement component-tests for experimental virtual-scrolling

hoebbelsB:
Implemented test cases for all provided scroll strategies covering rendering, scroll behavior, data mutations, error handling. The final step is to integrate it into the pipeline and get a review on the cypress tests implementation and configuration.

image

image

image


hoebbelsB:
> Looks good overall, just a detail I think async/await is easier to read than Promise.then(), especially when it's combined with some cy.then().

`cy.then` does not return a promise and having async `it` blocks caused issues I couldn't resolve. This is why I went that way. If there is an easier solution, I'm happy to hear about it :)

edbzn:
I meant using async await for mounting the component, it's a minor detail, just less nested callback code.

PR #1543, opened 2023-04-19
Add inject functions

edbzn:
### `rxActions` function

```ts
@Component({})
class TestComponent {
  readonly actions = rxActions<{ searchStringChanged: Event }>(
    // to have it unified between template (event.target.value) and the class (just a string) we can add a transform
    withTransforms({ searchStringChanged: (e) => e.target.value })
  );
}
```

Composition functions:

- `withTransforms`

### `rxState` function

```ts
const interval$ = time(0, 1000);
const sideEffectFn = (state) => console.log(state);

@Component({})
class TestComponent {
  readonly state = rxState(
     withInitialState({ count: 10 }),
     witHold(interval$, sideEffectFn),
  );
}

@Component({})
class TestComponent {
  private readonly service = inject(Service);
  private readonly actions = inject(RxActions);

  readonly state = rxState(
    withConnect(() => {
      return {
        prop: this.service.prop$,
        prop2: this.actions.someAction$,
      }
    })
  );
}
```

See [this comment](https://github.com/rx-angular/rx-angular/pull/1543#issuecomment-1532434831) for other syntax support.

### As an Injection Token

```ts
const CounterStore = provideRxState<{ count: 0 }>({
  features: [withInitialState({ count: 0 })],
});

@Component({
  template: `{{ state.get().count }}`,
})
class StatefulComponent {
  readonly state = inject(CounterStore);
}
```

Composition functions:

- `withHold`
- `withConnect`
- `withInitialState`
- `withAccumulator`

## Work in progress

- rxEffects support
- align standalone rxEffects and rxState withHold
- inline doc and web doc


libs/state/actions/src/lib/inject.ts
```diff
@@ -0,0 +1,22 @@
...
+  Actions extends object
+>(): RxActionFactory<Actions> {
+  const errorHandler = inject(ErrorHandler, { optional: true }) ?? undefined;
+  const rxActionFactory = new RxActionFactory<Actions>(errorHandler);
+  /**
+   * @todo: Use DestroyRef instead when upgrading to Angular 16
+   */
+  const viewRef = inject(ChangeDetectorRef) as ViewRef;
+
+  viewRef.onDestroy(() => rxActionFactory.ngOnDestroy());
```
hoebbelsB, line 19: When used in directives, this most probably leads to unexpected behaviors as this instance will be tied to the onDestroy hook of the component which hosts the directive.

libs/state/src/lib/inject.ts
```diff
@@ -0,0 +1,57 @@
...
+}
+
+export function rxState<State extends object>(
+  ...features: RxStateFeatures<State>
+): RxState<State> {
+  const rxState = new RxState<State>();
+  /**
+   * @todo: Use DestroyRef instead when upgrading to Angular 16
+   */
+  const viewRef = inject(ChangeDetectorRef) as ViewRef;
```
hoebbelsB, line 50: When used in directives, this most probably leads to unexpected behaviors as this instance will be tied to the onDestroy hook of the component which hosts the directive.

libs/state/actions/src/lib/inject.ts
```diff
@@ -0,0 +1,22 @@
...
+import { RxActionFactory } from './actions.factory';
+
+export function injectRxActionFactory<
+  Actions extends object
+>(): RxActionFactory<Actions> {
+  const errorHandler = inject(ErrorHandler, { optional: true }) ?? undefined;
+  const rxActionFactory = new RxActionFactory<Actions>(errorHandler);
+  /**
+   * @todo: Use DestroyRef instead when upgrading to Angular 16
+   */
```
hoebbelsB, line 16: imo we should not introduce this with v15, because it will be hard to communicate about the component level destroy hook.

libs/state/actions/src/lib/inject.ts
```diff
@@ -0,0 +1,22 @@
+import {
+  ChangeDetectorRef,
+  ErrorHandler,
+  ViewRef,
+  inject,
+} from '@angular/core';
+import { RxActionFactory } from './actions.factory';
+
+export function injectRxActionFactory<
```
hoebbelsB, line 9: should it be named `rxActionFactory`, or even simpler `rxActions` ? Why do we even have to call the create function as a user?

libs/state/effects/src/lib/inject.ts
```diff
@@ -0,0 +1,20 @@
+import {
+  ChangeDetectorRef,
+  ErrorHandler,
+  ViewRef,
+  inject,
+} from '@angular/core';
+import { RxEffects } from './effects.service';
+
+export function injectRxEffects(): RxEffects {
```
hoebbelsB, line 9:
```suggestion
export function rxEffects(): RxEffects {
```

libs/state/src/lib/inject.ts
```diff
@@ -0,0 +1,57 @@
...
+export type RxStateFeatures<State extends object> = RxStateFeature<State>[];
+
+/**
+ * @todo: see how to infer generic type from accumulator to accumulatorFn.
+ */
+export function accumulator<State extends object>(
+  accumulatorFn: AccumulationFn
+) {
+  return (rxState: RxState<State>) => rxState.setAccumulator(accumulatorFn);
+}
```
hoebbelsB, line 19: as far as I enjoy the shortness of the API, i'm concerned we will run into problems when we call the feature factories like this. Maybe we should stick to the angular pattern and call it `withAccumulator` instead? Same goes for `initialState` and others.

Maybe before even releasing those functions we should go for a public RFC?

package.json
```diff
@@ -136,6 +136,7 @@
     "size-limit": "^6.0.4",
     "ts-jest": "28.0.8",
     "ts-node": "10.9.1",
+    "tsd": "^0.28.1",
```
hoebbelsB, line 139: what is that?

hoebbelsB (commented):
yo @edbzn, what a nice PR this is :pray: thank u very much! I've reviewed the code and left some comments and suggestions. Happy to hear your feedback :)

Also, do you think we have a chance to finally find a way to make the `queueScheduler` configurable in `RxState` / `createAccumulationObservable`?

https://github.com/rx-angular/rx-angular/blob/main/libs/state/selections/src/lib/accumulation-observable.ts

image

libs/state/actions/src/lib/inject.ts
edbzn: We would need to provide a `withTransforms` function in order to pass input transforms to `actions.create(transforms)` but yes it would simplify the boilerplate and align with composition functions in `rxState`.

libs/state/src/lib/inject.ts
edbzn: I agree, it's nice to have a prefix for those composition functions in order to reduce the confusion with other APIs.

Not sure about the need for an RFC, I'm afraid that we just lose time and effort but do not gather the feedback we need in the end. What about discussing it publicly here and avoiding side discussions?

Marking the APIs as experimental would allow us to test them, get feedback and eventually break them if we want to.

package.json
edbzn: I introduced this package in order to test types, see this: 

```ts
import { expectType } from 'tsd';

it('should infer state from initial state', () => {
  const { component } = setupStatefulComponent(initialState({ count: 0 }));
  expectType<RxState<{ count: number }>>(component.state);
});

it('should infer state from generic type', () => {
  const { component } = setupStatefulComponent<{ count: number }>();
  expectType<RxState<{ count: number }>>(component.state);
});
```

I think we can benefit from it in other places.

libs/state/src/lib/inject.ts
hoebbelsB: how can we mark single APIs as experimental? You want to introduce a new subpackage for it?

hoebbelsB: agree, let's discuss it here

edbzn: > how can we mark single APIs as experimental? You want to introduce a new subpackage for it?

Like Angular, marking them with a symbol and/or using JSDocs. We can also introduce a subpackage but I think the first way makes more sense for this case.

hoebbelsB: v16 releases this week, we can just go for `DestroyRef` :)

hoebbelsB:
@edbzn I thought a bit about the API and imo we could improve it a bit. It looks nice in your example, but in reality you will face more than only 1 `hold` or `connect` scenario.

Right now, adding multiple connects looks like:

```ts

@Component({})
class TestComponent {
  readonly state = rxState(
     withConnect(inject(DataService).data$, 'data'),
     withConnect(inject(DataService).otherData$, 'otherData'),
 // more to come...
  );
}
```

What about an API that looks like:

```ts
@Component({})
class TestComponent {
  readonly state = rxState(
    withConnect(() => {
      const service = inject(Service);
      const actions = inject(RxActions);
      return {
        prop: service.prop$,
        prop2: actions.someAction$,
      }
    })
  );
}

// or alternative way

@Component({})
class TestComponent {
  readonly state = rxState(
    withConnect(() => {
      const service = inject(Service);
      return service.prop$.pipe(
        switchMap(() => ) // map to some state object
      )
    })
  );
}
```

The same approach can be applied to `withHold` or `withEffects` (would be a perfect fit imo).

```ts
@Component({})
class TestComponent {
  readonly state = rxState(
    withHold((hold) => { // or withEffects?
      hold(interval$, sideEffectFn);
      hold(otherInterval$, sideEffectFn);
    })
  );
}
```

wdyt?

BioPhoton:
I really like the withConnect for multiple!

Should we consider the smosh operator pr too here?

edbzn:
I implemented your suggestions @hoebbelsB, I'm now working on the InjectionToken support (to provide the state using DI).

BioPhoton:
# `withHold`

I would want to unify the effect service and the effect function from state service in this PR.

To do so i suggest we at least sketch the effect also here so soo if we are align with its API.

edbzn:
Here is what I will implement for `rxEffect`:

```ts
@Component({})
class TestComponent {
  readonly destroy1 = rxEffect(triggerAndEffect$);
  readonly destroy2 = rxEffect(trigger$, sideEffect);
}
```
Here is what we have with `hold`:

```ts
@Component({})
class TestComponent {
  readonly state = rxState(
    withHold((hold) => {
      hold(interval$, sideEffectFn);
      hold(otherInterval$, sideEffectFn);
    })
  );
}
```

@BioPhoton What do you have in mind to align both?

libs/state/src/lib/inject.ts
edbzn: @hoebbelsB Even if we use DestroyRef, I would suggest marking all those new APIs as experimental.

edbzn: I'll rebase once #1567 is merged.

libs/state/actions/src/lib/inject.ts
```diff
@@ -0,0 +1,32 @@
...
+export function rxActions<
+  Actions extends object,
+  Transforms extends ActionTransforms<Actions> = {}
+>(
+  withTransformsFn?: (
+    rxActionFactory: RxActionFactory<Actions>
+  ) => RxActions<Actions, Transforms>
+): RxActions<Actions, Transforms> {
+  const errorHandler = inject(ErrorHandler, { optional: true }) ?? undefined;
+  const rxActionFactory = new RxActionFactory<Actions>(errorHandler);
```
edbzn, line 22: @BioPhoton I think we should refactor the RxActionFactory and use the inject function as well for the error handler, I'm not sure to get why it is optional tho.

BioPhoton:
@edbzn thx for the example and question of hold

as hold id a effect and it should be align with RxEffect i edited the comment to "effect". 
I think it is also a better name as it aligns more with effect from signals
WDYS?

```
@Component({})
class TestComponent {
  readonly state = rxState(
    withEffect((hold) => {
      effect(interval$, sideEffectFn);
      effect(otherInterval$, sideEffectFn);
    })
  );
}
```

libs/state/actions/src/lib/inject.ts
BioPhoton: The reason that is it optional is if I don't pass anything, Angular's default ErrorHandler is used

edbzn: I refactored the ErrorHandler usage in RxActionFactory and RxEffects to use inject instead of DI through ctors in [07a916b](https://github.com/rx-angular/rx-angular/pull/1543/commits/07a916b36c1d27ce569d7ad3919ebcf5dd6ef48e). However, I kept the provider optional to not break the previous behavior.

BioPhoton (changes requested):
As there is a lot of discussions let's chunk it up in:
- actions
- effect
- state

this way we will have proper time to test and think.

we cna also use a "next" version for that if we like

WDYS?

PR #1553, opened 2023-04-26
fix(state): handle error emission in `connect` using `ErrorHandler`

### LINKED ISSUE #1536: Handle error from Observable in rxState.connect()
# Description

RxStat.connect() doesn't handle errors from Observable<>

```
a$ = of(null).pipe(
  switchMap(() => throw new Error('error'))
);
somestate.connect(a$);
```
ErrorHandler will not call in this case

RxState 2.0.0

edbzn:
It's potentially a breaking change as with these changes instantiating RxState outside of Angular context will raise the following error: 

```
NG0203: inject() must be called from an injection context such as a constructor, a factory function, a field initializer, or a function used with `EnvironmentInjector#runInContext`. Find more at https://angular.io/errors/NG0203
```

Fixes #1536

### DIFF
--- a/libs/state/selections/src/lib/accumulation-observable.ts
+++ b/libs/state/selections/src/lib/accumulation-observable.ts
@@ -25,11 +25,12 @@ const defaultAccumulator: AccumulationFn = <T>(st: T, sl: Partial<T>): T => {
   return { ...st, ...sl };
 };
 
-export function createAccumulationObservable<T extends object>(
+export function createAccumulationObservable<T extends object>({
   stateObservables = new Subject<Observable<Partial<T>>>(),
   stateSlices = new Subject<Partial<T>>(),
-  accumulatorObservable = new BehaviorSubject(defaultAccumulator)
-): Accumulator<T> {
+  accumulatorObservable = new BehaviorSubject(defaultAccumulator),
+  handleError = (e: any) => console.error(e),
+} = {}): Accumulator<T> {
   const signal$ = merge(
     stateObservables.pipe(
       distinctUntilChanged(),
@@ -45,7 +46,7 @@ export function createAccumulationObservable<T extends object>(
     ),
     tap(
       (newState) => (compositionObservable.state = newState),
-      (error) => console.error(error)
+      (error) => handleError(error)
     ),
     // @Notice We catch the error here as it get lost in between `publish` and `publishReplay`. We return empty to
     catchError((e) => EMPTY),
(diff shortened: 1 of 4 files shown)


libs/state/src/lib/rx-state.service.ts
```diff
@@ -53,8 +53,10 @@ export type ProjectValueReducer<T, K extends keyof T, V> = (
 @Injectable()
 export class RxState<T extends object> implements OnDestroy, Subscribable<T> {
   private subscription = new Subscription();
-
-  private accumulator = createAccumulationObservable<T>();
+  private handleError = inject(ErrorHandler).handleError;
+  private accumulator = createAccumulationObservable<T>({
+    handleError: this.handleError,
+  });
```
hoebbelsB, line 59: We have to make sure the correct execution context is provided. Not sure if the current solution works if you try to access fields inside the `handleError` function.

e.g.

```ts
class ErrorHandler {
  handleError() {
    if (this.env === 'prod') {
       // do something in prod
    }
  }
}

```

could we create an test case for that?

```suggestion
  private errorHandler = inject(ErrorHandler);
  private accumulator = createAccumulationObservable<T>({
    handleError: this.erroHandler.handleError.bind(this.errorHandler),
  });
```

hoebbelsB (changes requested):
hello @edbzn thx for the PR :)

I'm not even sure if my comment is actually correct, but I guess we will have a problem with the execution context. If not, please ignore me and I'm going to approve. Sorry, had no time to actually check that, it just came to my mind ^^

libs/state/selections/src/lib/accumulation-observable.ts
```diff
@@ -45,7 +46,7 @@ export function createAccumulationObservable<T extends object>(
     ),
     tap(
       (newState) => (compositionObservable.state = newState),
-      (error) => console.error(error)
+      (error) => handleError(error)
     ),
     // @Notice We catch the error here as it get lost in between `publish` and `publishReplay`. We return empty to
     catchError((e) => EMPTY),
```
valentin78, line 52: because of it subscription doesn't work after first error catch

valentin78, line 52: example: 
`state. hold(actions,someAction$, () => api.getData())`
so, here, if api fails, we will not call api second time on trigger action

libs/state/src/lib/rx-state.service.ts
edbzn: I pushed the changes, but the test for checking that bind works is failing, I'm not sure to get why. If you have an idea @hoebbelsB.

PR #1558, opened 2023-04-29
docs(template):  add setup example with standalone cmp and fix typos

KevTale:
I added an example to the basic setup with standalone components and also fixed several typos or consistency stuff in the `@rx-angular/template` doc.

### DIFF
--- a/apps/docs/docs/template/api/rx-for-directive.md
+++ b/apps/docs/docs/template/api/rx-for-directive.md
@@ -107,28 +107,14 @@ The following context variables are available for each template:
 
 ## Setup
 
-The `ForModule` can be imported as following:
-
-Module based setup:
+The `RxFor` can be imported as following:
 
 ```
-import { ForModule } from "@rx-angular/template/for";
-
-@NgModule({
-  imports: [ForModule],
-  // ...
-})
-export class AnyModule {}
-```
-
-Standalone component setup:
-
-```
-import { ForModule } from "@rx-angular/template/for";
+import { RxFor } from "@rx-angular/template/for";
 
 @Component({
     standalone: true,
-    imports: [ForModule],
+    imports: [RxFor],
     template: `...`
 })
 export class AnyComponent {}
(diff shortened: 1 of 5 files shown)


libs/template/README.md
```diff
@@ -82,7 +84,23 @@ import { UnpatchModule } from '@rx-angular/template/unpatch';
...
+import { ForModule } from '@rx-angular/template/for';
+import { PushModule } from '@rx-angular/template/push';
+import { UnpatchModule } from '@rx-angular/template/unpatch';
+
+@Component({
+  standalone: true,
+  imports: [ForModule, LetModule, PushModule, UnpatchModule],
+  template: `...`,
+})
+export class AnyComponent {}
```
edbzn, line 103: I would suggest to stick to the standalone approach only, no need to keep the module-based approach IMO.

Note that we removed all the modules, `ForModule` should be replaced with `RxFor`, `LetModule` with `LetDirective`, `PushModule` by `PushPipe`, and `UnpatchModule` with `UnpatchDirective`.

apps/docs/docs/template/api/rx-for-directive.md
```diff
@@ -126,9 +126,9 @@ Standalone component setup:
 ```
 import { ForModule } from "@rx-angular/template/for";
 
-@NgComponent({
+@Component({
     standalone: true,
-    imports: [ ForModule ],
+    imports: [ForModule],
```
edbzn, line 131: Here as well, `RxFor` instead.

edbzn (commented):
Hey, thanks for the PR! We removed all the modules so we need to remove any reference to them in the doc and replace them with the standalone equivalent.

edbzn (approved):
Great! Is there any other reference to these modules in the doc?

KevTale:
> Great! Is there any other reference to these modules in the doc?

Good call @edbzn! There was a mention of these modules in cdk/render-strategies doc. I've fixed that and double checked in the entire repo, the only mention left are the actual implementation of the features so I think we are good to go. 👌

edbzn:
Thanks for the contrib @KevTale! :rocket:

PR #1562, opened 2023-05-03
fix(template): virtual-scrolling rxjs 6 compat. autosize update handling

hoebbelsB:
# Description

this PR fixes two issues which occurred to me in other projects where I tried to use the autosize strategy. The first issue is simple, it used an overload of the `tap` operator which isn't released in rxjs v6.

The other issue happens on item updates (and is maybe related to some flakyness of the cypress component-tests i've noticed in other PRs). The autosize strategy before only was able to process scroll updates. updates made to single views could lead to completely broken lists.


libs/template/experimental/virtual-scrolling/src/lib/scroll-strategies/autosize-virtual-scroll-strategy.ts
```diff
@@ -290,9 +290,7 @@ export class AutoSizeVirtualScrollStrategy<
   ): void {
     this.viewport = viewport;
     this.viewRepeater = viewRepeater;
-    this.resizeObserver = new ResizeObserver((events) => {
-      this.viewsResized$.next(events);
-    });
+    this.resizeObserver = new RxaResizeObserver();
```
edbzn, line 293: Why not using DI instead of manual instantiation?

libs/template/experimental/virtual-scrolling/src/lib/scroll-strategies/autosize-virtual-scroll-strategy.ts
```diff
@@ -303,7 +301,7 @@ export class AutoSizeVirtualScrollStrategy<
     this.viewport = null;
     this.viewRepeater = null;
     this._virtualItems = [];
-    this.resizeObserver?.disconnect();
+    this.resizeObserver?.destroy();
```
edbzn, line 304: From what I see the resizeObserver should always be present.

```suggestion
    this.resizeObserver.destroy();
```

libs/template/experimental/virtual-scrolling/src/lib/scroll-strategies/autosize-virtual-scroll-strategy.ts
```diff
@@ -588,37 +571,33 @@ export class AutoSizeVirtualScrollStrategy<
...
-                    index = i + adjustIndexWith;
-                    position += this.getItemSize(index);
-                  }
-                  this.contentSize =
-                    position + this.getRemainingSizeFrom(index + 1);
-                })
-              )
-            )
+            mergeMap((toObserve) => this.observeViewSizes$(toObserve)),
+            // coalesceWith(unpatchedAnimationFrameTick()),
```
edbzn, line 580: Is it intentional to leave this comment?

libs/template/experimental/virtual-scrolling/src/lib/scroll-strategies/resize-observer.ts
```diff
@@ -0,0 +1,37 @@
...
+  private resizeObserver: ResizeObserver;
+
+  /** @internal */
+  private readonly viewsResized$ = new Subject<ResizeObserverEntry[]>();
+
+  constructor() {
+    this.resizeObserver = new ResizeObserver((events) => {
+      this.viewsResized$.next(events);
+    });
+  }
```
edbzn, line 13: Could be declared as prop initialization so you don't need the constructor.

libs/template/experimental/virtual-scrolling/src/lib/scroll-strategies/resize-observer.ts
```diff
@@ -0,0 +1,37 @@
...
+    this.resizeObserver!.observe(element, options);
+    return new Observable<ResizeObserverEntry>((observer) => {
+      const inner = this.viewsResized$.subscribe((events) => {
+        const event = events.find((event) => event.target === element);
+        if (event) {
+          observer.next(event);
+        }
+      });
+      return () => {
+        this.resizeObserver?.unobserve(element);
```
edbzn, line 28:
```suggestion
        this.resizeObserver.unobserve(element);
```

libs/template/experimental/virtual-scrolling/src/lib/scroll-strategies/resize-observer.ts
```diff
@@ -0,0 +1,37 @@
+import { Observable, Subject } from 'rxjs';
+
+export class RxaResizeObserver {
+  private resizeObserver: ResizeObserver;
```
edbzn, line 4:
```suggestion
  private readonly resizeObserver = new ResizeObserver((events) => {
      this.viewsResized$.next(events);
  });
```

libs/template/experimental/virtual-scrolling/src/lib/scroll-strategies/autosize-virtual-scroll-strategy.ts
hoebbelsB: would be possible, yes. Do u have an issue with skipping this for a next PR?

libs/template/experimental/virtual-scrolling/src/lib/scroll-strategies/autosize-virtual-scroll-strategy.ts
hoebbelsB: would be solved by using DI as well

libs/template/experimental/virtual-scrolling/src/lib/virtual-list-template-manager.ts
```diff
@@ -11,7 +11,8 @@ export interface RxVirtualListChange<T, C> {
 
 export type RxVirtualListChanges<T, C> = [
   number,
-  () => RxVirtualListChange<T, C>
+  () => RxVirtualListChange<T, C>,
+  boolean? // removed
 ][];
```
edbzn, line 16: What about using named tuples to be explicit? I'm guessing about the purpose of the elements and the corresponding name but something like:
```ts
export type RxVirtualListChanges<T, C> = [
  itemCount: number,
  onChange: () => RxVirtualListChange<T, C>,
  removed: boolean?
][];
```

edbzn (approved):

libs/template/experimental/virtual-scrolling/src/lib/virtual-list-template-manager.ts
hoebbelsB: didn't even know this is a thing

Karnaukhov-kh (approved):

PR #1563, opened 2023-05-03
fix(template): remove `scroll` listener

arturovt:
This requires the same object to be passed back to `removeEventListener` after calling `addEventListener` (otherwise it's a leak).


edbzn (commented):
Hi, I don't observe a leak on my side, removing the listener without passing the option again works well.

image

Following the [removeEventListener doc](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/removeEventListener#matching_event_listeners_for_removal), it would fail if another option would be passed like `{ capture: true }` but passing nothing means "match any event with this name and this handler".

hoebbelsB:
@arturovt thank u for the PR!! Is it possible to elaborate a bit on this? I did memory leak inspections before merging it, so same as @edbzn I cannot observe any problems with the current solution. Thanks for investigating btw :)

arturovt:
The spec defines the following:

> [Event listeners](https://dom.spec.whatwg.org/#concept-event-listener) can be removed by utilizing the [removeEventListener()](https://dom.spec.whatwg.org/#dom-eventtarget-removeeventlistener) method, passing **the same arguments**.

Browser engines store event listeners differently on elements, but usually by "joining" listeners and options.

Given the following example:

```ts
@Component({
  selector: 'app-root',
  template: `
    <section
      id="scroll"
      #scrollContainer
      style="height: 300px; width: 300px; overflow-y: scroll"
    >
      <p *ngFor="let _ of array">Hey</p>
    </section>
    <button (click)="destroy()"></button>
  `,
})
export class AppComponent {
  array = Array.from({ length: 40 });

  @ViewChild('scrollContainer', { static: true })
  scrollContainer!: ElementRef<HTMLElement>;

  private l = () => {
    console.log(this);
  };

  constructor(private ngModuleRef: NgModuleRef<unknown>) {}

  ngOnInit(): void {
    // @ts-ignore
    this.scrollContainer.nativeElement['__zone_symbol__addEventListener'](
      'scroll',
      this.l,
      { passive: true }
    );
  }

  ngOnDestroy() {
    // @ts-ignore
    this.scrollContainer.nativeElement['__zone_symbol__removeEventListener'](
      'scroll',
      this.l
    );
  }

  destroy() {
    this.ngModuleRef.destroy();
  }
}
```

If we click the 'destroy' button and save the snapshot in Firefox, we may see the folllowing:

Screenshot from 2023-05-08 16-33-59

Now let's find what captures section with its virtual address `0x7fe639b05900`:

```
$ fxsnapshot 67708.fxsnapshot 'nodes { id: 0x7fe639b05900 }'

CaptureMap {
    lambdas: IdVec(
        [
            LambdaInfo {
                arity: 0,
                parent: None,
                captured: {},
            },
        ],
        PhantomData<fxsnapshot::query::ast::LambdaId>,
    ),
    uses: IdVec(
        [],
        PhantomData<fxsnapshot::query::ast::UseId>,
    ),
}
```

Which means the event listener still captures the element.

Chrome's Blink also expects the reference to be passed when calling `removeEventListener`:

```cpp
bool EventTarget::removeEventListener(const AtomicString &event_type,
                                      const EventListener *listener,
                                      EventListenerOptions *options) {
  return RemoveEventListenerInternal(event_type, listener, options);
}
```

Firefox Gecko is also checking both listener and options passed:

```cpp
if (listener->mListener == aListenerHolder &&
    listener->mFlags.EqualsForRemoval(aFlags)) {
  mListeners.RemoveElementAt(i);
}
```

One of the engines that's doesn't care about options is Safari's WebKit:

```cpp
bool EventTarget::removeEventListener(const AtomicString& eventType, EventListener* listener, bool useCapture)
```

Since it doesn't have a signature where options should be passed when calling `removeEventListener`. Everything yields to the fact that browsers have different implementations of the spec.

Material is following the same approach and passes options back when removing event listeners: https://github.com/angular/components/blob/285f46dc2b4c5b127d356cb7c4714b221f03ce50/src/material/legacy-slider/slider.ts#L572-L573

PR #1571, opened 2023-05-31
fix(template): untrack subscription and unsubscription in push pipe

eneajaho:
Same issue as this one: https://github.com/angular/angular/issues/50382
Same fix as this one: https://github.com/angular/angular/pull/50522

Repro: https://stackblitz.com/edit/stackblitz-starters-mec8lw

### DIFF
--- a/libs/template/push/src/lib/push.pipe.ts
+++ b/libs/template/push/src/lib/push.pipe.ts
@@ -4,6 +4,7 @@ import {
   OnDestroy,
   Pipe,
   PipeTransform,
+  untracked,
 } from '@angular/core';
 import {
   RxStrategyNames,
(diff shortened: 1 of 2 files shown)


libs/template/push/src/lib/push.pipe.ts
```diff
@@ -179,26 +180,38 @@ export class RxPush implements PipeTransform, OnDestroy {
...
+              tap((v) => {
+                this._renderCallback?.next(v);
+              })
+            )
           )
         )
-      )
-      .subscribe();
+        .subscribe()
+    );
```
edbzn, line 214: A unit test to ensure untracked is called would be great. Other than that looks good.

eneajaho: Hello @edbzn 
I added the same test that is added for async pipe.

PR #1572, opened 2023-06-02
fix(demos): restore initial state of basics tutorial

michaelbe812:
It looks like that by accident the final result of the basics tutorial has been commited to `setup.start.comopnent.ts`.
This PR will restore the state before commit 91419a1ebcb64f91cd6fa0277b9cad8fd93cd3a7 and 78a79bef4d28df1d4b02389e519d789cdc99b74b


mikelgo:
pls note that there's a conflicting PR #1573

PR #1573, opened 2023-06-08
docs(demos): add link to exercise instructions into main readme of tutorials

michaelbe812:
Currently there's not really a link from the source code of the basics-tutorials to concrete instructions what should be done in each exercise.


BioPhoton:
Would you mind moving that out into a separate app?

mikelgo:
> Would you mind moving that out into a separate app?

Sure I can do this. Good idea 👍

mikelgo:
I did update the PR by creating a dedicated `tutorials`- app.

In order to match the existing styles etc, I used e.g. the `rxa-app-shell` component from the `demos`-app. However this of course will throw an `@nx-enforce-module-boundaries` linting error. 
How to handle? I did hesitate to create a lib e.g. `demos-shared` bc. I did not want to "pollute" them. 
What is your suggestion?

PR #1583, opened 2023-06-22
refactor(template): use `takeUntilDestroyed` to cleanup observables

edbzn:
Now we can use the built-in `takeUntilDestroy` operator and cleanup observable without using extra properties or even the onDestroy lifecycle hook which results in less boilerplate.

Notice: this feature is in [developer preview](https://angular.io/guide/releases#developer-preview) so it may change.


edbzn:
Something I didn't anticipate, takeUntilDestroy needs to be used in the injection context (ctors) so it needs a bit of refactoring to be working. Will close it for now.

PR #1586, opened 2023-07-04
fix(template): virtual-scroll scrollStrategy dependency error in es2022

samvimes01:
use inject for dependency injection to prevent class instance members undefined errors

When virtual scroll feature used in a project with "target": "es2022" it's broken because class filed uses dependency injection field before initialization
`readonly viewRange = this.scrollStrategy.renderedRange$;`
throws an error and nothing is rendered.

Using inject instead of constructor injection fixes the issue

### DIFF
--- a/libs/template/experimental/virtual-scrolling/src/lib/virtual-scroll-viewport.component.ts
+++ b/libs/template/experimental/virtual-scrolling/src/lib/virtual-scroll-viewport.component.ts
@@ -6,9 +6,9 @@ import {
   ContentChild,
   ElementRef,
   OnDestroy,
-  Optional,
   Output,
   ViewChild,
+  inject,
 } from '@angular/core';
 import { Observable, ReplaySubject, Subject } from 'rxjs';
 import { distinctUntilChanged, takeUntil } from 'rxjs/operators';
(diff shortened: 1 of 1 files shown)


libs/template/experimental/virtual-scrolling/src/lib/virtual-scroll-viewport.component.ts
```diff
@@ -142,7 +147,7 @@ export class RxVirtualScrollViewportComponent
    * in sync with the DOM when the next `renderCallback` emitted an event.
    */
   @Output()
-  readonly viewRange = this.scrollStrategy.renderedRange$;
+  readonly viewRange = this.scrollStrategy?.renderedRange$;
```
edbzn, line 150: I think we can remove the optional chaining here, `scrollStrategy` should always be defined.

libs/template/experimental/virtual-scrolling/src/lib/virtual-scroll-viewport.component.ts
```diff
@@ -151,17 +156,14 @@ export class RxVirtualScrollViewportComponent
    * item actually being visible to the user.
    */
   @Output()
-  readonly scrolledIndexChange = this.scrollStrategy.scrolledIndex$;
+  readonly scrolledIndexChange = this.scrollStrategy?.scrolledIndex$;
```
edbzn, line 159: Same here.

edbzn (commented):
I have added a comment about the optional chaining, could you change it? Thank you for the fix!

libs/template/experimental/virtual-scrolling/src/lib/virtual-scroll-viewport.component.ts
samvimes01: done

PR #1585, opened 2023-06-27
feat(state): introduce actions as a function

BioPhoton:
# Todos:
- [x] Docs Code
- [x] Doce Web
  - [x] Migrate from class based to functions
- [x] Tests
- [x] Code
- [ ] Don't deprecate old version
- [x] Refactor tests to before each usage
- [x] Group test information at the bottom
- [ ] Bundle analysis (measure the impact)

PR #1589, opened 2023-07-17
feat(state): add `rxState` function

edbzn:
The follow-up to #1585, introduce functional `rxState` leveraging the inject function.


apps/docs/docs/state/setup.mdx
```diff
@@ -0,0 +1,645 @@
...
+```typescript
+import { rxState } from '@rx-angular/state';
+import { RxFor } from '@rx-angular/template/for';
+
+@Component({
+  template: `<movie *rxFor="let movie of movies$" [movie]="movie" />`,
+  imports: [RxFor],
+})
+export class MovieListComponent {
+  private store = inject(Store<MovieState>);
```
eneajaho, line 60: Type inference doesn't work correctly with inject. 
```suggestion
  private store = inject<Store<MovieState>>(Store);
```

eneajaho: Or we remove the generic type completly

apps/docs/docs/state/setup.mdx
```diff
@@ -0,0 +1,645 @@
...
+import { RxIf } from '@rx-angular/template/if';
+import { map, catchError, startWith, endWith } from 'rxjs';
+
+@Component({
+  template: `
+    <loader *rxIf="loading$" />
+    <error *rxIf="error$" />
+    <movie *rxFor="let movie of movies$" [movie]="movie" />
+  `,
+  imports: [RxFor, RxIf],
```
eneajaho, line 92: Shouldn't we add loader, error and movie components in the imports for the example to work correctly when copied?

apps/docs/docs/state/setup.mdx
```diff
@@ -0,0 +1,645 @@
...
+    <div>{{ count$ | async }}</div>
+    <button (click)="increment.next()">increment</button>
+    <button (click)="decrement.next()">decrement</button>
+  `,
+})
+export class CounterComponent {
+  readonly count$ = this.state.select('count');
+
+  //highlight-next-line
+  @Input() set count(count: Observable<number> | number) {
```
eneajaho, line 204: Can we do this ? 
```suggestion
  @Input({ transform: coerceObservable }) set count(count: Observable<number> | number) {
```

apps/docs/docs/state/setup.mdx
```diff
@@ -0,0 +1,645 @@
...
+    // decrement
+    connect('count', this.decrement, ({ count }) => count--);
+  });
+}
+```
+
+## Updates based on previous state
+
+Often it is needed calculate your new state based off some input and a previous state. The following example shows a
+local filtering alogrithm implemented with `RxState`.
```
eneajaho, line 275:
```suggestion
local filtering algorithm implemented with `RxState`.
```

apps/docs/docs/state/setup.mdx
```diff
@@ -0,0 +1,645 @@
...
+
+Create a local `Service` by using `rxState` as factory function.
+
+```typescript
+import { InjectionToken, inject } from '@angular/core';
+import { rxState } from '@rx-angular/state';
+
+export const MovieState = new InjectionToken({
+  factory: () => {
+    // inject dependencies here
```
eneajaho, line 404: We can say also: Inject dependencies outside rxState

libs/state/src/lib/inject.ts
eneajaho, line 1: the name of the file is inject.ts? Shouldn't it be rx-state.ts ?

apps/docs/docs/state/setup.mdx
edbzn: Then we should add all missing properties like standalone, selector, imports... I would suggest keeping the example minimal.

PR #1596, opened 2023-07-23
refactor(state): add void transform to actions

BioPhoton:


libs/state/actions/src/lib/transforms.ts
```diff
@@ -21,24 +19,30 @@ export function stopPropagation(e: Event): Event {
...
   }
   return e as T;
 }
+
+/**
+ * @description
+ * This transform helps to void all arguments.
+ */
+export function toVoid(_?: unknown): void {
+  _;
```
hoebbelsB, line 46: why? because typescript complains otherwise?

PR #1601, opened 2023-08-04
feat(template): viewport and scrolling element separation

hoebbelsB:
# Description

This PR introduces changes to support custom scroll elements for the virtual-scrolling implementation. It enables developers to declare nodes as scrollable element for the `rx-virtual-scroll-viewport` which are not part of it. Furthermore, a `scrollWindow` directive is introduced, that will make the window object the scrollable element for `rx-virtual-scroll-viewport`.

- [x] feature
- [x] docs
- [x] tests

## Usage

It is also possible to define a scroll container which is not part of the `rx-virtual-scroll-viewport` itself.
This is useful if you like to include other containers into the scroll container of your virtual list.

> You find demos for custom scroll elements in the [`demos`](https://github.com/rx-angular/rx-angular/tree/main/apps/demos/src/app/features/template/rx-virtual-for) app of the rx-angular monorepo

**Perf notice**

The included `RxVirtualScrollStrategy`s are smart enough to detect if they are completely hidden from the viewport, by also accounting
for the size of the contents before & after. When not part of the viewport, it will render less views (minimum of `runwayItems` or `runwayItemsOpposite`).

### With user defined scroll element

In order to define a custom scroll container, you need to apply the `rxVirtualScrollElement` directive to the element you would like to act
as scroll container for the virtual scroll viewport.

```html
<div rxVirtualScrollElement>
  <div>You can place any content you like before</div>
  <rx-virtual-scroll-viewport [itemSize]="50">
    <div *rxVirtualFor="let item of items$;">{{ item }}</div>
  </rx-virtual-scroll-viewport>
  <div>You can also place any content you like after</div>
</div>
```

### With window scroll

`@rx-angular/template` also ships a directive to use the window as scroll element. This is especially useful for mobile applications.
To enable window scrolling, simply add the `scrollWindow` directive to the `rx-virtual-scroll-viewport`.

```html
<rx-virtual-scroll-viewport [itemSize]="50" scrollWindow>
  <div *rxVirtualFor="let item of items$;">{{ item }}</div>
</rx-virtual-scroll-viewport>
```


edbzn (approved):

lincolnthree:
> In order to define a custom scroll container, you need to apply the `rxVirtualScrollElement` directive to the element you would like to act

@hoebbelsB Is there a way to do this programmatically for elements that we may not have access to (e.g. a scrolling viewport that's contained in a 3rd party component's shadow dom. See Ionic `ion-content` )?

https://ionicframework.com/docs/api/content

Would be nice if we could pass the scrolling viewport element ref in directly. Is that doable with this patch?

hoebbelsB:
hey @lincolnthree it should be possible to do such things. You can provide your own custom `RxVirtualScrollElement` and pass any elementRef you like.

```ts
@Injectable()
export class CustomVirtualScrollElement implements RxVirtualScrollElement {
  private elementRef = new ElementRef(/* choose your element */);
  elementScrolled$ = from(this.elementRef.nativeElement, 'scroll');
  measureOffset() {
     /* return whatever offset there is, maybe you need to calculate it on the fly each time, depends on your use case */
   return 0;
  }
}


@Component({
   template: `<rx-virtual-scroll-viewport><!-- will use your custom element as scroll container --><rx-virtual-scroll-viewport/>`,
   providers: [{ provide: RxVirtualScrollElement, useClass: CustomVirtualScrollElement }]
})
export class MyComponent {}
```

Issue #1604, opened 2023-08-17
Isr should have the ability to dynamically adjust the key used to store the cache instead of always using the url of the request


Pii2:
# Problem Solved By The Feature

If you want to cache a page in different variants on the server side, e.g. a variant for logged-in and non-logged-in users, or for different viewports, this cannot be implemented with one cache per URL.

In such scenarios, the page cached in another variant would lead to a content shift or flickering when loading the page.

# Solution

Provide a factory method that allows to modify the key used to store the cache. 

For example:

```ts
    const isr = new ISRHandler({
        indexHtml,
        invalidateSecretToken: process.env['INVALIDATE_TOKEN'] || 'MY_TOKEN',
        enableLogging: false,
        cacheKeyFactory: (req: Request) => {
            if (req.cookies && req.cookies.access_token) {
                return `${$req.url}%authenticated`;
            }
            return req.url;
        }
    });
```

eneajaho:
@Pii2 I merged the MR from @maxisam which should fix this issue. Closing as completed. Let us know if that one fixes the issue after it gets released.

PR #1608, opened 2023-09-11
feat(isr): add the possibility to create the cache in several variants

Pii2:
Implementation for Issue #1604 

This implementation has no breaking changes. It allows to specify different variants of the cache, e.g. for users who have a cookie to verify their logged in status. If no additional variants are specified, the functionality is unchanged. Also no modification to existing CacheHandler implementations is necessary.

Any number of variants can be specified via the ISRHandler configuration.
Each variant has an identifier and a callback function that recognizes the variant for a given request.

To enable on-demand cache invalidation, a simulateVariant function can also be specified, which can modify the request so that it can be identified for the variant.

### DIFF
--- a/libs/isr/models/src/cache-handler.ts
+++ b/libs/isr/models/src/cache-handler.ts
@@ -24,7 +24,7 @@ export interface RenderVariant {
 
 export interface VariantRebuildItem {
   url: string;
-  cachKey: string;
+  cacheKey: string;
   reqSimulator: (req: Request) => Request;
 }
 
--- a/libs/isr/models/src/cache-handler.ts
+++ b/libs/isr/models/src/cache-handler.ts
@@ -1,3 +1,5 @@
+import { Request } from 'express';
+
 /**
  * CacheISRConfig is the configuration object that is used to store the
  * cache data in the cache handler.
@@ -12,6 +14,18 @@ export interface CacheData {
   html: string;
   options: CacheISRConfig;
   createdAt: number;
+}
+
+export interface RenderVariant {
+  identifier: string;
+  detectVariant: (req: Request) => boolean;
+  simulateVariant?: (req: Request) => Request;
+}
+
+export interface VariantRebuildItem {
+  url: string;
+  cachKey: string;
+  reqSimulator: (req: Request) => Request;
 }
 
 export abstract class CacheHandler {
--- a/libs/isr/models/src/isr-handler-config.ts
+++ b/libs/isr/models/src/isr-handler-config.ts
@@ -46,7 +46,7 @@ export interface ISRHandlerConfig {
   /**
    * An optional way to define multiple variants of a page.
    * This can be useful if the appearance page differs, for example,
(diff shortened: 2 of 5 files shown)


libs/isr/models/src/cache-handler.ts
```diff
@@ -14,6 +16,18 @@ export interface CacheData {
...
 
+export interface RenderVariant {
+  identifier: string;
+  detectVariant: (req: Request) => boolean;
+  simulateVariant?: (req: Request) => Request;
+}
+
+export interface VariantRebuildItem {
+  url: string;
+  cachKey: string;
```
eneajaho, line 27:
```suggestion
  cacheKey: string;
```

libs/isr/models/src/isr-handler-config.ts
```diff
@@ -42,6 +42,30 @@ export interface ISRHandlerConfig {
    * If not provided, defaults to true.
    */
   skipCachingOnHttpError?: boolean;
+
+  /**
+   * An optional way to define multiple variants of a page.
+   * This can be useful if the appearance page differs, for example,
+   * based on a cookie and the chaed variant would thus lead to a content shift.
```
eneajaho, line 49:
```suggestion
   * based on a cookie and the cached variant would thus lead to a content shift.
```

libs/isr/server/src/isr-handler.ts
```diff
@@ -70,19 +71,26 @@ export class ISRHandler {
...
     const urlWithErrors: Record<string, any> = {};
 
-    for (const url of urlsToInvalidate) {
+    // Include all possible variants in the list of URLs to be invalidated including
+    // their modified request to regenerate the pages
+    const variantUrlsToInvalidate =
+      this.getVariantUrlsToInvalidate(urlsToInvalidate);
+
+    for (const variantUrl of variantUrlsToInvalidate) {
+      const { cachKey, url, reqSimulator } = variantUrl;
```
eneajaho, line 80:
```suggestion
      const { cacheKey, url, reqSimulator } = variantUrl;
```

eneajaho:
Hello @Pii2 
Thank you very much for this! This feature makes sense! 
I would really like if we could have some docs for this feature so users can see and use it too! 

I would like to have the `cachKey` renamed to `cacheKey`.

Pii2:
Hi @eneajaho,
And thank you!

I'm glad to hear that the feature has received positive feedback.

I will work on a documentation as soon as I am back from my vacation :)

Best regards

Pii2:
Hi @eneajaho !
I have added documentation on how to use the feature. Currently it is the last chapter, please feel free to rearrange the order as you like.

Also let me know if you are still missing something.

Best regards

eneajaho:
@Pii2 Can you rebase the branch so we can merge this? Thank you again for this!

Pii2:
Hi @eneajaho  :)
I hope it worked.
It was a pleasure!

Issue #1617, opened 2023-10-15
Make the state readonly


jon9090:
I work with people who are not so smart, and they intend to make mistakes even though I explain to them.

For example:

```
loadSomething() {
   this.get().accounts = null;
   ...
}
```
The problem with that code, the state is changing unexpectedly and not by the rx-state flow (`connect`, `hold`), because of the rx-state reactivity. and because I get the actual state, I can change that because it references to the origin source.

The solution to prevent that is to make the state read-only. 

I want to prevent them to do that mistake. in Redux/ngrx the state is read-only.

edbzn:
Related to #1449, we should make it readonly with types, not runtime lock.

edbzn:
Fixed by #1618, thanks for raising this.

jon9090:
Thanks, @edbzn ! When will the new version be released so that I can start using it?

edbzn:
@jon9090 We just published the v16.1.0 release.

edbzn:
Bad news @jon9090, we just reverted these changes as it was breaking some applications and we haven't anticipated that. We are now discussing introducing again the change in v17 but we are still not sure.

jon9090:
@edbzn Well, the idea is to use the API that RxState provides such `set` or `connect` or `hold` right? 

What is good is when I can get the state and make some changes directly in anticipation way in my application flow.

I would tell them to fix that issue because, in the first place, they do something not by the best practices.

edbzn:
Yes, mutating the get return or the projection function is a bad practice, that's why we agreed on introducing the readonly type. What we hadn't anticipated is that it breaks applications that expect `RxState.get` to return `T`, not `readonly T`, it can result on tsc errors, particularly on arrays, so we can't release this as a minor change as we did. What you can do to prevent such usage is to write an ESLint rule to prevent that.

Issue #1449, opened 2022-10-26
RxState should be readonly?


jon9090:
# Description

When I change the company object, it changes the store, also.

Which causes unexpected behavior in the state.

Why does the store not readonly? (using object freeze like ngrx)

```
vm$ = this.state.select(selectSlice(["company"])).pipe(
    map(({ company }) => ({ company })),
    tap(({ company }) => {
      company.id = 2; // <-------------- I change the id here, and it change the store because it's ref object.
      console.log({ company });
    })
  );
```

https://codesandbox.io/s/condescending-thompson-e51sh1?file=/src/app/app.component.ts:334-670

```
   template: `
    in app! vm = {{ vm$ | async | json }}<br />
    state = {{ state$ | async | json }}                          <---- state changed.
  `,
  providers: [RxState]
})
export class AppComponent {
  title = "CodeSandbox";

  state$ = this.state.select();

  vm$ = this.state.select(selectSlice(["company"])).pipe(
    map(({ company }) => ({ company })),
    tap(({ company }) => {
      company.id = 2; // <-- should be readonly??
      console.log({ company });
    })
  );

  constructor(private state: RxState<{ company: { id: number } }>) {
    this.state.set({ company: { id: 1 } });
  }
```

hoebbelsB:
hey @jon9090 that's a very nice finding, thanks! Could it be solved by just typing it as readonly object? :thinking:

jon9090:
@hoebbelsB thanks :)
I think it won't be enough because I select the state then I pass it to `FormGroup` and inside the controls can change the value..
If you want I can provide codesandbox example.

vmasek:
I like the idea of typing the state as `readonly` but I don't really understand the need to force (lock) the object on runtime. It actually comes with performance downside so I would not consider it a viable feature request.

edbzn:
+1 for `readonly` typing but not in runtime

edbzn:
Fixed by #1618, thanks for raising this.

PR #1618, opened 2023-10-15
feat(state): make state readonly

edbzn:
Fixes make the state readonly #1617, and also fixes RxState should be readonly? #1449.


hoebbelsB (approved):
love it, thanks

Issue #1526, opened 2023-03-02
RFC: `@rx-angular/state` - new `rxState`


hoebbelsB:
# Motivation

> All changes are non breaking

On top of the general motivations stated in #1592 a couple of things are worth to mention:

- As signals will land in angular 16, we want to provide first class support for the new reactive primitive provided by the
framework and integrate them smoothly with our existing state management solution. The goal is to be able to combine the power of `Observable`s and `Signal`s and integrate them seamlessly to have the best of both worlds.
- As we have now a onDestroy hook (`DestroyRef`) we no longer need to rely on the class based architecture at all and use the inject functions only.
- We can reconsider the old shorthand (`RxState#hold`) for a direct connection between state and effect. 
- We can include a signal compatible API

We've came up with the following core features that have to be supported (on top of all existing observable based APIs):

* connect accepts `Signal`
* get computed from state
* get single property `Signal` from state
* computed is added as top level API
* create `Effect` from on state for:
  * selection
  * computed

# Minimal Example

```typescript
import { rxState } from '@rx-angular/state/state';

@Component()
export class AnyComponent {
  state = rxState<{ name: string }>();
}
```

# State

The following code snippets and explanations showcase the APIs as we think they should be implemented in `RxState`.

## Usage in Component

```ts
type ListViewState = {
    list: Movie[],
    query: string
};

type ListViewUi = {
    searchInput: string;
};

@Component({
    selector: 'list-view',
    template: `
      <input name="search" (change)="ui.searchInput($event)" /><br/>
      <ul>
         ...
      </ul>
  `
})
export class ListViewContainerComponent {
   protected ui = rxActions<ListViewUi>();
    private movieState = inject(MovieService);
    private state = rxState<ListViewState>(({connect}) => {
        connect('list', this.movieState.movies);
        connect('query', this.ui.searchInput$.pipe(debounceTime(300)));
    });
}
```

## Usage in a Service

```ts
type MovieStateActions = {
    refresh: void
    save: MoviePostDto
}
type MovieState = {
    movies: Movie[]
}

@Injectable({
    providedIn: 'root'
})
export class MovieService {
    private moviesResource = inject(MovieResource);
    private state = rxState<MovieState>(({set, connect}) => {
        set({movies: []});
        connect('movies', this.actions.save$,
            ({movies}, updatedMovie) => upsert(movies, updatedMovie, 'id')
        );
        connect(
            'movies',
            this.actions.refresh$.pipe(
[...]

PR #1630, opened 2023-10-25
feat(state): introduce signal APIs

hoebbelsB:
# Description

based on #1526 & #1523 and our latest discussion, this PR should finally introduce the new signal APIs for `RxState`.

* [x] connect accepts signals
* [x] get computed from state
* [x] read signal from state
* [x]  docs
* [x]  unit tests (1 is missing, as we don't have `flushEffects` yet available
* [x] unit test `SignalStateProxy`
* [x] discuss `computedFrom` @eneajaho 

```ts
// component state
private state = rxState<{ todos: Todo[]; filter: string }>(({ set })) => {
  set({ todos: [], filter: '' });
})
```

**Connect a signal**

```ts
private todos: Signal<Todo[]> = signal(todoData);

this.state.connect('todos', todos);
```

**Computed from state**

```ts
// takes a function that gets each property of the state as `Signal<T[K]>`.
// in this case, todos will be `Signal<Todo[]>`

const count: Signal<number> = this.state.computed(({ todos }) => {
    return todos().length;
});

const filteredTodos: Signal<Todo[]> = this.state.computed(({ todos, filter }) => {
    return todos().filter(todo => todo.title.includes(filter()))
});
```

**Signal from state**

```ts
const todosSignal: Signal<Todo[]> = this.state.signal('todos');
```

**Mix computed w/ Observables "computedFrom"**

> TBD


edbzn (commented):
Looks good! :+1:

libs/state/src/lib/rx-state.spec.ts
```diff
@@ -95,11 +95,86 @@ describe(rxState, () => {
...
+
+      state.set({ count: 1 });
+      expect(count()).toBe(1);
+
+      state.connect(of({ count: 2 }));
+      expect(count()).toBe(2);
+    });
+
+    xit('should connect a signal', () => {
+      // TODO: we need TestBed flushEffect for it
```
eneajaho, line 136: Instead of using flushEffect we can call the signal in cmp template and run detectChanges, it will act the same as the templates act as effects.

libs/state/src/lib/rx-state.spec.ts
```diff
@@ -95,11 +95,86 @@ describe(rxState, () => {
...
     expect(RxState.prototype.ngOnDestroy).toHaveBeenCalled();
   });
+
+  describe('signals', () => {
+    it('should be undefined when key is undefined', () => {
+      const { component } = setupComponent<{ count: number }>();
+      const state = component.state;
+      const count = state.signal('count');
+
+      expect(count()).toBe(undefined);
```
eneajaho, line 105: Does the type also include undefined?

eneajaho: Yes it is, because signals need an initial value.

hoebbelsB: the initial value can be `undefined` in case the state has no value for that key. Same is true when calling `get`, yet it's not part of the type.

libs/state/src/lib/rx-state.service.ts
```diff
@@ -1,4 +1,13 @@
...
+  computed,
+  inject,
+  Injectable,
+  Injector,
+  isSignal,
+  OnDestroy,
+  Signal,
+} from '@angular/core';
+import { toObservable } from '@angular/core/rxjs-interop';
 // eslint-disable-next-line @nx/enforce-module-boundaries
```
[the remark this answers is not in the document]
hoebbelsB: no, my ide is yelling at me?

libs/state/src/lib/signal-state-proxy.ts
```diff
@@ -0,0 +1,56 @@
+import {
+  DestroyRef,
+  inject,
+  signal,
+  Signal,
+  WritableSignal,
+} from '@angular/core';
+import { takeUntilDestroyed } from '@angular/core/rxjs-interop';
+import { Observable } from 'rxjs';
+// eslint-disable-next-line @nx/enforce-module-boundaries
```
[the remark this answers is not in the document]
hoebbelsB: no, my ide is yelling at me?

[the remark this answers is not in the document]
edbzn: Run the lint command, you should see no error.

libs/state/src/lib/rx-state.service.ts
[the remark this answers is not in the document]
edbzn: I tested you can remove it.

libs/state/src/lib/rx-state.service.ts
```diff
@@ -592,6 +654,42 @@ export class RxState<T extends object> implements OnDestroy, Subscribable<T> {
...
+   * @description
+   * Lets you create a computed signal derived from state and rxjs operators.
+   *
+   * @throws If the initial value is not provided and the signal is not sync. Use startWith() to provide an initial value.
+   */
+  computedFrom<Output>(
+    ...operators: OperatorFunction<T, Output>[]
+  ): Signal<Output> {
+    return toSignal(
+      // @ts-ignore
```
edbzn, line 687: Can we avoid that?

libs/state/src/lib/rx-state.spec.ts
```diff
@@ -95,12 +97,200 @@ describe(rxState, () => {
...
+      }>(({ set }) => {
+        set({ count: 1337, multiplier: 1 });
+      });
+      const state = component.state;
+
+      const multiplied = state.computedFrom(
+        pipe(
+          selectSlice(['count', 'multiplier']),
+          map(({ count, multiplier }) => count * multiplier)
+        )
```
edbzn, line 240: Do we need `pipe` here?

libs/state/src/lib/rx-state.spec.ts
```diff
@@ -95,12 +97,200 @@ describe(rxState, () => {
...
+      });
+      const state = component.state;
+
+      const multiplied = state.computedFrom(
+        pipe(
+          selectSlice(['count', 'multiplier']),
+          map(({ count, multiplier }) => count * multiplier),
+          delay(1000),
+          startWith(10)
+        )
```
edbzn, line 263: Same question, would be nice to test that we can pass a list of operators instead of using functional pipe.

libs/state/src/lib/rx-state.service.ts
eneajaho: Already avoided.

libs/state/src/lib/rx-state.spec.ts
eneajaho: @hoebbelsB  Working on having both ways

libs/state/src/lib/rx-state.spec.ts
eneajaho: Totally

edbzn (approved):

PR #1637, opened 2023-11-06
refactor: drop usage of `pluck` operator

edbzn:
Pluck operator is deprecated in RxJS v7 and removed in up coming v8, let's incrementally drop deprecated APIs to prepare v8 migration the best.


apps/demos/src/app/rx-angular-pocs/cdk/utils/rxjs/operators/select.ts
```diff
@@ -156,7 +168,11 @@ export function select<T>(
     if (!opOrMapFn || opOrMapFn.length === 0) {
       return state$.pipe(stateful());
     } else if (isStringArrayGuard(opOrMapFn)) {
-      return state$.pipe(stateful(pluck(...opOrMapFn)));
+      return state$.pipe(
+        stateful(
+          map((state) => opOrMapFn.reduce((acc, key) => acc?.[key], state))
```
edbzn, line 173: Question, do we want to use our `safePluck` implementation here instead?

hoebbelsB: it's in the pocs, let's not care about it

hoebbelsB (approved):
thanks!

PR #1638, opened 2023-11-06
refactor: drop `publish`, `publishReplay`, and `ConnectableObservable`

edbzn:
`publish`, `publishReplay`, and `ConnectableObservable` are deprecated in RxJS v7 and removed in upcoming v8.


apps/demos/src/app/rx-angular-pocs/cdk/render-aware/render-aware.ts
```diff
@@ -39,34 +51,33 @@ export function createRenderAware<U>(cfg: {
...
+      catchError((e) => {
         console.error(e);
         return EMPTY;
-      }),
-      publishReplay()
-    );
+      })
+    ),
+    {
+      connector: () => new ReplaySubject(),
```
edbzn, line 77: Is it intended here to have no buffer size limit? @hoebbelsB

hoebbelsB: you can safely remove those pocs. Let's not try to fix something that has no value

libs/state/effects/src/lib/effects.service.ts
```diff
@@ -74,7 +74,7 @@ export class RxEffects implements OnDestroy, OnDestroy$ {
   private static nextId = 0;
   readonly _hooks$ = new Subject<DestroyProp>();
   private readonly observables$ = new Subject<Observable<unknown>>();
-  // we have to use publish here to make it hot (composition happens without subscriber)
+  // we have to use share here to make it hot (composition happens without subscriber)
```
hoebbelsB, line 77: is this comment even true?

hoebbelsB: no, it's not... https://stackblitz.com/edit/rxjs-nwerqp?file=index.ts

hoebbelsB: probably it is meant in a different way

libs/state/selections/src/lib/accumulation-observable.ts
```diff
@@ -30,28 +30,35 @@ export function createAccumulationObservable<T extends object>(
...
-    tap(
-      (newState) => (compositionObservable.state = newState),
-      (error) => console.error(error)
-    ),
-    // @Notice We catch the error here as it get lost in between `publish` and `publishReplay`. We return empty to
-    catchError((e) => EMPTY),
-    publish()
+    {
+      connector: () => new Subject<T>(),
+      resetOnDisconnect: false,
```
hoebbelsB, line 55: is this really what we want?

> If false, upon disconnection, the connecting subject will remain the
connecting subject, meaning the resulting observable will not go "cold" again,
and subsequent repeats or resubscriptions will resubscribe to that same subject.

libs/state/effects/src/lib/effects.service.ts
edbzn: Why it is not true? `share` will multicast the stream (hot), so the computation is only made once even if there are multiple subscribers. [Here is an example](https://stackblitz.com/edit/rxjs-lhe4qp?file=index.ts).

hoebbelsB: `composition happens without subscriber` is not true for `share`, only for `connectable` after calling the `connect` method :)

edbzn:
TODO: check RxJS version compatibility

hoebbelsB:
@edbzn this def. changes our minimum rxjs requirement. `connectable` is not present in `6.5.3` 

image

PR #1642, opened 2023-11-09
chore: migrate to Angular v17

edbzn:
Angular v17 support


apps/ssr/server.ts.bak
vmasek, line 1: I guess this backup file could be deleted as new one will be at use and "backup" will live in git history

libs/cdk/package.json
```diff
@@ -42,7 +42,7 @@
     "url": "https://github.com/rx-angular/rx-angular.git"
   },
   "peerDependencies": {
-    "@angular/core": "^16.0.0",
+    "@angular/core": "^17.0.0",
```
lincolnthree, line 45: These version ranges seem to cause issues when building against newer angular major versions.

Thoughts on making them a little less restrictive? E.g. `>=17.0.0`

```
npm WARN Could not resolve dependency:
npm WARN peer @angular/core@"^16.0.0" from @rx-angular/cdk@16.0.0
npm WARN node_modules/@rx-angular/cdk
npm WARN   @rx-angular/cdk@"^16.0.0" from the root project
npm WARN   1 more (@rx-angular/template)
```

hoebbelsB: cc @LayZeeDK

edbzn: It's the goal, we chose to sync RxA major version with Angular, this way we can guarantee compatibility, the problem with `>=17.0.0` is that we cannot guarantee compatibility for the next majors, it's a bet.

lincolnthree: Makes sense. Thanks!

apps/ssr/server.ts
```diff
@@ -1,44 +1,56 @@
...
-  const indexHtml = existsSync(join(distFolder, 'index.original.html')) ? 'index.original.html' : 'index';
+  const indexHtml = existsSync(join(distFolder, 'index.original.html'))
+    ? join(distFolder, 'index.original.html')
+    : join(distFolder, 'index.html');
 
-  // Our Universal express-engine (found @ https://github.com/angular/universal/tree/master/modules/express-engine)
-  server.engine('html', ngExpressEngine({
-    bootstrap: AppServerModule,
-  }));
+  const commonEngine = new CommonEngine();
```
BioPhoton, line 18: Nice Learned something!

eneajaho: And that's not all that changed. There's a lot more.

eneajaho:
LGTM! Thanks  🎉

libs/cdk/package.json
```diff
@@ -42,7 +42,7 @@
...
     "rxjs": "^6.5.3 || ^7.4.0"
```
hoebbelsB, line 46: is the rxjs version still valid? We changed some APIs like `connectable` as I remember correctly.

edbzn: Yes, it's still valid, the PR where I changed some APIs is not merged yet (#1638), I'll check about compatibility in the corresponding PR.

Issue #1632, opened 2023-10-29
RxState: additional connect-overload 


mikelgo:
# Problem Solved By The Feature

Less Code needed to connect partial state observables. The current way is to call multiple times connect:
```ts
connect('a', a$)
connect('b', b$)
```

# Solution

I would like to add this overload to `connect`: 
`  connect(object: Partial<{ [P in keyof T]: Observable<T[P]> }>): void;`

With this overload one could just do `connect({a: a$, b: b$})`

Wdyt? If you like this feature request I would be happy to contribute the implementation

hoebbelsB:
i really like the idea, we also need this overload for `rxLet` ^^

hoebbelsB:
I think @BioPhoton worked on something that goes in this direction: https://github.com/rx-angular/rx-angular/pull/1192

mikelgo:
> #1192

Oh yes. `smosh` seems like exactly doing this. but it seems like there's no progress? 
Also if I see it correctly there it is not considered that the combineLatest will not emit if there's by accident a source connected which will not emit a value. This should be handled.

Just let me know how to proceed =) Shall we consider a similar implementation like `smosh` for this feature? I would then try to open a PR soon :)

mikelgo:
just want to mention that i started to work on this. I do have an initial implementation i just now need to write some tests

PR #1644, opened 2023-11-13
WIP Rx state/connect partial obs and signals

mikelgo:
WIP PR for #1632 

Tests are still in WIP


apps/docs/docs/state/tutorials/basic-tutorial/01-setup.md
```diff
@@ -87,7 +87,7 @@ We will use `set()` as we already have initial values assigned to the `initCompo
 
 ```typescript
  constructor(...) {
-  ...
+...
```
mikelgo, line 90: revert. Don't know where this is comming from

mikelgo:
@hoebbelsB as we had the initial discussion you might want to check this PR.
Tests are still WIP (need to get familiar with the `scheduled` helper and how to use it with signals correctly)

libs/state/src/lib/convert-partials-to.observable.ts
```diff
@@ -0,0 +1,45 @@
...
+  return combineLatest(observables).pipe(
+    map((values) => {
+      const result: Partial<T> = {} as Partial<T>;
+      for (let i = 0; i < keys.length; i++) {
+        // @ts-ignore
+        result[keys[i]] = values[i];
+      }
+      return result;
+    }),
+    coalesceWith(timer(0))
```
mikelgo, line 43: is this the correct coalescing operator?

libs/state/src/lib/convert-partials-to.observable.ts
```diff
@@ -0,0 +1,45 @@
...
+
+    if (isObservable(obsOrSignal)) {
+      return obsOrSignal?.pipe(startWith(undefined));
+    }
+    if (isSignal(obsOrSignal)) {
+      return toObservable(obsOrSignal, { injector });
+    }
+  });
+
+  return combineLatest(observables).pipe(
```
mikelgo, line 34: should I filter out any key which has a value of undefined?
Undefined might come from line 27

libs/state/src/lib/convert-partials-to.observable.ts
hoebbelsB: it is the correct operator, but why would you coalesce it on the next `macrotask`? It makes much more sense to coalesce on `microtask` instead. Also, it shouldn't trigger ngZone.

Please try with `coalesceWith(from(Promise.resolve()))` and use the unpatched promise from the rx angular cdk.

We have to write tests though, that confirm `RxState` is not running out of zone. That could potentially cause issues when state selections are running out of zone.

hoebbelsB: also, I think coalescing should happen before the mapping. This way, the mapping would be less often executed

libs/state/src/lib/rx-state.service.ts
```diff
@@ -394,32 +396,53 @@ export class RxState<T extends object> implements OnDestroy, Subscribable<T> {
...
       | Observable<T[K] | V>
       | Signal<T[K] | V>,
     projectValueFn?: ProjectValueReducer<T, K, V>
   ): void {
+    if (
+      isPartialOfSignalsOrObservablesGuard(keyOrInputOrSliceOrPartials$) &&
+      !projectOrSlices$ &&
+      !projectValueFn
+    ) {
+      console.log('isPartialOfSignalsOrObservablesGuard');
```
hoebbelsB, line 424: should be removed :)

libs/state/src/lib/convert-partials-to.observable.ts
mikelgo: makes absolutely sense, will rework it soon

libs/state/src/lib/convert-partials-to.observable.ts
```diff
@@ -0,0 +1,40 @@
...
+    coalesceWith(from(Promise.resolve())),
+    map((values) => {
+      const result: Partial<T> = {} as Partial<T>;
+      for (let i = 0; i < keys.length; i++) {
+        // @ts-ignore
+        result[keys[i]] = values[i];
```
edbzn, line 35: Can we get rid of the comment here?

libs/state/src/lib/convert-partials-to.observable.ts
```diff
@@ -0,0 +1,40 @@
...
+export function convertPartialsToObservable<
+  T extends object,
+  K extends keyof T
+>(
+  input: Partial<{ [K: string]: Observable<T[K]> | Signal<T[K]> }>,
+  injector?: Injector
+): Observable<Partial<T>> {
+  const keys = Object.keys(input);
+
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-return
```
edbzn, line 17: I would get rid of this comment and probably throw like so:

 ```ts
 const observables = keys.map((key) => {
    const obsOrSignal = input[key];

    if (isObservable(obsOrSignal)) {
      return obsOrSignal?.pipe(startWith(undefined));
    }
    if (isSignal(obsOrSignal)) {
      return toObservable(obsOrSignal, { injector });
    }

    throw new TypeError(`Expected type Observable or Signal for key "${key}"`);
  });
  ```
  
  Wdyt?

libs/state/selections/src/lib/utils/guards.ts
```diff
@@ -69,3 +70,22 @@ export function isStringArrayFunctionAndOptionalObjectTupleGuard<R>(
...
+export function isPartialOfSignalsOrObservablesGuard<T, K extends keyof T>(
+  arg: unknown
+): arg is Partial<{ [K: string]: Observable<T[K]> | Signal<T[K]> }> {
+  if (!isObjectGuard(arg)) {
+    return false;
+  }
+  for (const key in arg) {
+    // eslint-disable-next-line no-prototype-builtins
+    if (arg.hasOwnProperty(key)) {
+      // @ts-ignore
```
edbzn, line 83: I would also get rid of this comment.

libs/state/src/lib/rx-state.service.ts
```diff
@@ -394,32 +396,51 @@ export class RxState<T extends object> implements OnDestroy, Subscribable<T> {
...
+        inputOrSlice$ = keyOrInputOrSliceOrPartials$;
       } else {
         // why can't typescript infer the correct type?
         inputOrSlice$ = toObservable(
-          keyOrInputOrSlice$ as Signal<Partial<T> | V>,
+          keyOrInputOrSliceOrPartials$ as Signal<Partial<T> | V>,
           { injector: this.injector }
         );
       }
     }
```
edbzn, line 440: I think this nested if/else branching starts to be hard to read. I think there's a better way to handle overloads, would be nice to refactor it at some point.

libs/state/src/lib/convert-partials-to.observable.ts
mikelgo: Like!

libs/state/src/lib/rx-state.service.ts
mikelgo: Totally agree. Overloads in Typescript are a complete mess

libs/state/selections/src/lib/utils/guards.ts
mikelgo: done

libs/state/src/lib/convert-partials-to.observable.ts
mikelgo: done

libs/state/src/lib/convert-partials-to.observable.ts
mikelgo: done

libs/state/src/lib/rx-state.service.ts
mikelgo: @edbzn I started in a separate PR to simplify the `connect`-method: https://github.com/rx-angular/rx-angular/pull/1652 
wdyt? 
To me it looks better. However it is not finished yet. There are two failing test cases bc my guards are not perfect yet I believe

eneajaho:
Hello @mikelgo anything I can help with? In order to move this PR forward?

mikelgo:
> Hello @mikelgo anything I can help with? In order to move this PR forward?

hey @eneajaho thanks for asking. 
I just lack a bit of time right now :D And I'm struggling a bit on how to write the tests in the same way as the existing tests using the `scheduled`-helper from `rxjs`. I would need to do some conversion from signals to observable to use this helper, but not sure if this would be really the way to go (I would not do it).

PR #1648, opened 2023-11-24
fix(state): on shorthand shouldn't toLowerCase

Jordan-Hall:
This repairs an issue where the on shorthand for side effects only supports single case lower case action. 

repo step https://stackblitz.com/edit/stackblitz-starters-ltqcqj?file=src%2Fmain.ts,src%2Fdata%2Fstate.ts

### DIFF
--- a/libs/state/actions/src/lib/proxy.ts
+++ b/libs/state/actions/src/lib/proxy.ts
@@ -71,7 +71,11 @@ export function actionProxyHandler<T extends object, U extends object>({
 
       // the user wants to get a single EventEmitter and trigger a side effect on event emission
       if (prop.toString().startsWith('on')) {
-        const propName = prop.toString().slice(2).toLowerCase() as KeysOfT;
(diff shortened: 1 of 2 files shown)


edbzn (approved):
Hey @Jordan-Hall thanks for the fix! Could you just provide the scope in the commit message like so fix(state): ...

Jordan-Hall:
Sorry for delay, I had to sort work around for this out as we still on Angular 16 for next few days. Speaking @eneajaho  he mentioned adding a test case for it. I've added it the the current side effect test now. If you can think of a better commit title let me know

edbzn:
Thank you @Jordan-Hall, it seems that some tests are failing. If you have time to have a look, otherwise I'll check that later.

Jordan-Hall:
I'll have a look and resolve it very shortly.

edbzn:
Thank you for the fix @Jordan-Hall!

PR #1649, opened 2023-11-26
feat(state): provide state config through injection tokens

### LINKED ISSUE #1621: Provide `AccumulatorFn` through DI

I've looked through the new functional docs and either im blind or there really is no provider for the accumulator, to set it on a global scope. I think it would make sense to provide that, so parts of the application can have the same accumulator, without specifying the same accumulator multiple times (and possibly violating DRY).

# Solution

- Add a function (something like: `provideRxStateAccumulator(...)`) which takes a factory function¹ as parameter, which in turn returns the accumulator (be it a immerjs accumulator or just a shallow reference break (aka. `{...a, ...b}`)).
- The provider should be only available on module level (aka. `makeEnvironmentProviders(...)`), if it were on the Component Level, we could just use the `rxState` Function parameter.
- The provider should be overridable and only use the nearest accumulator and stop searching for another when one was found, or just use the default one when none were found.

# alternatives Considered

None

# Additional Context

¹ : The factory function should maybe be considered so that it wont accidently create a shared accumulator (i dont even know if that is possible at this point, but it's worth considering)

edbzn:
Add provider functions to configure state through injection tokens:

- `provideRxStateConfig`
  - `withAccumulator`
  - `withSyncScheduler`
  - `withScheduler`
  
 Usage example:

```ts
bootstrapApplication(AppComponent, {
  providers: [
    provideRxStateConfig(
      withAccumulator(myCustomAccumulator),
      withSyncScheduler(),
    ),
  ],
});
```
  
It also exposes underlying injection tokens: 
 
 - `RX_ACCUMULATOR_FN`
 - `RX_STATE_SCHEDULER`
 
And finally, deprecate `RxState.setAccumulator` function.

closes #1621

### DIFF
--- a/apps/docs/docs/state/getting-started.mdx
+++ b/apps/docs/docs/state/getting-started.mdx
@@ -1,14 +1,13 @@
 ---
-sidebar_label: Setup
+sidebar_label: Getting Started
 sidebar_position: 2
-title: Setup
-# Renamed from libs/state/docs/usage.md
+title: Getting Started
 ---
 
 import Tabs from '@theme/Tabs';
 import TabItem from '@theme/TabItem';
 
-## Basic Setup
+## Create a State Instance
 
 The new functional creation API lets you create and configure `RxState` in only one place.
 
@@ -113,8 +112,8 @@ export class MovieListComponent {
         // start with loading true
         startWith({ loading: true }),
         // when request completes, we can set loading to false
-        endWith({ loading: false })
-      )
+        endWith({ loading: false }),
+      ),
     );
   });
 
@@ -301,9 +300,7 @@ export class MovieListComponent {
     // use the oldState and the searchInput to calculate the new state
     //highlight-start
     connect('movies', this.search.valueChanges, (oldState, searchInput) => {
-      return oldState.movies.filter((movie) =>
-        movie.title.includes(searchInput)
-      );
+      return oldState.movies.filter((movie) => movie.title.includes(searchInput));
     });
     //highlight-end
   });
@@ -339,27 +336,22 @@ export class MovieListComponent {
 
   search = new FormControl<string>();
 
-  private state = rxState<{ movies: Movie[]; searchValue: string }>(
-    ({ set, connect }) => {
-      // set initial state
-      set({ movies: [], searchValue: string });
-      // connect global state to your local state
-      connect('movies', this.store.select('movies'));
+  private state = rxState<{ movies: Movie[]; searchValue: string }>(({ set, connect }) => {
+    // set initial state
+    set({ movies: [], searchValue: string });
+    // connect global state to your local state
+    connect('movies', this.store.select('movies'));
 
-      // use the oldState and the searchInput to calculate the new state
-      //highlight-next-line
-      connect('searchValue', this.search.valueChanges);
-    }
-  );
+    // use the oldState and the searchInput to calculate the new state
+    //highlight-next-line
+    connect('searchValue', this.search.valueChanges);
+  });
 
   // derive filteredMovies$ from your stored state as an observable
   //highlight-next-line
-  filteredMovies$ = this.state.select(
-    ['movies', 'searchValue'],
-    (movies, searchValue) => {
-      return movies.filter((movie) => movie.title.includes(searchValue));
-    }
-  ); // Observable<Movie[]>
+  filteredMovies$ = this.state.select(['movies', 'searchValue'], (movies, searchValue) => {
+    return movies.filter((movie) => movie.title.includes(searchValue));
+  }); // Observable<Movie[]>
 
   // derive filteredMovies from your stored state as a signal
   //highlight-next-line
@@ -372,7 +364,7 @@ export class MovieListComponent {
   filteredMovies = this.state.computedFrom(
     select('searchValue'),
     switchMap((searchValue) => this.movieResource.fetchMovies(searchValue)),
-    startWith([] as Movie[]) // needed as the initial value otherwise it will throw an error
+    startWith([] as Movie[]), // needed as the initial value otherwise it will throw an error
   ); // Signal<Movie[]>
 }
 ```
@@ -389,7 +381,7 @@ import { rxState } from '@rx-angular/state';
 export class MovieService {
   private resource = inject(MovieResource);
 
-  private state = rxState<{ movies: Movie[] }>(({ set, connect }) => {
+  readonly state = rxState<{ movies: Movie[] }>(({ set, connect }) => {
     // set initial state
     set({ movies: [] });
     // connect global state to your local state
@@ -401,12 +393,32 @@ export class MovieService {
 
   // select a property for the template to consume as a signal
   movies = this.state.signal('movies'); // Signal<Movie[]>
+}
+```
 
-  // expose the select method
-  select: typeof this.state.select = this.state.select.bind(this.state);
+## expose readOnly rxState from a Service
 
-  // expose the signal method
-  signal: typeof this.state.signal = this.state.signal.bind(this.state);
+If you only want to expose your `RxState` instance as a readonly state, you can use the new `asReadOnly()` function.
+This allows you to only expose APIs that allows consumers to read from your state. Write access remains private to the
+owner of the `RxState` instance.
+
+```typescript
+import { inject, Injectable } from '@angular/core';
+import { rxState } from '@rx-angular/state';
+
+@Injectable({ providedIn: 'root' })
+export class MovieService {
+  private resource = inject(MovieResource);
+
+  private readonly _state = rxState<{ movies: Movie[] }>(({ set, connect }) => {
+    // set initial state
+    set({ movies: [] });
+    // connect global state to your local state
+    connect('movies', this.resource.fetchMovies());
+  });
+
+  // consumers can use `get`, `select`, `signal` and `compute`
+  readonly state = this._state.asReadOnly();
 }
 ```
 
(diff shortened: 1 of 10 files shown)


hoebbelsB:
hey @edbzn, can we go a step further and let users replace the whole accumulation observable? I would LOVE to get rid of the `queueScheduler` in some projects, e.g. 

We can also introduce another token just for the scheduler and leave the rest as is, e.g. `RX_STATE_SCHEDULER` ...

libs/state/selections/src/lib/accumulation-observable.ts
```diff
@@ -20,16 +20,36 @@ import {
...
+ *
+ * @example
+ * providers: [
+ *  {
+ *   provide: RX_ACCUMULATOR_FN,
+ *   useValue: (state, slice) => ({ ...state, ...slice })
+ *  }
+ * ]
+ */
+export const RX_ACCUMULATOR_FN = new InjectionToken<AccumulationFn>(
```
hoebbelsB, line 40: shouldn't this be provided in root? How does it work if nothing is provided?

edbzn: If we provide it in root users cannot use different accumulator fn, I don't see a specific scenario where we have to specify multiple accumulators but I do think it can happen in large apps. What's your opinion?

If nothing is provided it uses the token factory which returns the default accumulator.

libs/state/selections/src/lib/accumulation-observable.ts
```diff
@@ -20,16 +20,36 @@ import {
...
+  }
+);
+
 export function createAccumulationObservable<T extends object>(
   stateObservables = new Subject<Observable<Partial<T>>>(),
-  stateSlices = new Subject<Partial<T>>(),
-  accumulatorObservable = new BehaviorSubject(defaultAccumulator)
+  stateSlices = new Subject<Partial<T>>()
 ): Accumulator<T> {
+  const accumulatorFn = inject(RX_ACCUMULATOR_FN);
```
hoebbelsB, line 51: should this be `{ optional: true, self: true }` ?

We probably don't want accumulators to trickle down. Imo this is very specific behavior per instance.

libs/state/selections/src/lib/accumulation-observable.ts
hoebbelsB: Okay, if it anyway falls back to the default it's cool. But other question, is a non-root provided token non-treeshakable? 

Of course users have to be very careful where and when to provide a custom accumulator. Imo accumulators are very tightly couple to the state instance. See my comment here: https://github.com/rx-angular/rx-angular/pull/1649#discussion_r1405484501

hoebbelsB:
@edbzn we should also adjust docs for this

edbzn:
Thanks for the review @hoebbelsB. I think having a token for the whole observable would only complicate the definition of a custom accumulator and I don't think about a scenario where it would be useful. Do you have one in mind? 

However, I like the next suggestion about adding another token for the scheduler. I'm not entirely sure but I'm again not sure what scenario it would deserve. Other schedulers like the asap one or the async one could possibly create race conditions and produce unpredictable state results, right?

edbzn:
> @edbzn we should also adjust docs for this

Yeah, good point I wanted to discuss it a bit before working on any docs.

hoebbelsB:
>Thanks for the review @hoebbelsB. I think having a token for the whole observable would only complicate the definition of a custom accumulator and I don't think about a scenario where it would be useful. Do you have one in mind?

agree 💯 

> However, I like the next suggestion about adding another token for the scheduler. I'm not entirely sure but I'm again not sure what scenario it would deserve. Other schedulers like the asap one or the async one could possibly create race conditions and produce unpredictable state results, right?

Yes, maybe just distinguish between queueScheduler and no scheduler? My use cases are mostly requiring to have it sync, so removing the scheduler entirely

libs/state/selections/src/lib/accumulation-observable.ts
edbzn: If we mark self: as true then we can't specify the token for multiple components using the DI three. Not sure it is what we want. It also has no need to be optional as there's a default factory, we want a default value anyway so the token should not be optional.

libs/state/selections/src/lib/accumulation-observable.ts
edbzn: Non-root tokens are not tree-shakable as far as I know. Do we want to mark it provided in root? Knowing that we won't be able to specify multiple accumulators per application, which seems to go against your statement: 

> Imo accumulators are very tightly couple to the state instance

I would let users decide where they want to specify a custom accumulator and document that will be applied for children components as well.

libs/state/selections/src/lib/accumulation-observable.ts
hoebbelsB: I don't understand how the defaultfactory works. Imo `inject` throws an error if it cannot find a provider for a requested token.

Anyway, I think it will be VERY uncommon to have it shared between multiple state instances. Having it without `{ self: true }` would also mean any parent component dictates the accumulator for any given child component. Given the fact the accumulator can change the entire behavior of your state, this is probably not what you want. There is also no chance of controlling this trickle-down effect, as it's baked into the DI tree. 

I'm still for having it `{ self: true }`. In case someone wants to have multiple states with the same accumulator, we can help how to define a wrapper function

hoebbelsB: let's ask for more opinions on this. @Karnaukhov-kh @eneajaho @BioPhoton

edbzn:
Looks good, but I would rewrite some commit messages (or squash?) to be more explicit, especially those starting with feat as they will end up in the changelog. @hoebbelsB @eneajaho

eneajaho:
We can squash I guess

hoebbelsB:
@edbzn will the pieces of the squash message be picked up for the changelog? There we can rewrite easily

edbzn:
@hoebbelsB No, it only displays individual commit messages with feat, fix, revert and perf, we could also squash everything in one commit? Up to you.

hoebbelsB:
please tell me how to do that, or can u just do it?

libs/state/src/lib/provide-rx-state-config.ts
```diff
@@ -0,0 +1,64 @@
+import { InjectionToken, Provider } from '@angular/core';
+import { AccumulationFn, RX_ACCUMULATOR_FN } from '../../selections/src';
+
+export const enum RX_STATE_CONFIGS {
+  Scheduler,
+  Accumulator,
+}
```
edbzn, line 11: I would suggest not using a const enum but rather a string + union type.

libs/state/src/lib/provide-rx-state-config.ts
```diff
@@ -0,0 +1,98 @@
...
+/**
+ * Injection token for the default scheduler for rxState.
+ * @param fn
+ */
+export function withSyncScheduler(): RxStateConfigFn {
+  return {
+    kind: RX_STATE_CONFIGS.Scheduler,
+    providers: [{ provide: RX_STATE_SCHEDULER, useValue: 'sync' }],
+  };
+}
```
edbzn, line 78: I'm not sure to understand the terminology here, sync is the name for the default scheduler that uses the `queueScheduler` internally, right?

libs/state/src/lib/provide-rx-state-config.ts
```diff
@@ -0,0 +1,98 @@
...
+
+export const RX_STATE_SCHEDULER = new InjectionToken<SchedulerLike | 'sync'>(
+  'RX_STATE_SCHEDULER',
+  {
+    providedIn: 'root',
+    factory: () => queueScheduler,
+  },
+);
```
edbzn, line 54: Inline doc is missing here.

edbzn:
@hoebbelsB I rebased everything to have a comprehensive git history and force pushed.

libs/state/src/lib/provide-rx-state-config.ts
hoebbelsB: no, sync is no scheduler -> sync. it's also described in the docs

edbzn: Ok that's why I was confused, so why the description is "injection token for the default scheduler", seems incorrect.

hoebbelsB: maybe it's worded a little bit weird

apps/docs/docs/state/getting-started.mdx
```diff
@@ -449,20 +461,117 @@ export class MovieListComponent {
...
+  ],
+});
+```
+
+### Scheduler
+
+By default, `RxState` observes changes and computes new states on the [`queueScheduler`](https://rxjs.dev/api/index/const/queueScheduler). You can modify
+this behavior by using the `withScheduler()` or `withSyncScheduler()` configuration features.
+
+The most applicable use case will be to have sync scheduling:
```
edbzn, line 490: It's not clear to me when / why we would need to use sync scheduling. I would suggest to provide more guidance here.

libs/state/src/lib/provide-rx-state-config.ts
edbzn:
```suggestion
/**
 * Provide injection token for configuring RxState with the sync scheduler.
 */
export function withSyncScheduler(): RxStateConfigFn {
  return {
    kind: RX_STATE_CONFIGS.Scheduler,
    providers: [{ provide: RX_STATE_SCHEDULER, useValue: 'sync' }],
  };
}
```

PR #1651, opened 2023-11-30
docs(CDK): Update import paths to zone-configurations

Jordan-Hall:


apps/docs/docs/cdk/zone-configurations/how-to-setup-zone-flags.md
```diff
@@ -123,7 +123,7 @@ import 'zone.js/dist/zone';
 1. Create file `zone-runtime.ts` parallel to your `polyfills.ts` and insert the following content:
 
 ```typescript
-import { zoneConfig } from '@rx-angular/cdk/zone-flags';
+import { zoneConfig } from '@rx-angular/cdk/zone-configurations';
```
Jordan-Hall, line 126: The rest of this document states the run time should appear after zone.js but I dont believe this is the case anymore due to it still running https://github.com/rx-angular/rx-angular/blob/9e2dbc07c9b75932ae1db2411ac0972ae90f0f6c/libs/cdk/zone-configurations/src/lib/zone-config.ts#L28

edbzn: Yes, the doc is incorrect, zone flags should be imported before.

PR #1652, opened 2023-11-30
refactor(state): simplify connect implementation

mikelgo:
Simplification and elimination of handling the `connect`-overloads.

for reference: https://github.com/rx-angular/rx-angular/pull/1644#discussion_r1408870876

### DIFF
--- a/libs/state/src/lib/rx-state.service.ts
+++ b/libs/state/src/lib/rx-state.service.ts
@@ -434,7 +434,7 @@ export class RxState<T extends object> implements OnDestroy, Subscribable<T> {
       const slice$ = keyOrInputOrSlice$.pipe(
         map((v) => projectionStateFn(this.accumulator.state, v as V))
       );
-      this.accumulator.nextSliceObservable(slice$);
+      this.accumulator.nextSliceObservable(slice$ as Observable<V>);
       return;
     }
 
@@ -448,7 +448,7 @@ export class RxState<T extends object> implements OnDestroy, Subscribable<T> {
       const slice$ = toObservable(keyOrInputOrSlice$, {
         injector: this.injector,
       }).pipe(map((v) => projectionStateFn(this.accumulator.state, v as V)));
-      this.accumulator.nextSliceObservable(slice$);
+      this.accumulator.nextSliceObservable(slice$ as Observable<V>);
       return;
     }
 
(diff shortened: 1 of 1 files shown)


mikelgo:
On my local machine all tests are passing 🤔

libs/state/src/lib/rx-state.service.ts
```diff
@@ -405,64 +405,104 @@ export class RxState<T extends object> implements OnDestroy, Subscribable<T> {
...
+      projectOrSlices$ &&
+      typeof projectOrSlices$ === 'function' &&
+      !projectValueFn
     ) {
-      this.accumulator.nextSliceObservable(inputOrSlice$);
+      const projectionStateFn = projectOrSlices$;
+      const slice$ = keyOrInputOrSlice$.pipe(
+        map((v) => projectionStateFn(this.accumulator.state, v as V))
+      );
+      this.accumulator.nextSliceObservable(slice$);
```
edbzn, line 437: Type error could be fixed like so: 

```suggestion
      this.accumulator.nextSliceObservable(slice$ as Observable<V>);
```

libs/state/src/lib/rx-state.service.ts
```diff
@@ -405,64 +405,104 @@ export class RxState<T extends object> implements OnDestroy, Subscribable<T> {
...
+      isSignal(keyOrInputOrSlice$) &&
+      projectOrSlices$ &&
+      typeof projectOrSlices$ === 'function' &&
+      !projectValueFn
+    ) {
+      const projectionStateFn = projectOrSlices$;
+      const slice$ = toObservable(keyOrInputOrSlice$, {
+        injector: this.injector,
+      }).pipe(map((v) => projectionStateFn(this.accumulator.state, v as V)));
+      this.accumulator.nextSliceObservable(slice$);
```
edbzn, line 451: Same here: 

```suggestion
      this.accumulator.nextSliceObservable(slice$ as Observable<V>);
```

edbzn (commented):
I love it! :rocket: On my end, I get the errors locally, they could be fixed by following the suggestions. One more thing, I would add a test case to check that the function throws in case of passing wrong params, you could use // @ts-expect-error to disable type checking in the test when passing wrong params. Wdyt?

edbzn (approved):

Issue #1606, opened 2023-09-05
Push pipe does not work with if it is used within an input


kevinkasper:
# Description

I would like to pass an observable of the parent component to a child component in a zoneless application.

For this I use the @rx-angular/template push pipe instead of the native angular async pipe to resolve the observable for the child component input. Unfortunately, the corresponding setter is not called.

If I use the push pipe outside of inputs, the value is rendered correctly.

If I use a separate push pipe to the push pipe of the input, the child component is also rendered correctly.

# Steps to Reproduce the Issue

The problem can be reproduced in the following Stackblitz MVP:

[Stackblitz](https://stackblitz.com/edit/stackblitz-starters-guwjvw?file=src%2Fmain.ts)

luii:
The thing thats messing with not displaying the value, but logging is the `delay(1_000)` operator.
From just glancing over i think the Scheduling is messing with you here, `delay` uses `setInterval` on every value greater than 0 (iirc from @BioPhoton 's Talk about Scheduling; a really good source of information btw).

If you remove the delay from the pipe it shows just fine. So thats why i believe that the scheduling on the macrotask (e.g. setInterval) level is wrong and since the application doesnt run change-detection it wont pick up on these changes made on the macro-level. 

I think the push pipes rendering strategy needs to be adjusted for that case (see [here](https://www.rx-angular.io/docs/cdk/render-strategies#features))

kevinkasper:
I tested all rendering strategies in advance and unfortunately no value was rendered with any of them.

I modified the MVP so that a REST call is sent instead of the delay. The same problem occurs here:

[Stackblitz](https://stackblitz.com/edit/stackblitz-starters-refnnd?file=src%2Fmain.ts)

kevinkasper:
I have looked at the code and within the `handleChangeDetection` method of the push pipe the `renderedValue` is set correctly.

However, since this is an impure pipe without zone.js, it is not called again, which I believe causes the result to not render.

Does anyone have an idea how to solve this problem?

hoebbelsB:
thx, going to have a look on this

hoebbelsB:
@eneajaho is this maybe a result of your [refactoring](https://github.com/rx-angular/rx-angular/commit/fbdbf5b330334d46319e3790bc78265bd0a2cf74)?

eneajaho:
[...]

PR #1654, opened 2023-12-02
fix(template): use correct CdRef instance in RxPush

hoebbelsB:
# Description

as elaborated in #1606 , using the `inject` method in pipes is causing them to use the ChildComponents CdRef. This is a problem for `RxPush`, as `detectChanges` won't update input bindings.

### DIFF
--- a/libs/template/push/src/lib/push.pipe.ts
+++ b/libs/template/push/src/lib/push.pipe.ts
@@ -91,8 +91,6 @@ export class RxPush implements PipeTransform, OnDestroy {
   /** @internal */
   private strategyProvider = inject(RxStrategyProvider);
   /** @internal */
-  private cdRef = inject(ChangeDetectorRef);
-  /** @internal */
   private ngZone = inject(NgZone);
   /**
    * @internal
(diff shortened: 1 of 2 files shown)


edbzn:
I think we have to check if this won't produce any problem related to [useDefineForClassFields](https://angular.schule/blog/2022-11-use-define-for-class-fields) tsc option (the reason why we introduced inject at the first place). One question also, what if we use { skipSelf: true } flag for the inject fn instead of constructor DI?

edbzn:
So I tried, and the option skipSelf doesn't fix the issue. However it seems that the wrong behavior comes from the constructor injection, not the inject function, cf [this comment](https://github.com/angular/angular/issues/50952#issuecomment-1622415512). I would like to have feedback from the Angular team before merging this. Do you mind creating an issue @hoebbelsB?

hoebbelsB:
Should we still ship it as hotfix as is now? The pipeline doesn't seem to have issues, what would you like to check about [useDefineForClassFields](https://angular.schule/blog/2022-11-use-define-for-class-fields) ?

@alxhub also told me he is already aware of that issue and he knows how to fix it

edbzn (approved):

bryanforbes:
Is there any chance a backport could be done for the 16.x versions?

hoebbelsB:
@bryanforbes yes, going to ship it this weekend

Issue #1634, opened 2023-10-31
It's much more readable to have a complete generic name instead of a single letter, just like variables.


eneajaho:
Global remark, don't think it's mandatory to fix it right now but I believe it's much more readable to have a complete generic name instead of a single letter, just like variables.

_Originally posted by @edbzn in https://github.com/rx-angular/rx-angular/pull/1630#discussion_r1377629766_

AdrianRomanski:
@eneajaho may i take that issue?

edbzn:
@AdrianRomanski go ahead!

AdrianRomanski:
@edbzn Sorry for a long time of waiting, i had covid : <, issue is already in PR

PR #1660, opened 2023-12-17
Readable generic names #1634

AdrianRomanski:
resolves https://github.com/rx-angular/rx-angular/issues/1634

I'm feel very sorry for a long time with this issue, but covid hits me hard last weeks : < 

**Test**
    ✔  nx run state:test (15s)
    ✔  nx run demos:test (924ms)
    ✔  nx run ssr:test (912ms)

 NX   Successfully ran target test for 3 projects 

Can't wait for the next issue i can take :D

### DIFF
--- a/libs/state/src/lib/rx-state.service.ts
+++ b/libs/state/src/lib/rx-state.service.ts
@@ -30,15 +30,20 @@ import {
 import { catchError, map, tap } from 'rxjs/operators';
 import { createSignalStateProxy, SignalStateProxy } from './signal-state-proxy';
 
-export type ProjectStateFn<T> = (oldState: T) => Partial<T>;
-export type ProjectValueFn<T, K extends keyof T> = (oldState: T) => T[K];
+export type ProjectStateFn<TYPE> = (oldState: TYPE) => Partial<TYPE>;
+export type ProjectValueFn<TYPE, KEY extends keyof TYPE> = (
+  oldState: TYPE
+) => TYPE[KEY];
 
-export type ProjectStateReducer<T, V> = (oldState: T, value: V) => Partial<T>;
+export type ProjectStateReducer<TYPE, VALUE> = (
+  oldState: TYPE,
+  value: VALUE
+) => Partial<TYPE>;
 
-export type ProjectValueReducer<T, K extends keyof T, V> = (
-  oldState: T,
-  value: V
-) => T[K];
+export type ProjectValueReducer<TYPE, KEY extends keyof TYPE, VALUE> = (
+  oldState: TYPE,
+  value: VALUE
+) => TYPE[KEY];
 
 /**
  * @description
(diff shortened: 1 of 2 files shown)


edbzn (commented):
Hey, thanks for the PR! What about setting up the corresponding eslint rule for [enforcing generics name](https://typescript-eslint.io/rules/naming-convention/#enforce-that-type-parameters-generics-are-prefixed-with-t)? I would also suggest using PascalCase for generics with a prefix like `TReturn` or `TKey`.

AdrianRomanski:
Sure i will do that

AdrianRomanski:
@edbzn i finished CamelCasing Generics, changed variable names, fix some doc issues, and line general doc cleanup

AdrianRomanski:
There are some erros from linter configuration for generics in other files, should i fix them also?

edbzn:
Before fixing everything, let's open the discussion with the rest of the team. What do you think about that rule @hoebbelsB @eneajaho?

libs/state/src/lib/rx-state.service.ts
```diff
@@ -60,22 +66,24 @@ export type ProjectValueReducer<T, K extends keyof T, V> = (
  * @docsPage RxState
  */
 @Injectable()
-export class RxState<T extends object> implements OnDestroy, Subscribable<T> {
+export class RxState<TType extends object>
```
edbzn, line 69: I would name the generic `TState` instead.

edbzn (commented):
Now that I see it, I'm less convinced by the prefix `T`. What do you think?

libs/state/src/lib/rx-state.service.ts
AdrianRomanski: I'm waiting with this fix until we will be sure with naming convention for generics, from the other hand happy christmas and new year to everyone :)

AdrianRomanski:
> Before fixing everything, let's open the discussion with the rest of the team. What do you think about that rule @hoebbelsB @eneajaho?

@edbzn  Do we have any update on this?

hoebbelsB:
I generally like the idea, but I don't get why we should prefix everything with `T`. Wouldn't `RxState<State>` be completely fine? What is the benefit of adding the `T` in the beginning?

edbzn:
Some projects do it (eg: [TanStack Query](https://github.com/TanStack/query/blob/main/packages/angular-query-experimental/src/create-base-query.ts#L17)) and some don't, just a matter of coding style. A prefix could help to distinguish generics against class names for example.

AdrianRomanski:
@hoebbelsB @edbzn I can make both versions if you want to compare them.

edbzn:
@AdrianRomanski I would suggest that we go without first, and see later if we want to have generics prefixed.

AdrianRomanski:
@edbzn @hoebbelsB 

Changes made from last commit:

- deleted every T prefix (rx-state)
- introduced consistency in naming convention for chaining generic names (Type**A**, Type**B** etc)
[before we used Key**One**, Key**Two** and Typa**A**, Type**B**]

ideas: rule for suffix naming convention in .eslintrc

**Letter Suffix**      
```
        "@typescript-eslint/naming-convention": [
          "error",
          {
            "selector": "typeParameter",
            "format": ["PascalCase"],
            "suffix": [
              "A", "B", "C", "D", "E", "F", "G", "H", "I", "J", "K", "L", "M",
              "N", "O", "P", "Q", "R", "S", "T", "U", "V", "W", "X", "Y", "Z"
            ]
          }
        ]
```

**Number Suffix**          
```
        "@typescript-eslint/naming-convention": [
          "error",
          {
            "selector": "typeParameter",
            "format": ["PascalCase"],
            "suffix": [
              "One", "Two", "Three", "Four", "Five", "Six", "Sevend", "Eight", "Nine", "Ten"
            ]
          }
        ]
```

Pros - Keeping consistency in generic names
Cons - We have to use it even if we are not planning to chain Generic names for example (**StateA, KeyA, ValueA**)

Rule will enforce as to moves like this one
```
@Injectable()
export class RxState<StateA extends object>
  implements OnDestroy, Subscribable<StateA>
{(...) }
```

edbzn (approved):

Issue #1659, opened 2023-12-12
feat: allow state.signal() to accept more than one field in order to increase the depth of selection


eneajaho:
```ts

interface State {
  movies: {
   data: Movie[];
   loading: boolean;
  }
}

state = rxState<State>(({ set }) => {
   set({ movies: { data: [], loading: false }})
});

// in order to get movie loading we can do only 

moviesLoading = this.state.computed(s => s.movies().loading);

// I want to be able to do
moviesLoading = this.state.signal('movies.loading');
// or
moviesLoading = this.state.signal('movies', 'loading');
// or
moviesLoading = this.state.signal(['movies', 'loading']);
```

eneajaho:
So, basically I want to bring what select can do, in signal method

BioPhoton:
You can use computed

luii:
just out of curiosity, when you select something nested, will then the nested property be watched for new values or is still the observable that was passed into it the producer?

hoebbelsB:
the new signal would only emit when the nested property changed

luii:
@hoebbelsB is this also the case with the already provided `select`?

hoebbelsB:
@luii yes

hoebbelsB:
> So, basically I want to bring what select can do, in signal method

I remember that the typing behind of it is awful. But i'm all up for it, maybe we have a pairing session on that?

AdrianRomanski:
@hoebbelsB @eneajaho I can take this one :)

PR #1662, opened 2023-12-23
feat: add signal method overloads to select nested keys

eneajaho:
Enables signal method overloads 🚀

I went with this approach:
```ts
moviesLoading = this.state.signal('movies', 'loading');
```

TODO: add docs 

Fixes https://github.com/rx-angular/rx-angular/issues/1659

cc @hoebbelsB @edbzn

### DIFF
--- a/libs/state/src/lib/rx-state.service.ts
+++ b/libs/state/src/lib/rx-state.service.ts
@@ -703,7 +703,65 @@ export class RxState<T extends object> implements OnDestroy, Subscribable<T> {
    * current keys value in RxState. Whenever the key gets updated, the signal
    * will also be updated accordingly.
    */
-  signal<K extends keyof T>(k: K): Signal<T[K]> {
+  signal<K1 extends keyof T>(k1: K1): Signal<T[K1]>;
+  /**
+   * @internal
+   */
+  signal<K1 extends keyof T, K2 extends keyof T[K1]>(
+    k1: K1,
+    k2: K2
+  ): Signal<T[K1][K2]>;
+  /**
+   * @internal
+   */
+  signal<
+    K1 extends keyof T,
+    K2 extends keyof T[K1],
+    K3 extends keyof T[K1][K2]
+  >(k1: K1, k2: K2, k3: K3): Signal<T[K1][K2][K3]>;
+  /**
+   * @internal
+   */
+  signal<
+    K1 extends keyof T,
+    K2 extends keyof T[K1],
+    K3 extends keyof T[K1][K2],
+    K4 extends keyof T[K1][K2][K3]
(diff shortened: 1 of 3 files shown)


libs/state/src/lib/signal-state-proxy.ts
```diff
@@ -8,30 +8,33 @@ import {
...
-  stateFn: <K extends keyof State>(k: K) => State[K]
+  stateFn: RxState<State>['get']
 ) {
   const destroyRef = inject(DestroyRef);
 
   const signalState = {} as SignalStateProxy<State>;
   return new Proxy<SignalStateProxy<State>>(signalState, {
     get<K extends keyof State>(
       target: SignalStateProxy<State>,
-      p: K | string | symbol
```
edbzn, line 26: Why removing the generic and union type here?

libs/state/src/lib/signal-state-proxy.ts
```diff
@@ -8,30 +8,33 @@ import {
...
+import { RxState } from './rx-state.service';
 
 export type SignalStateProxy<State extends object> = {
   [K in keyof State]: Signal<State[K]>;
 };
 
 export function createSignalStateProxy<State extends object>(
   state$: Observable<State>,
-  stateFn: <K extends keyof State>(k: K) => State[K]
+  stateFn: RxState<State>['get']
```
edbzn, line 19: Could you explain the type change here? I'm not sure why stateFn becomes type of RxState.get

libs/state/src/lib/rx-state.service.ts
```diff
@@ -703,7 +703,65 @@ export class RxState<T extends object> implements OnDestroy, Subscribable<T> {
...
+    k3: K3,
+    k4: K4,
+    k5: K5,
+    k6: K6
+  ): Signal<T[K1][K2][K3][K4][K5][K6]>;
+
+  /**
+   * @internal
+   */
+  signal<R>(...args: string[]): Signal<R> {
```
edbzn, line 763: What about naming it keys instead of args?

libs/state/src/lib/signal-state-proxy.ts
```diff
@@ -8,30 +8,33 @@ import {
...
       if (!_signal) {
-        const val = stateFn(p as K);
+        // get initial value from stateFn
+        const val = stateFn(...(keyFields as [K])); // get the value from the stateFn; TODO: fix types
         _signal = signal(val);
-        target[p as K] = _signal;
+        target[key] = _signal;
         state$
-          .pipe(select(p as K), takeUntilDestroyed(destroyRef))
+          .pipe(select(...(keyFields as [any])), takeUntilDestroyed(destroyRef))
```
edbzn, line 37: Not sure why we dropped the generic cast in favor of [any]?

edbzn (commented):
Hey, I left a few questions about the types.

Issue #1663, opened 2023-12-28
feat: allow unpatch to unpatch all except specific events


Laurens-makel:
# Problem Solved By The Feature

Currently the `unpatch` directive could be used in two ways:

1. Unpatch all events: `<div [unpatch]></div>`
2. Unpatch specific events: `<div [unpatch]='['dragenter', 'dragleave']"></div>`

This works fine, but becomes a bit problematic when you would like to exclude one or two events and unpatch all the others.

One would have to specify all the events, except the ones to exclude.

# Solution

Ideally one could write the following, which would imply that all events except `dragover` and `mouseover` should be unpatched.

`<div [unpatch]="['!dragover', '!mouseover']"></div>`

# Alternatives Considered

Specifying the entire list of events except the ones I do not want unpatched.

# Additional Context

In our app we render a lot of `<input>` fields, we noticed that alot of DOM events would trigger change detection on these fields. For example hovering, clicking, dragging over, whatever event would happen, would cause a CD cycle.

We decided to add `[unpatch]` but this had the side effect that a lot events such as `blur` and `change` would not trigger the `FormControl.valueChanges` observable to emit.

To solve this we would have to specify all the other events to the unpatch directive except these ones.

The docs show the following warning:
![image](https://github.com/rx-angular/rx-angular/assets/37303445/77414699-23ce-4eed-8e6a-c83fcc89c4db)

Shouldn't there be a warning to for when the above situation applies?

```
Warning: Do not use [unpatch] on elements that reference a FormControl. Specify all events (for now) except the `blur` and `change` events. Otherwise user input is ignored, the FormControl.valueChanges observable will not emit and attached validations to the FormControl will not run until next change detection that affects the component in which the element is rendered.
```

BioPhoton:
We could think of another option.

Some negation maybe. Do you have an idea of how the API could look like?

Laurens-makel:
> We could think of another option.
> 
> Some negation maybe. Do you have an idea of how the API could look like?

What about the initial draft in #1668 ?

Laurens-makel:
@hoebbelsB  @BioPhoton
Update regarding the additional context:

Turns out that a directive was applied to all rendered input fields, this directive provides additional functionality to by automatically padding/unpadding a desired character till a desired length. The user could type `123` and it would get padded to `0000000123` when the user has left the input field and would get unpadded to `123` again on `focus` and `mouseenter` .

[...]

PR #1668, opened 2023-12-29
feat(template): unpatch all but specified events

Laurens-makel:
#1663 - Initial draft idea

### DIFF
--- a/libs/template/unpatch/src/lib/tests/unpatch.directive.spec.ts
+++ b/libs/template/unpatch/src/lib/tests/unpatch.directive.spec.ts
@@ -116,4 +116,41 @@ describe(RxUnpatch.name, () => {
       addEventListener.mockRestore();
     }
   });
+
+  it('should re-apply all except provided negated event listeners', () => {
+    // Arrange
+    const fixture = TestBed.createComponent(TestComponent);
+    fixture.componentInstance.unpatch = ['!mouseenter'];
+    const appRef = TestBed.inject(ApplicationRef);
+    const div = fixture.debugElement.query(By.css('div'));
+    const addEventListener = jest.spyOn(
+      div.nativeElement,
+      Zone.__symbol__('addEventListener')
+    );
+    const removeEventListener = jest.spyOn(
+      div.nativeElement,
+      'removeEventListener'
+    );
+
+    // Act
+    fixture.detectChanges();
+    const tick = jest.spyOn(appRef, 'tick');
+    div.nativeElement.dispatchEvent(new Event('mouseenter'));
+    div.nativeElement.dispatchEvent(new Event('click'));
+
+    try {
+      // Assert
+      expect(logs).toEqual([
+        [LogEvent.Mouseenter, true],
+        [LogEvent.Click, false],
+      ]);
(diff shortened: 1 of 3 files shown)


edbzn (approved):
Very nice!

eneajaho:
LGTM! Thanks 🙏

eneajaho (approved):

Laurens-makel:
@eneajaho & @edbzn 
Thanks for the review! It's a pleasure to contribute to one of my favorite libraries. 

What about the additional context in the original issue related to this PR? I've created a demo repo in which I've tried to explain the problem. 

It's about optimizing change detection when the following the following factors are applied to a single `input` element:

- `unpatch` directive
- `FormControl` validators & errors 
- additional directive that provides `@HostListener` functions on `mouse` related events.

libs/template/unpatch/src/lib/tests/unpatch.directive.spec.ts
```diff
@@ -116,4 +116,41 @@ describe(RxUnpatch.name, () => {
...
   });
+
+  it('should re-apply all except provided negated event listeners', () => {
+    // Arrange
+    const fixture = TestBed.createComponent(TestComponent);
+    fixture.componentInstance.unpatch = ['!mouseenter'];
+    const appRef = TestBed.inject(ApplicationRef);
+    const div = fixture.debugElement.query(By.css('div'));
+    const addEventListener = jest.spyOn(
+      div.nativeElement,
```
hoebbelsB, line 127: one question, what happens when you put `['!mouseenter', 'mouseenter']` or `['!mouseenter', 'mousemove']`?

Laurens-makel: Good question.

My first thought here is, why would a user specify this? What is the intention, what is expected?

Secondly, when looking at the implementation, whenever the user specifies a negated event by adding the `!` prefix, all events except the negated ones are unpatched. Other non-negated events specified are basically ignored.

```typescript
const negatedEvents = this.events
        .filter((event) => event.startsWith('!'))
        .map((event) => event.replace('!', ''));

const nextEvents = negatedEvents.length
        ? zonePatchedEvents.filter(
            (zonePatchedEvent) => !negatedEvents.includes(zonePatchedEvent)
          )
        : this.events;
```

Do you think it would be a good idea to add unit tests to proof this behaviour and prevent possible regression in the future?

hoebbelsB: so negatedEvents always wins? We just need to put this information in the docs so no one is actually confused. You could also think about throwing an error in devMode when some weird input is specified

Laurens-makel: ok, gotcha.

I've updated the documentation, I hope it's more clear now.

Regarding validating input, I agree with that, we should also throw an error in dev mode when non-existing events are specified.

hoebbelsB:
hey @Laurens-makel could you rebase your branch? Maybe it fixes the pipeline

AdrianRomanski:
@Laurens-makel Would you mind to rebase the branch? If you have no time at the moment for that, i can help you

PR #1669, opened 2023-12-31
feat(template): viewport external attrs in virtual-scrolling

mlc-mlapis:
This PR allows passing external attributes to the runway viewport element through `@Input() viewportAttrs: Array<{name: string; value?: string}> = [];` It's very beneficial when integrating with other solutions, like scrollbars libs, which style and customize native scrollbars, and it's required to finally show only styled ones and protect flickering when custom ones replace native ones.

An example could be the [OverlayScrollbars](https://github.com/KingSora/OverlayScrollbars/tree/master/packages/overlayscrollbars-ngx) library. It uses the attribute `<div data-overlayscrollbars-initialize>` to skip flickering.

### DIFF
--- a/apps/docs/docs/template/api/virtual-scrolling.mdx
+++ b/apps/docs/docs/template/api/virtual-scrolling.mdx
@@ -84,6 +84,7 @@ all pre-packaged ScrollStrategies as well as control the majority of inputs.
 - works also with static variables `*rxVirtualFor="let i of myData"`
 - Immutable as well as mutable data structures (`trackBy`)
 - Notify when rendering of templates is finished (`renderCallback`)
+- Allow passing external attributes to the viewport element (`[viewportAttrs]="[{name: '...', value: '...'}]"`)
 
 **Performance Features**
 
--- a/libs/template/experimental/virtual-scrolling/src/lib/virtual-scroll-viewport.component.ts
+++ b/libs/template/experimental/virtual-scrolling/src/lib/virtual-scroll-viewport.component.ts
@@ -7,8 +7,11 @@ import {
   ContentChild,
   ElementRef,
   inject,
+  Input,
   OnDestroy,
+  OnInit,
   Output,
+  Renderer2,
   ViewChild,
   ViewEncapsulation,
 } from '@angular/core';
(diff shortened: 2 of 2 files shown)


edbzn:
Hey, yes this makes sense to me, what do you think @hoebbelsB?

Laurens-makel:
> Hey, yes this makes sense to me, what do you think @hoebbelsB?

I've noticed there weren't any `spec` files for this component or related directive before this PR was opened, but what about test coverage in this PR?

I see two options, one easier than the other.

1. Unit test in a `spec` file which asserts that the given `viewportAttrs` are actually passed to the rendered element.
2. Cypress based test that proofs the integration with one or more of the mentioned libraries, this might be undesirable since the scope of the tests suddenly includes external libraries. These tests could also be placed in their own 'suite' and excluded from the requirements to succeed a build and might just act as informational tests.

What do you think?

mlc-mlapis:
@Laurens-makel Hi, we hoped to hear a positive reaction from maintainers that our idea would be accepted before finalizing all the details. 😄

Laurens-makel:
> @Laurens-makel Hi, we hoped to hear a positive reaction from maintainers that our idea would be accepted before finalizing all the details. 😄

Don't consider me as a maintainter, I'm just peeking around at some libraries of interest and try to contribute.

Also don't consider my feedback as blocking, I probably could help you out on the unit test if desired :)

hoebbelsB:
hi @mlc-mlapis, first of all thanks for the contribution :-) Just to get the idea, you want to be able to set custom properties on the scrollable element? Isn't the ability to define a custom scrollable element enough for your use case? https://www.rx-angular.io/docs/template/api/virtual-scrolling#custom-scroll-elements

hoebbelsB:
> I've noticed there weren't any `spec` files for this component or related directive before this PR was opened, but what about test coverage in this PR?

@Laurens-makel there are cypress component tests that test most of the behavior of all components in the virtual-scrolling package. We haven't introduced code coverage extraction for it yet, though. And yes, we would need to introduce new tests if we change or add features

mlc-mlapis:
@hoebbelsB Thanks for your comments. I've been using a custom scrollable element at the beginning, but I had some problems to make it work. I was still determining whether there was a way, so I used another path and introduced the above PR, where it works, and confirmed that it can work.

If you think that such a PR is redundant, I will try the whole logic again and concentrate only on using your custom scrollable element. Let me have a day to confirm if it works or not.

mlc-mlapis:
@hoebbelsB Hi Julian, as I said, I tried to dust off the entire logic from the beginning once more, and ... you were right. Now, using your custom scrollable element works as expected. I don't know the difference, but actually, this PR is not necessary.

hoebbelsB:
i'm very glad to hear that, thank u for your contribution again!

Issue #976, opened 2021-10-09
Enable strict mode for libraries


hoebbelsB:
# Enable strict mode

Please enable strict mode for the rx-angular project based on [angulars guide](https://angular.io/guide/strict-mode) on how to configure it properly.

If it's more appropriate, please consider doing it in iterative approach and start with only 1 lib.

* [ ] #989
* [ ] #990
* [ ] #991

exequiel09:
The old times @santoshyadavdev seeing you doing large refactors 🥳

PR #1672, opened 2024-01-07
feat(cdk): add native `postTask` scheduling support

edbzn:
This pull request introduces an initial, **albeit naive**, implementation of the `scheduler.postTask` native scheduling API. The primary goal is to gather feedback before further refinement.

### Drawback of Custom Schedulers

The pull request addresses the limitation associated with custom schedulers. While it's feasible to create a custom scheduler to prioritize tasks within your controlled codebase, third-party scripts may not utilize the same scheduler. Consequently, prioritizing work becomes challenging in environments where you don't have control over all the code on the page. The alternatives are limited to chunking up tasks or explicitly yielding to user interactions. The new implementation aims to overcome these drawbacks by leveraging the native `scheduler.postTask` API.

### Further Work

- Task Aborting: Enhance the implementation to support task aborting, allowing for the cancellation of scheduled tasks.
- Task Prioritization Configuration: Explore options for configuring task prioritization based on specific requirements or application needs.
- Testing: Implement testing to ensure the reliability and compatibility of the native scheduler.postTask API integration across different browsers and environments.

### DIFF
--- a/libs/cdk/internals/scheduler/src/lib/scheduler.ts
+++ b/libs/cdk/internals/scheduler/src/lib/scheduler.ts
@@ -18,6 +18,10 @@ import { PriorityLevel } from './schedulerPriorities';
  */
 declare const ngDevMode: boolean;
 
+const isPostTaskSupported =
+  typeof ɵglobal.scheduler !== 'undefined' &&
+  typeof ɵglobal.scheduler.postTask === 'function';
+
 let getCurrentTime: () => number;
 const hasPerformanceNow =
   typeof ɵglobal.performance === 'object' &&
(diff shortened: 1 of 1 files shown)


libs/cdk/render-strategies/src/lib/concurrent-strategies.ts
```diff
@@ -126,6 +169,39 @@ function scheduleOnQueue<T>(
       )
     );
 }
+function scheduleOnPostTaskQueue<T>(
+  work: (...args: any[]) => void,
```
Laurens-makel, line 173: Just wondering,

I've noticed the usage on `any` in many libraries, whilst Sonarqube scans recommend to replace `any` with `unknown` where possible. Sometimes it causes too strict typing and `any` could be preferred.

When I have a first look at this code, it seems that `any` could be replaced here with `unknown`. I'm not saying it's wrong however I wonder how serious this rule should be taken.

What is your opinion on this?

hoebbelsB: agree, `unknown` is better than `any`. This would in best case be done as part of this effort: https://github.com/rx-angular/rx-angular/issues/976 :)

PR #1673, opened 2024-01-07
refactor(cdk): drop unused scheduler APIs

edbzn:
Clean-up unused scheduler APIs, only `forceFramerate`, `scheduleCallback`, and `cancelCallback` are effectively used.

### DIFF
--- a/libs/cdk/internals/scheduler/src/lib/scheduler.ts
+++ b/libs/cdk/internals/scheduler/src/lib/scheduler.ts
@@ -51,9 +51,6 @@ const timerQueue = [];
 // Incrementing id counter. Used to maintain insertion order.
 let taskIdCounter = 1;
 
-// Pausing the scheduler is useful for debugging.
-let isSchedulerPaused = false;
-
 let currentTask: ReactSchedulerTask = null;
 let currentPriorityLevel = PriorityLevel.NormalPriority;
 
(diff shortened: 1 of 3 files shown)


libs/cdk/internals/scheduler/src/lib/scheduler.ts
```diff
@@ -71,18 +67,10 @@ const clearTimeout = ɵglobal.clearTimeout;
...
 
-const isInputPending =
-  typeof ɵglobal.navigator !== 'undefined' &&
-  ɵglobal.navigator.scheduling !== undefined &&
-  ɵglobal.navigator.scheduling.isInputPending !== undefined
-    ? ɵglobal.navigator.scheduling.isInputPending.bind(
-        ɵglobal.navigator.scheduling
-      )
-    : null;
-
```
hoebbelsB, line 82: this is actually something we want to keep.

libs/cdk/internals/scheduler/src/lib/scheduler.ts
```diff
@@ -364,33 +289,13 @@ function scheduleCallback(
...
-  isSchedulerPaused = true;
-}
-
-function continueExecution() {
-  isSchedulerPaused = false;
-  if (!isHostCallbackScheduled && !isPerformingWork) {
-    isHostCallbackScheduled = true;
-    requestHostCallback(flushWork);
-  }
-}
```
hoebbelsB, line 377: same goes for them, I guess they are connected

hoebbelsB (changes requested):
hey @edbzn could you restore the things we want to keep and rebase?

libs/cdk/internals/scheduler/src/lib/scheduler.ts
edbzn: Do we want to keep something not used? Why?

libs/cdk/internals/scheduler/src/lib/scheduler.ts
edbzn: Same question.

libs/cdk/internals/scheduler/src/lib/scheduler.ts
hoebbelsB: because it is part of the roadmap to test this API :D

edbzn: Then we could introduce it when we need it.

hoebbelsB: okay, fine by me

hoebbelsB:
> hey @edbzn could you restore the things we want to keep and rebase?

can u just rebase then? :D

PR #1674, opened 2024-01-09
fix(cdk): use `globalThis` instead of `ɵglobal` in zone-config

### LINKED ISSUE #1670: Importing `@rx-angular/cdk/zone-configurations` produces bundles with 2 imports of @angular/core which results to bundle size jump
# Description

Importing `@rx-angular/cdk/zone-configurations` produces bundles with 2 imports of @angular/core which results to bundle size jump. It's now both imported in `main` and `polyfills`

Before: 

image

After:

image

Just follow the instructions here: https://www.rx-angular.io/docs/cdk/zone-configurations/how-to-setup-zone-flags

# Environment

Angular v17
RxAngular v17

edbzn:
Resolves #1670

### DIFF
--- a/libs/cdk/zone-configurations/src/lib/zone-config.ts
+++ b/libs/cdk/zone-configurations/src/lib/zone-config.ts
@@ -1,4 +1,3 @@
-import { ɵglobal } from '@angular/core';
 import {
   RxZoneFlagsHelperFunctions,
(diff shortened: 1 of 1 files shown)


eneajaho (approved):

hoebbelsB (approved):
lovely, thanks

Issue #1647, opened 2023-11-18
ISR Angular 17 not working 


SkyZeroZx:
Today update my aplication to Angular 17 with ISR and use currently Angular SSR ( Previous Angular Universal ) , the caching of ISR not work and not show the comment in the HTML (by default use in memory or custom ) or register in Redis ( i use a custom handler) 

![image](https://github.com/rx-angular/rx-angular/assets/73321943/26b13439-f36f-4a5a-8395-fac2045bac95)

My server.ts file , and the end of file add the ISR , but the documentation currently not show the new configurations of Angular SSR in Angular 17 , possible the error is for **commonEngine** replace the previous use with Angular is different.

 `
export function app(): express.Express {
	const server = express();
	const distFolder = join(process.cwd(), 'dist/apps/MY_FOLDER/browser');
	const indexHtml = existsSync(join(distFolder, 'index.original.html'))
		? join(distFolder, 'index.original.html')
		: join(distFolder, 'index.html');

	const redisCacheHandler = new RedisCacheHandler({
		host: env['CACHE_HOST'] || '',
		name: env['CACHE_NAME'] || '',
		port: Number(env['CACHE_PORT']),
		ttl: Number(env['CACHE_TTL']),
		username: env['CACHE_USERNAME'] || '',
		password: env['CACHE_PASSWORD'] || ''
	});

	const isr = new ISRHandler({
		indexHtml,
		cache: redisCacheHandler,
		buildId: environment.buildTimestamp.toString(),
		invalidateSecretToken: env['CACHE_REVALIDATE_SECRET_TOKEN'] || '',
		enableLogging: isDevMode()
	});

	const commonEngine = new CommonEngine();

	server.set('view engine', 'html');
	server.set('views', distFolder);

	// Example Express Rest API endpoints
	// server.get('/api/**', (req, res) => { });
	// Serve static files from /browser
	server.get(
		'*.*',
		express.static(distFolder, {
			maxAge: '1y'
		})
	);

	// All regular routes use the Angular engine
	server.get('*', (req, res, next) => {
		const { protocol, originalUrl, baseUrl, headers } = req;
 
		commonEngine
			.render({
				bootstrap,
				documentFilePath: indexHtml,
				url: `${protocol}://${headers.host}${originalUrl}`,
				publicPath: distFolder,
				providers: [{ provide: APP_BASE_HREF, useValue: baseUrl }]
			})
			.then((html) => res.send(html))
			.catch((err) => next(err));
[...]

PR #1680, opened 2024-01-23
feat: add support for application builder and v17 in ISR

eneajaho:
Closes https://github.com/rx-angular/rx-angular/issues/1647


libs/isr/package.json
```diff
@@ -6,6 +6,7 @@
   "peerDependencies": {
     "@angular/common": "^17.0.0",
     "@angular/core": "^17.0.0",
+    "@angular/ssr": "^17.0.0",
     "@types/node": "^18.0.0 || ^20.0.0",
```
LayZeeDK, line 10: Update the lockfile
https://github.com/rx-angular/rx-angular/actions/runs/7619336794/job/20861381969#step:4:166

LayZeeDK (changes requested):
Request change

edbzn (commented):
Nice!

What about e2e testing the `ssr-isr` app with cypress? Just to make sure ISR runs successfully in the CI. Idealy, inlined in the project like we have for the `ssr` app.

I would also suggest squashing the commits at the end.

edbzn (approved):

Issue #1689, opened 2024-02-10
Issue with RxVirtualScrollViewport: Position of Data(item) Changes on Navigation


shiv8652:
**Problem Description:**
I'm encountering an issue with the rx-virtual-scroll-viewport package in my Angular application. Here's a simplified version of the relevant code:
```
<div rxVirtualScrollElement style="min-height: 100%">
    <!-- Content Before -->
    <div style="height: 300px; border: 1px solid red; margin-bottom: 10px; margin-top: 10px;" class="mr-x">
        Content Before Content Before Content Before
    </div>
    <div style="position: relative;" class="mr-x">
        <rx-virtual-scroll-viewport (scrolledIndexChange)="rxaScrolledIndex$.next($event)" autosize
            [resizeObserverConfig]="{  extractSize: extractSize }" [runwayItemsOpposite]="runwayItemsOpposite"
            [runwayItems]="runwayItems" class="viewport">
            <div class="item" *rxVirtualFor="let item of data$; let index$ = index$; let i = index;
                trackBy: trackItem; renderCallback: rendered; strategy: $any(strategy$)"
                class="item" (click)="GoToDetails(item,index$,i);">
                <div class="content">
                    <p style="font-weight: 700;"><span>{{ i}}. </span><span>{{ item.title }}</span></p>
                    <p>{{item.info}}</p>
                </div>
            </div>
        </rx-virtual-scroll-viewport>
    </div>
</div>
```
**Problem Details:**
- I have a list of data(item), upon clicking on an item from the list, the application navigates to a detail page.
-  When returning to the list page, the position of the data within the rx-virtual-scroll-viewport changes.
- Similarly, navigating between (two) tabs leads to the same issue of data position changing.

**Requested Assistance:**
I'm seeking guidance on how to maintain the position of the data within the rx-virtual-scroll-viewport when navigating between pages or tabs. Any suggestions or insights into resolving this issue would be greatly appreciated.

Thank you in advance for your assistance!

hoebbelsB:
hey @shiv8652 i'm working on a feature to set an initial scroll position. If you store the last scroll position you can set it as an input to easily restore the position you had before.

But anyway, for the autosize strategy you'll always be able to set the position to a specific view index. As of now, you should be able to do that by using the `VirtualScrollViewport#scrollToIndex` method.

[...]

Issue #1679, opened 2024-01-22
RxVirtualScroll do not fill empty space when an item in the viewport changes position (sorting the array)


Lougnar:
Hi, 

# Description

When the array provided to *rxVirtualFor is sorted, space left by items in viewport that changed positions are not filled. If the user scroll down to push the empty space out of the viewport and scroll up, the space disapear and the items are displayed correctly

# Steps to Reproduce the Issue

I made a minimal repoduction of the bug: 
https://stackblitz.com/edit/stackblitz-starters-izzuwy?file=src%2Fmain.ts

Click on the button below the virtual scroll, the item 2 will be repositionned at the end of the array. You will see a space between item 1 and item 3.

# Environment

Tested on "@rx-angular/template": "16.1.1", + Angular 16 and "@rx-angular/template": "17.0.1" and Angular 17

# Note

I haven't found a workaround at the moment

Best regards,
Lougnar

hoebbelsB:
thanks for the issue and sorry for the delay. Going to investigate this week :crossed_fingers:

Issue #1686, opened 2024-02-02
How RxVirtualScroll supposed to work with drag and drop?


volser:
# Problem Solved By The Feature

RxVirtualScroll is created to render huge lists, very often such lists are draggable.
The most popular library for dnd in angular is Cdk dnd.
Cdk dnd does not work if while dragging items were removed from DOM, but it's how RxVirtualScroll working
How to make it works together?

# Solution

Cdk virtual scroll has different workaround solutions, like `appendOnly`.
Better to implement dnd support out of the box, or at least allow workaround solution, cuz now it looks like impossible. 

# Alternatives Considered

# Additional Context

hoebbelsB:
hey @volser thanks for the issue, i'm looking into it!

hoebbelsB:
hello @volser we've published a new version which also ships the new `appendOnly` mode. Could you please see if this solves your issue?

hoebbelsB:
@volser did it solve your issue or is there further investigation needed?

volser:
> @volser did it solve your issue or is there further investigation needed?

thank you for your fix!
did not try yet, will return to my PoC soon and let you know

PR #1693, opened 2024-02-16
feat(template): introduce appendOnly mode, initialScrollIndex & bugfixes

hoebbelsB:
# Description

Sorry for delivering multiple things within a single PR...

This PR introduces the following new features:

* appendOnly mode
* initialScrollIndex input

as well as a bugfix for the virtual list template manager that sometimes produced bad diffs and a fix for a viewglitch caused by a mispositioned sentinel.

## AppendOnly

There is a new input `appendOnly: boolean` on the scroll strategies. Enabling this mode causes the rendered range to only increase - in any direction. See https://material.angular.io/cdk/scrolling/overview#append-only-mode

I hope this addresses https://github.com/rx-angular/rx-angular/issues/1686

##  InitialScrollIndex

There is a new input `initialScrollIndex: number` on the `RxVirtualScrollViewportComponent` which lets users define an initial index to scroll to. fixes https://github.com/rx-angular/rx-angular/issues/1689

> the virtual-scroll-viewport now accepts an index as input which sets the first view to be visible to the user. The viewport waits for the data to arrive and scrolls to the given index immediately.

## Bugfix

Fixes https://github.com/rx-angular/rx-angular/issues/1679

I've recorded the following demo ( I did move the logic from the demo stackblitz to our demo repo for this purpose )

fix-bug

### DIFF
--- a/libs/template/experimental/virtual-scrolling/src/lib/util.ts
+++ b/libs/template/experimental/virtual-scrolling/src/lib/util.ts
@@ -6,6 +6,10 @@ import {
 } from '@rx-angular/cdk/zone-less/browser';
 import { from, Observable } from 'rxjs';
 
+export function toBoolean(input: null | boolean | string | undefined): boolean {
+  return input != null && `${input}` !== 'false';
+}
+
 export function unpatchedAnimationFrameTick(): Observable<void> {
   return new Observable<void>((observer) => {
     const tick = requestAnimationFrame(() => {
--- a/libs/template/experimental/virtual-scrolling/src/lib/virtual-for.directive.ts
+++ b/libs/template/experimental/virtual-scrolling/src/lib/virtual-for.directive.ts
@@ -54,6 +54,7 @@ import {
   RxVirtualScrollStrategy,
   RxVirtualViewRepeater,
 } from './model';
+import { unpatchedMicroTask } from './util';
 import {
   createVirtualListTemplateManager,
   RxVirtualListTemplateManager,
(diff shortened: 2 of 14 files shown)


libs/template/experimental/virtual-scrolling/src/lib/virtual-for.directive.ts
```diff
@@ -609,11 +608,14 @@ export class RxVirtualFor<T, U extends NgIterable<T> = NgIterable<T>>
       updateViewContext: this.updateViewContext.bind(this),
       templateCacheSize: this.templateCacheSize,
     });
-    this.render()
-      .pipe(takeUntil(this._destroy$))
-      .subscribe((v) => {
-        this._renderCallback?.next(v as U);
-      });
+    // let the scroll strategy initialize before
+    Promise.resolve().then(() => {
```
edbzn, line 612: What about using the primitive operation `queueMicrotask` instead?

hoebbelsB: both are patched by zone and I'm using the `Promise` from `cdk/zone-less/browser`. Of course I could also use the unpatched `queueMicrotask`. But is there any difference besides it looks a little bit better?

edbzn: No big difference, it takes less memory as there's no object to hold, but nothing noticeable I guess. However it's easier to read.

libs/template/experimental/virtual-scrolling/src/lib/util.ts
```diff
@@ -6,6 +6,10 @@ import {
 } from '@rx-angular/cdk/zone-less/browser';
 import { from, Observable } from 'rxjs';
 
+export function toBoolean(input: null | boolean | string | undefined): boolean {
+  return input != null && `${input}` !== 'false';
+}
```
edbzn, line 11: Angular exposes a [`booleanAttribute`](https://angular.io/api/core/booleanAttribute), we could use it directly.

PR #1694, opened 2024-02-16
fix(template): use constructor injection for templateRef

### LINKED ISSUE #1616: Broken template typings
# Description

I think [this commit](https://github.com/rx-angular/rx-angular/commit/02b62fb5da76cb3e765a35848c4961a91f96a64a) broke our template type support. Before, the types were properly inferred by whatever observable I've provided. I think it has something todo with the way how the `TemplateRef` is injected into the component. From what I know it has to be part of the constructor.

@edbzn @eneajaho any idea?

image

hoebbelsB:
# Description

Multiple users reported the issue that their IDE cannot infer the typings from our structural directives due to the `inject` method. Using the constructor based injection solves that issue. fixes #1616


edbzn (approved):

kkachniarz220:
maybe this should be reported to angular team?

PR #1704, opened 2024-03-18
[State] feat - default transform function for select

### LINKED ISSUE #1683: [State] select default transform fn
```typescript
this._state.select(['foo', 'bar'], s => s) // { foo, bar }
```

To

```typescript
this._state.select(['foo', 'bar']) // { foo, bar }
```
Probably the most common way of use `select(keys: K[])` is to transform to a simple object with same data structure as in state.

# Solution
Make default mapping in `select(keys: K[])`

AdrianRomanski:
Default transform function for selecting state

fixes #1683

### DIFF
--- a/libs/state/selections/src/lib/operators/select.ts
+++ b/libs/state/selections/src/lib/operators/select.ts
@@ -242,9 +242,9 @@ export function select<T extends Record<string, unknown>>(
           opOrMapFn[2] as KeyCompareMap<{ [P in keyof T]: (T & object)[P] }>
         ),
         stateful(
-          map((pickSlice) => {
-            // @ts-ignore
+          map((pickSlice: PickSlice<T & Object, keyof T>) => {
             return opOrMapFn[0].reduce(
+              // @ts-ignore
               (acc: { [key: keyof T]: T }, key: keyof T) => {
                 // @ts-ignore
                 acc[key] = pickSlice[key as keyof T];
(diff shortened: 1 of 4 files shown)


AdrianRomanski:
I need help with getting rid of @ts-ignore (generics puzzle)

AdrianRomanski:
@hoebbelsB @eneajaho @edbzn 

Can you guys help me with getting rid of ts-ignore (generics puzzle), and check the code please :D

libs/state/selections/src/lib/operators/select.ts
```diff
@@ -234,6 +235,26 @@ export function select<T extends Record<string, unknown>>(
...
+              // @ts-ignore
+              (acc: { [key: keyof T]: T }, key: keyof T) => {
+                // @ts-ignore
+                acc[key] = pickSlice[key as keyof T];
+                return acc;
+              },
+              {}
+            );
+          })
+        )
```
hoebbelsB, line 256: question: is this actually needed?

Couldn't it be just:

```ts
return state$.pipe(
        selectSlice(
          opOrMapFn[0],
          opOrMapFn[2]
        )
      );
```
???

At least your test passes that way :)

libs/state/selections/src/lib/operators/select.ts
```diff
@@ -222,7 +223,7 @@ export function select<T extends Record<string, unknown>>(
       return state$.pipe(
         selectSlice<T & object, keyof T>(
           opOrMapFn[0] as (keyof T)[],
-          opOrMapFn[2]
+          opOrMapFn[2] as KeyCompareMap<{ [P in keyof T]: (T & object)[P] }>
         ),
```
hoebbelsB, line 227: just a thought, wouldn't it be possible to insert that logic into this branch? we only need to omit calling the `map` function when it's not given

AdrianRomanski: @hoebbelsB sure good idea

hoebbelsB (approved):

PR #1710, opened 2024-04-04
refactor(template): deprecate parent flag

hoebbelsB:
# Description

We had this discussion for a long time and now we can finally start deprecating the `parent` flag in the render strategies and as input for our directives.

With the new signal queries, the parent flag is essentially obsolete, it will only cause unnecessary change detection cycles.

I have confirmed this with new tests and a new demo.

- [x] deprecate parent flag
- [x] write tests
- [x] adjust docs
- [ ] think about final removal

## Further Steps

We need to discuss when to finally drop the flag, how to communicate that change, provide update information in the docs.


eneajaho (approved):
LGTM

PR #1711, opened 2024-04-06
feat: rx state as read only

### LINKED ISSUE #1703: rxState `asReadOnly`

right now, if you have e.g. a service with a global state:

```ts
class Service {
  state = rxState();
}
```

everyone using `Service` also has full access to its state. 
This is a problem as you don't want others to write into your state.

If you want to expose only slices, you have to provide a public property per slice:

```ts
class Service {
 private  state = rxState();

 slice$ = this.state.select('slice1');
 slice2$ = this.state.select('slice2');

}
```

# Solution

To have a more convenient way to share a read only access to your state, we could introduce a `asReadOnly` function on top of `RxState` which only exposes `select` & `get`

```ts

class Service {
 private  _state = rxState();

 state = this.state.asReadonly();

}
```

AdrianRomanski:
Fixes #1703

### DIFF
--- a/apps/docs/docs/state/api/rx-state.md
+++ b/apps/docs/docs/state/api/rx-state.md
@@ -67,7 +67,7 @@ Use the `$` property if you want to read the state without having applied [state
 
 ## asReadOnly
 
-##### typeof: Pick&#60;RxState&#60;State&#62, 'get' | 'select' | 'computed' | 'signal'&#62
+##### typeof: Pick&#60;RxState&#60;State&#62;, 'get' | 'select' | 'computed' | 'signal'&#62;
 
 Return RxState in ReadOnly mode that is exposing
 get(), select(), computed() and signal() methods.


libs/state/src/lib/rx-state.service.ts
```diff
@@ -99,6 +99,25 @@ export class RxState<State extends object>
...
+   * @example
+   * ```typescript
+   * const readOnlyState = state.asReadOnly();
+   * const getNum = state.get('num');
+   * const selectNum$ = state.select('num');
+   * ```
+   *
+   * @return Pick<RxState<State>, 'get' | 'select'>
+   */
+  asReadOnly(): Pick<RxState<State>, 'get' | 'select'> {
```
hoebbelsB, line 117: we of course also want to have the `signal` & `computed` APIs to be exposed.

libs/state/src/lib/rx-state.service.ts
```diff
@@ -99,6 +99,25 @@ export class RxState<State extends object>
...
+    return this as Pick<RxState<State>, 'get' | 'select'>;
```
hoebbelsB, line 118: with this approach we only return a new type, the disallowed methods are still available though. It's maybe safer to return a new object, e.g.

```ts
return { get: this.get, select: this.select, computed: this.computed, signal: this.signal }
```

libs/state/spec/rx-state.service.spec.ts
```diff
@@ -139,25 +139,33 @@ describe('RxStateService', () => {
     it('should return undefined as initial value', () => {
```
hoebbelsB, line 139: can u also add dedicated tests for the `readOnly` that check if an error is thrown when trying to write to the readonly state? The apis shouldnt be accessible anymore

hoebbelsB (changes requested):
hey @AdrianRomanski thank u for the PR <3 very much appreciated. I left some comments with thoughts on the current implementation.

And by any chance, would it be possible to also provide docs for this new behavior? I can assist you with getting familiar with the docs app if you like

AdrianRomanski:
@hoebbelsB Thanks for the comments :) i will fix them soon. Of course i can provide dosc, will try on my own if got any problem i'll shoot a question :D

apps/docs/docs/state/api/rx-state.md
```diff
@@ -65,6 +65,33 @@ Use the `$` property if you want to read the state without having applied [state
 
 ---
 
+## asReadOnly
+
+##### typeof: Pick<RxState<State>, 'get' | 'select' | 'computed' | 'signal'>
+
+Return RxState in ReadOnly mode that is exposing
+get(), select(), computed() and signal() methods.
+This can be helpful when you don't want other's to write in your state.
```
hoebbelsB, line 74:
```suggestion
This can be helpful when you don't want others to write in your state.
```

libs/state/src/lib/rx-state.service.ts
```diff
@@ -99,6 +101,31 @@ export class RxState<State extends object>
     this.subscription.unsubscribe();
   }
 
+  /**
+   * @description
+   *
+   * Return RxState in ReadOnly mode exposing only methods for reading state
+   * get(), select(), computed() and signal() methods.
+   * This can be helpful when you don't want other's to write in your state.
```
hoebbelsB, line 109:
```suggestion
   * This can be helpful when you don't want others to write in your state.
```

hoebbelsB (approved):
Thank u @AdrianRomanski for the contribution ❤️ just left some comments with fixes for a typo. Looks good otherwise

AdrianRomanski:
@hoebbelsB fixed the typo and docs building ⌨️

PR #1714, opened 2024-04-21
feat(template): implementation for signals support in template package

eneajaho:
- Copied all observables tests and ported to use signals 
- Having issues with context for signals as they are different from observables which have different states 
  -  Should we throw an error here to tell the user that context is not available with signals? 
- **rxIf** now supports having a signal value
- **rxLet** now supports having a signal value
-  **rxFor** now supports having a signal value
- **rxVirtualFor** now supports having a signal value

```html
<div *rxIf="someSignalValue"> 
  Hello world!
</div>

<div *rxLet="someSignalValue; let v"> 
  Hello world! {{ v }}
</div>

<ul> 
  <li *rxFor="let movie of moviesSignal">{{movie}}</li>
</ul>
```


libs/template/if/src/lib/if.directive.ts
```diff
@@ -486,7 +491,20 @@ export class RxIf<T = unknown>
...
+  private currentRxIfSignal = signal<Signal<T> | undefined>(undefined);
+
+  constructor(private readonly templateRef: TemplateRef<RxIfViewContext<T>>) {
+    effect(() => {
+      if (this.currentRxIfSignal() !== undefined) {
+        const rxIfValue = this.currentRxIfSignal()();
+        untracked(() => this.templateNotifier.next(rxIfValue));
+      }
+    });
+  }
```
hoebbelsB, line 507: isn't rxIfValue of type `Signal` in this case? Shouldn't it be `currentRxIfSignal()()?`

libs/template/if/src/lib/if.directive.ts
```diff
@@ -549,7 +567,15 @@ export class RxIf<T = unknown>
       this.templateManager.addTemplateRef(RxIfTemplateNames.error, this.error);
     }
     if (changes.rxIf) {
-      this.templateNotifier.next(this.rxIf);
+      if (isSignal(this.rxIf)) {
+        // add to currentRxIfSignal so we can listen to changes
+        this.currentRxIfSignal.set(this.rxIf);
+
+        // send value synchronously
+        this.templateNotifier.next(this.rxIf());
```
hoebbelsB, line 575: won't this result in two emissions of `templateNotifier`? Sync + promise coalesced from the effect?

eneajaho: Yes, it will emit twice, but I thought that we are already using distinctUntilChanged() in the templateNotifier, so the same value doesn't retrigger change detection.

libs/template/if/src/lib/if.directive.ts
eneajaho: That's how it is already 🤔

hoebbelsB: lol, I'm blind

libs/template/if/src/lib/if.directive.ts
hoebbelsB: why triggering it in the first place?

eneajaho: Don't we want to keep the synchronous behavior? Do we want to always have it scheduled based on the effect scheduler?

hoebbelsB: It would make sense to me to do so, yes

libs/template/let/src/lib/let.directive.ts
```diff
@@ -573,7 +591,14 @@ export class RxLet<U> implements OnInit, OnDestroy, OnChanges {
     }
 
     if (changes.rxLet) {
-      this.observablesHandler.next(this.rxLet);
+      if (isSignal(this.rxLet)) {
+        this.currentRxLetSignal.set(this.rxLet);
+
+        // set the initial value synchronously
+        this.observablesHandler.next(this.rxLet());
```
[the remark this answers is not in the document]
eneajaho: In a zoneless world, that should work when using something like: 

```ts
await fixture.whenStable();
```

apps/docs/docs/template/api/push-pipe.md
```diff
@@ -29,17 +29,15 @@ components and does not work in zone-less mode.
 _Example_
 
 ```html
-<hero-search [term]="searchTerm$ | push"> </hero-search>
-<hero-list-component [heroes]="heroes$ | push"> </hero-list-component>
+<hero-search [term]="searchTerm$ | push"> </hero-search> <hero-list-component [heroes]="heroes$ | push"> </hero-list-component>
```
eneajaho, line 32: Can we revert this formatting?

apps/docs/docs/template/api/push-pipe.md
```diff
@@ -29,17 +29,15 @@ components and does not work in zone-less mode.
...
 ```
 
 The rendering behavior can be configured per RxPush instance using the strategy parameter.
 
 _Example_
 
 ```html
-<hero-search [term]="searchTerm$ | push: 'global'"> </hero-search>
-<hero-list-component [heroes]="heroes$ | push: 'global'"> </hero-list-component>
+<hero-search [term]="searchTerm$ | push: 'global'"> </hero-search> <hero-list-component [heroes]="heroes$ | push: 'global'"> </hero-list-component>
```
eneajaho, line 40: Revert formatting?

apps/docs/docs/template/api/rx-let-directive.mdx
```diff
@@ -15,17 +18,71 @@ To name a few:
...
+<Tabs>
 
-The ngIf hack looks like this:
+<TabItem value="ngIf" label="ngIf hack">
 
 ```html
-<ng-container *ngIf="observableNumber$ | async as n">
+@if (number$ | async; as n) {
+<app-number [number]="n"></app-number>
+<app-number-special [number]="n"></app-number-special>
```
eneajaho, line 28:
```suggestion
  <app-number [number]="n" />
  <app-number-special [number]="n" />
```

apps/docs/docs/template/api/rx-let-directive.mdx
```diff
@@ -15,17 +18,71 @@ To name a few:
...
+
+</TabItem>
+
+<TabItem value="signal" label="rxLet usage with signals">
+
+```html title="src/counter.component.html"
+<ng-container *rxLet="number; let n">
+  <app-number [number]="n"></app-number>
+  <app-number-special [number]="n"></app-number-special>
+</ng-container>
```
eneajaho, line 40:
```suggestion
<ng-container *rxLet="number; let n">
  <app-number [number]="n" />
  <app-number-special [number]="n" />
</ng-container>
```

apps/docs/docs/template/api/rx-let-directive.mdx
```diff
@@ -139,18 +193,66 @@ export class AnyComponent {}
...
+
+</TabItem>
+
+<TabItem value="obs" label="Usage with observables">
+
+```html title="src/counter.component.html"
+<ng-container *rxLet="number$; let n">
+  <app-number [number]="n"></app-number>
+  <app-number-special [number]="n"></app-number-special>
+</ng-container>
```
eneajaho, line 235:
```suggestion
<ng-container *rxLet="number$; let n">
  <app-number [number]="n" />
  <app-number-special [number]="n" />
</ng-container>
```

Issue #1687, opened 2024-02-08
ISR cache invalidation error


moddi3:
# Description 

I'm using angular 17 with the new release of @rx-angular/isr. Everything is working well except on-demand invalidation, when I trigger localhost:4000/api/invalidation I'm getting this error - `ERROR RuntimeError: NG04002: Cannot match any routes. URL Segment: 'api/invalidate'`
Either you trigger it via the postman or via the button - the result is the same.
It's also noteworthy that the http request answers 200.

@eneajaho help, please

# Steps to Reproduce the Issue
https://codesandbox.io/p/devbox/divine-pine-mq2vq7

when you click the **invalidate** button you will see error in console
<img width="818" alt="image" src="https://github.com/rx-angular/rx-angular/assets/11387313/208f2b88-4ac4-4183-a263-cdb1d838f3c6">

# Environment

```
Angular CLI: 17.1.2
Node: 18.19.0
OS: macOS Sonoma / codesandbox

Angular: 17.1.2
Ivy Workspace: Yes

Package                           Version
-----------------------------------------------------------
```
<img width="551" alt="image" src="https://github.com/rx-angular/rx-angular/assets/11387313/a01dfe54-b79c-47f1-9c00-2868971fc2f0">

moddi3:
> Hello @moddi3 I just tried with a simple example https://github.com/rx-angular/rx-angular/pull/1688/files and everything worked fine.
> 
[...]

PR #1719, opened 2024-05-02
fix(isr): resolved the ISR cache invalidation issue

jginorio:
Resolved the ISR cache invalidation issue by modifying the URL within the commonEngine inside the renderURL function. #1687


yarn.lock
eneajaho, line 1: Can you revert this file change please?

eneajaho:
Hi @jginorio 
Thanks for the PR, but we're going with this other one instead https://github.com/rx-angular/rx-angular/pull/1735 
Closing as completed in https://github.com/rx-angular/rx-angular/pull/1735

PR #1730, opened 2024-05-28
feat: upgrade to ng 18

hoebbelsB:


libs/template/package.json
```diff
@@ -43,9 +43,9 @@
     "url": "https://github.com/rx-angular/rx-angular.git"
   },
   "peerDependencies": {
-    "@angular/core": "^17.0.0",
+    "@angular/core": "^18.0.0",
     "rxjs": "^6.5.3 || ^7.4.0",
```
edbzn, line 47: I can't find the information on the website but I'm wondering if Angular v18 changed its compat range with RxJS?

edbzn (changes requested):
The commit message should include the breaking change in order to follow semver. The rest looks good!

Edit: The new dependency `@angular/build` seems to be missing.

libs/template/package.json
hoebbelsB: doesn't look like it https://github.com/angular/angular/blob/main/packages/core/package.json

JeanMeche:
`@angular-devkit/build-angular` has `@angular/build` as dependency, so it shouldn't be an issue, right ?

Issue #1691, opened 2024-02-13
preview images for rx-angular.io are not working


hoebbelsB:
# Description

the meta tags for https://www.rx-angular.io/ are not properly configured so that no link preview is possible

hoebbelsB:
@LayZeeDK please unassign if you don't like :D I thought you have the greatest knowledge about the docusaurus setup

AdrianRomanski:
@hoebbelsB I'll do it

PR #1732, opened 2024-05-30
fix: meta tags for docusaurus

AdrianRomanski:
resolves https://github.com/rx-angular/rx-angular/issues/1691

meta tags preview

Zrzut ekranu 2024-05-30 o 11 48 41

### DIFF
--- a/apps/docs/docusaurus.config.js
+++ b/apps/docs/docusaurus.config.js
@@ -72,6 +72,36 @@ module.exports = {
   themeConfig:
     /** @type {import('@docusaurus/preset-classic').ThemeConfig} */
     ({
+      metadata: [
+        {
+          name: 'description',
+          content:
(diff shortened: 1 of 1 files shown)


apps/docs/docusaurus.config.js
```diff
@@ -72,6 +72,36 @@ module.exports = {
   themeConfig:
     /** @type {import('@docusaurus/preset-classic').ThemeConfig} */
     ({
+      metadata: [
+        {
+          name: 'description',
+          content:
+            'Official documentation for RxAngular, the ultimate reactive programming library for Angular.',
+        },
```
hoebbelsB, line 80: where did u get that slogan from? :D

AdrianRomanski: You don't like It?:D Actually i had no idea what to put there, we can change It

hoebbelsB (approved):
thank u <3

Issue #1738, opened 2024-06-22
ISR: Allow for redirect response to occur during rendering


Nabelops:
# Problem Solved By The Feature

I am implementing a redesign for a site.  With this, we are transitioning from a SPA to SSR. Given the importance SEO, we would like to provide a 301 response to old urls that have moved as well as 404 to our wildcard catchall route.  I was looking at implementing a solution discussed at https://dev.to/cubesoft/mastering-angular-ssr-enabling-custom-http-responses-404-301-55pc.

This implementation currently works but results in a server side error `Error [ERR_HTTP_HEADERS_SENT]: Cannot set headers after they are sent to the client`

# Solution

Provide logic in the ISRHandler render to check if the response was already set before `res.send()`.  In the tutorial above, it checks `res. headersSent` to determine if the response should be sent.

# Alternatives Considered

I'm not entirely convinced that this logic doesn't all belong in the `server.ts` vs in Angular code.  With React implementing server side react, those lines are blurred.  The nice thing about this approach is that you can handle them differently based on if the url is hit directly and rendered by the server or if it is hit once the client side code takes over.

E.g., with the 404, if it is a bad router link in the code, the 404 page can be shown but if it is a bad back link, the server can respond with a 404 code.

# Additional Context

The ability to reproduce this that deviates from the tutorial linked above is how to provide the injection token.  With ISR, you provide it here

```typescript
async (req, res, next) =>
  await isr.render(req, res, next, {
    providers: [
      {
        provide: RESPONSE,
        useValue: res,
      },
    ],
  })
)
```

PR #1739, opened 2024-06-24
fix(isr): handle response sent in render

Nabelops:
The main reason for this change is described in the issue https://github.com/rx-angular/rx-angular/issues/1738.  With this change, render exits if the `renderUrl` already handles the server response.


eneajaho (approved):

eneajaho:
Thanks @Nabelops

PR #1741, opened 2024-06-26
fix(isr): refactor cache regeneration function

### LINKED ISSUE #1740: ISR: Background regeneration used in FileSystemHandler ignores custom config options and variants
# Description

Currently, when using the `FileSystemCacheHandler`, the background regeneration task ignores the `skipCachingOnHttpError` value and the created variants when regenerating the cache from a request.

This means that even when `skipCachingOnHttpError` is set to false, pages with HTTP errors will not be regenerated in the background after expiry from the first cached file, which causes pages with HTTP errors to be cached indefinitely regardless of the revalidation time.

Also, the background regeneration does not refresh the cache for variant pages. This means that all variant pages will also be cached indefinitely, because it will always try and refresh the cache for the non-variant page when requested.

1. Start a new Angular 17 project and install the corret rx-angular/isr package and dependencies
2. Create routes with `revalidate: 10` as data value
3. Create an api request that will throw an error to the Angular client to test
4. Create the `FileSystemCacheHandler` custom cache handler.
5. Create a new ISRHandler instance with the `FileSystemCacheHandler` as custom handler, and add `skipCachingOnHttpError = false` and a `variant` to the parameters of the ISRHandler (the variant should be enabled when you visit the page, so just return true)
6. Build the project and generate the first cached HTML file by visiting the homepage of your angular app
7. After the revalidation time has ended, refresh the page. If you have any API errors, the page will not be regenerated. Also, if you are using a variant page, that page will not be regenerated; only the page without variant will be regenerated.

# Environment

This will probably will happen regardless of what version you are using, but I used this:

"@angular/core": "^17.3.1",
"@rx-angular/isr": "^17.1.0"

# Tasks to Resolve This

Refactor the code that handles regeneration for the `FileSystemCacheHandler` to make sure the correct config options are being set and that the variants URL's are being used to regenerate pages in the background if the variant page is requested.

SanderVO-BC:
Refactor regeneration function for FileSystemCacheHandler so it supports variants and skipCachingOnHttpError config parameter. This fixes #1740 .

### DIFF
--- a/libs/isr/server/src/cache-regeneration.ts
+++ b/libs/isr/server/src/cache-regeneration.ts
@@ -7,6 +7,7 @@ import {
 } from '@rx-angular/isr/models';
 import { Request, Response } from 'express';
 import { ISRLogger } from './isr-logger';
+import { getCacheKey, getVariant } from './utils/cache-utils';
 import { getRouteISRDataFromHTML } from './utils/get-isr-options';
 import { renderUrl } from './utils/render-url';
 
@@ -21,7 +22,7 @@ export class CacheRegeneration {
     public indexHtml: string,
     public commonEngine?: CommonEngine,
     public bootstrap?: CommonEngineRenderOptions['bootstrap'],
-    public browserDistFolder?: string
+    public browserDistFolder?: string,
   ) {}
 
   async regenerate(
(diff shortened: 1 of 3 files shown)


Dhan3777:
When will it get merge ? It has very crucial fix we are waiting for. Issue is with variant page is being generated but old page is getting serve only. Since it's not using cacheKey made using variant to regenerate the page

eneajaho (approved):
LGTM! CI isn't green, after we fix that, we can merge!

SanderVO-BC:
Hello @eneajaho,

Do you expect something else from me? I see that some of the errors in the pipeline include files I did not touch, so I'm not sure if these errors are branch specific.

SanderVO-BC:
@eneajaho 

Fixed!

eneajaho:
Released https://github.com/rx-angular/rx-angular/releases/tag/isr%4018.0.2

PR #1749, opened 2024-08-16
chore: reduce CI complexity. use nx cloud agents

hoebbelsB:
not sure about the `parallel` usage, but everything else should be just fine. Spent some time on tweaking the agents setup. We of course could opt-in to use those xtra-large images ... but i'm kind of hesitating to use so much power for our little lib :D

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -25,6 +25,10 @@ env:
   docs-path: dist/apps/docs
   NX_BRANCH: ${{ github.event.number || github.ref_name }}
 
+permissions:
+  actions: read
+  contents: read
+
 jobs:
   main:
     name: Main CI
@@ -59,7 +63,7 @@ jobs:
       - name: Run commands in parallel
         run: |
           npx nx-cloud record -- npx nx format:check
-          npx nx affected -t lint build test e2e component-test --parallel=4 --exclude=docs
+          npx nx affected -t lint build test component-test e2e --parallel=4 --exclude=docs
 
       # Upload coverage reports to Codecov
       - name: Upload coverage
@@ -74,27 +78,21 @@ jobs:
   build-docs:
     name: Build docs
     runs-on: ubuntu-latest
-    permissions:
-      actions: read
-      contents: read
 
     steps:
       - name: Checkout all commits
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
         with:
           fetch-depth: 0
       - name: Set up dependencies
         uses: ./.github/actions/setup
 
       - name: Build docs
-        env:
-          NX_CLOUD_DISTRIBUTED_EXECUTION: false
-          NX_DISTRIBUTED_TASK_EXECUTION: false
         run: yarn nx build docs
 
       - name: '[Merge] Upload docs'
         if: github.ref == 'refs/heads/main'
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
           name: ${{ env.docs-name }}
           path: ${{ env.docs-path }}
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -55,7 +55,7 @@ jobs:
         uses: nrwl/nx-set-shas@v4
 
       - name: Initialize the Nx Cloud distributed CI run
-        run: npx nx-cloud start-ci-run --distribute-on="./nx/workflows/dynamic-changesets.yml" --stop-agents-after="e2e"
+        run: npx nx-cloud start-ci-run --distribute-on="./nx/workflows/dynamic-changesets.yml" --stop-agents-after="e2e" --require-explicit-completion
 
       - name: Set up dependencies
         uses: ./.github/actions/setup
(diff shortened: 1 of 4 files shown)


hoebbelsB:
@eneajaho do u think it's working with the current `--parallel` setting? We also run e2e and component tests with that. Not sure what issues can be caused because of it

LayZeeDK:
> @eneajaho do u think it's working with the current `--parallel` setting? We also run e2e and component tests with that. Not sure what issues can be caused because of it

@hoebbelsB 

When using Nx Agents, it's recommended not to have a `--parallel` setting [as seen in the docs](https://nx.dev/ci/features/split-e2e-tasks#configure-automated-task-splitting-on-ci) and in [the workflow generator](https://github.com/nrwl/nx/blob/18e5535940a0aff86dc845f3714b07b94a0db8d8/packages/workspace/src/generators/ci-workflow/files/github/.github/workflows/__workflowFileName__.yml__tmpl__). It's also recommended to [use Nx Atomizer](https://nx.dev/ci/features/split-e2e-tasks), that is the `e2e-ci` tasks to distribute individual end-to-end tests. In this case, `e2e-ci` should also be passed to `--stop-agents-after`

.github/workflows/build-and-test.yml
```diff
@@ -81,53 +44,26 @@ jobs:
...
-            fi
-          done
-
-          exit 0
-
-      - name: Stop Nx Cloud DTE agents
-        if: ${{ always() }}
-        run: yarn nx-cloud stop-all-agents
+          npx nx-cloud record -- npx nx format:check
+          npx nx affected -t lint build test component-test e2e --parallel=4 --exclude=docs
```
LayZeeDK, line 66: suggestion: Remove the `--parallel` option as distribution is optimized by Nx Agents based on the activated changeset.

```suggestion
          npx nx affected -t lint build test component-test e2e --exclude=docs
```

LayZeeDK, line 66: suggestion: Try using the `e2e-ci` tasks. [Configuration docs for the Nx Cypress plugin](https://nx.dev/nx-api/cypress#nxcypress-configuration).

Specify `e2e-ci` both for `nx cloud start-ci-run --stop-agents-after` and `nx affected -t`.

```suggestion
          npx nx affected -t lint build test component-test e2e-ci --parallel=4 --exclude=docs
```

.github/workflows/build-and-test.yml
```diff
@@ -81,53 +44,26 @@ jobs:
...
           fetch-depth: 0
 
-      - name: Set up dependencies
-        uses: ./.github/actions/setup
+      - name: Derive appropriate SHAs for base and head for `nx affected` commands
+        uses: nrwl/nx-set-shas@v4
 
       - name: Initialize the Nx Cloud distributed CI run
-        run: yarn nx-cloud start-ci-run
+        run: npx nx-cloud start-ci-run --distribute-on="./nx/workflows/dynamic-changesets.yml" --stop-agents-after="e2e"
```
LayZeeDK, line 58: suggestion: Try using the `e2e-ci` tasks. [Configuration docs for the Nx Cypress plugin](https://nx.dev/nx-api/cypress#nxcypress-configuration).

Specify `e2e-ci` both for `nx cloud start-ci-run --stop-agents-after` and `nx affected -t`.

```suggestion
        run: npx nx-cloud start-ci-run --distribute-on="./nx/workflows/dynamic-changesets.yml" --stop-agents-after="e2e-ci"
```

LayZeeDK (commented):
Suggest changes.

.github/workflows/build-and-test.yml
hoebbelsB: thanks!

.github/workflows/build-and-test.yml
hoebbelsB: @LayZeeDK tried, didn't work. I don't want to mess around with it any longer :D if you have any other suggestion, i'm going to revert the e2e-ci changes

.github/workflows/build-and-test.yml
hoebbelsB: @LayZeeDK tried, didn't work. I don't want to mess around with it any longer :D if you have any other suggestion, i'm going to revert the e2e-ci changes

LayZeeDK: @eneajaho

Have you tried ISR with Cypress and Nx Atomizer (`e2e-ci` tasks)?

LayZeeDK: Oh, it's also SSR without ISR

https://cloud.nx.app/runs/16A7H6itJx/task/ssr%3Ae2e-ci--cypress%2Fe2e%2Fapp.cy.ts

hoebbelsB: @LayZeeDK I don't see any evidence about `--parallel` is a bad idea for cloud agents. In fact, tried it on my private projects.

image

You can see that the agent is running 4 tasks in parallel, as I am running it with `--parallel=4`. Imo it should be better to have them run things in parallel, wdyt?

LayZeeDK: Are you using Nx Agents in your private project?

```shell
nx-cloud start-ci-run --distribute-on="./nx/workflows/dynamic-changesets.yml"
```

hoebbelsB: @LayZeeDK yes!

PR #1751, opened 2024-08-20
fix(isr): filter urlsOnHold by cacheKey instead of url

maxisam:
This pull request fixes a bug in the CacheRegeneration class where the filtering of urlsOnHold was done based on the wrong variable. Instead of filtering by cacheKey, it was filtering by url. This caused incorrect removal of urls from urlsOnHold. The fix updates the filtering logic to use cacheKey instead, ensuring that the correct urls are removed from urlsOnHold.


eneajaho:
Love it!

Issue #1736, opened 2024-06-04
Providing an instance of InMemoryCacheHandler doesn't work - instanceof CacheHandler is false 


perholmang:
When creating an instance of InMemoryCacheHandler and providing it to ISRHandler, the instance isn't being used but instead the default handler. It seems the problem is that it checks that the provided handler is an instance of `CacheHandler` which is false.

```
const cacheHandler = new InMemoryCacheHandler();
  console.log(cacheHandler instanceof CacheHandler); // this returns false

  const isr = new ISRHandler({
    indexHtml,
    invalidateSecretToken: 'helloworld', // replace with env secret key ex. process.env.REVALIDATE_SECRET_TOKEN
    enableLogging: true,
    cache: cacheHandler,
    serverDistFolder,
    browserDistFolder,
    bootstrap,
    commonEngine,
  });
```
# Environment

"@rx-angular/isr": "^17.1.0"

PR #1753, opened 2024-08-22
fix(isr): in memory cache handler should use extends  #1736

maxisam:
fix the issue #1736


maxisam:
image

but the link actually work

eneajaho:
Thank you!!

Issue #1720, opened 2024-05-08
ISR: Ignore query params


arsenasatryan:
# Problem Solved By The Feature
Sometimes it is required to use the same cached version even URL is not the same for particular 2 requests. for example, Google Search Ads adds a query parameter to the URL which is different for each request but the content of the page is the same

# Solution

As a solution, it is possible to add a boolean that says to ignore query parameters for a route. additionally, there can be an array of query param names that can specified not to be ignored in case of query param based pagination for example.

SkyZeroZx:
I add to this both google ads, linkedin, facebook, etc. in some publication it adds a query param in front of it which could end up saturating the cache whether in the file system or redis or others.
It would be good to take it into account

eneajaho:
Hello @arsenasatryan 
This was added in #1757 and will be released soon.

PR #1757, opened 2024-08-23
feat(isr): add allowed query params options #1743

### LINKED ISSUE #1743: ISR: Avoiding immense cache-keys generation (Adding the possibility of defining a custom cacheKey)

This is related to https://github.com/rx-angular/rx-angular/issues/1720 but would be a bit more generic
This is related to https://github.com/rx-angular/rx-angular/issues/1604 but the fix done there is not fixing the issue described here

We experienced problems using the ISR library since attackers tried to access a lot of urls. Since [getCacheKey](https://github.com/rx-angular/rx-angular/blob/main/libs/isr/server/src/isr-handler.ts#L309) uses the `req.url` this will lead to a separate cache-entry for each request. Within short time it filled our cache with more than 15GB of data. We had to patch the library in order to use `req.path` for now.

E.g. 

* Changing parameters: `/my-route?1`, `/my-route?something=1`, , `/my-route?something=3` ...
* Calling 404 pages: `/something-404` that would look exactly the same but are not

I'm aware that this might be the wanted behaviour because of pagination etc. But i want to have the control of modifying this and i am not sure if it's the best default. For simple pages req.path could be an alternative. Even with pagination one could simply iterate over pages and generate an immense amount of entries.

For the 404-Pages I found the solution to not cache it (`revalidate: null`). I will gladly add this as a hint to the documentation to protect other users. So i cannot cache the 404 page, but i also don't create a new entry for every 404-page called.

# Solution

I would suggest a configuration that lets you specifying your cacheKey-Function. Ideally the interface would allow to pass in the `req` and then users could still allow certain params.

For the 404-problem i was first thinking of having a beforeAddingToCache-hook where i can decide from the given response-code if i want to cache it or not.

# Alternatives Considered

I did not come up with any other ideas, but happy to discuss it here!

# My current workaround:

* usage of [patch-package](https://www.npmjs.com/package/patch-package) (follow the usage there)
* My patch (wont fit your need, but shows where the problems are): create/adapt this [patches/@rx-angular+isr+17.1.0.patch]([@rx-angular+isr+17.1.0.patch](https://github.com/user-attachments/files/16082370/%40rx-angular%2Bisr%2B17.1.0.patch)

maxisam:
this should fix #1743 #1720

### DIFF
--- a/apps/docs/docs/isr/getting-started.md
+++ b/apps/docs/docs/isr/getting-started.md
@@ -101,6 +101,9 @@ export function app(): express.Express {
 
   const commonEngine = new CommonEngine();
 
+  // This array of query params will be allowed to be part of the cache key. If null, all query params will be allowed. If empty, no query params will be allowed.
+  const allowedQueryParams = ['page'];
+
   // 2. 👇 Instantiate the ISRHandler class with the index.html file
   // highlight-start
   const isr = new ISRHandler({
@@ -111,14 +114,12 @@ export function app(): express.Express {
     browserDistFolder,
     bootstrap,
     commonEngine,
+    allowedQueryParams,
   });
   // highlight-end
 
   server.use(express.json());
-  server.post(
-    '/api/invalidate',
-    async (req, res) => await isr.invalidate(req, res)
-  );
+  server.post('/api/invalidate', async (req, res) => await isr.invalidate(req, res));
 
   server.set('view engine', 'html');
   server.set('views', browserDistFolder);
@@ -130,7 +131,7 @@ export function app(): express.Express {
     '*.*',
     express.static(browserDistFolder, {
       maxAge: '1y',
-    })
+    }),
   );
 
   // 3. 👇 Use the ISRHandler to handle the requests
@@ -140,7 +141,7 @@ export function app(): express.Express {
     // Serve page if it exists in cache
     async (req, res, next) => await isr.serveFromCache(req, res, next),
     // Server side render the page and add to cache if needed
-    async (req, res, next) => await isr.render(req, res, next)
+    async (req, res, next) => await isr.render(req, res, next),
   );
   // highlight-end
 
@@ -177,9 +178,7 @@ import { ISRHandler } from '@rx-angular/isr/server';
 export function app(): express.Express {
   const server = express();
   const distFolder = join(process.cwd(), 'dist/docs/browser');
-  const indexHtml = existsSync(join(distFolder, 'index.original.html'))
-    ? 'index.original.html'
-    : 'index';
+  const indexHtml = existsSync(join(distFolder, 'index.original.html')) ? 'index.original.html' : 'index';
 
   // 2. 👇 Instantiate the ISRHandler class with the index.html file
   // highlight-start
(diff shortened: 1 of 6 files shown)


eneajaho:
Hi @maxisam 
Thanks for doing this. I'd also like to have some docs for it, in order to showcase this new feature 🙌

maxisam:
sure, not sure where should I put this. I did put some doc in the comments with the config

libs/isr/models/src/isr-handler-config.ts
```diff
@@ -97,6 +97,13 @@ export interface ISRHandlerConfig {
    * ],
    */
   variants?: RenderVariant[];
+
+  /**
+   * This array of query params will be allowed to be part of the cache key.
+   * If not provided, which is null, all query params will be part of the cache key.
```
eneajaho, line 103:
```suggestion
   * If not provided, which is undefined, all query params will be part of the cache key.
```

apps/docs/docs/isr/getting-started.md
```diff
@@ -101,6 +101,9 @@ export function app(): express.Express {
 
   const commonEngine = new CommonEngine();
 
+  // This array of query params will be allowed to be part of the cache key. If null, all query params will be allowed. If empty, no query params will be allowed.
```
eneajaho, line 104:
```suggestion
  // This array of query params will be allowed to be part of the cache key. If undefined, all query params will be allowed. If empty array, no query params will be allowed.
```

eneajaho (approved):

maxisam:
@eneajaho Thanks for the review, could you rerun the build?

maxisam:
@eneajaho everything is green 🎉

libs/isr/server/src/utils/cache-utils.spec.ts
```diff
@@ -0,0 +1,38 @@
...
+  });
+
+  it('should return the URL with query parameters when it is null or undefined', () => {
+    const url = '/page?param1=value1&param2=value2';
+    const result = getCacheKey(url, null, null);
+    expect(result).toBe('/page?param1=value1&param2=value2');
+  });
+
+  it('should include only allowed query parameters in the result', () => {
+    const url = '/page?allowed=value&disallowed=value';
```
sihu, line 18: since penetration tests would still try weird things, what would be with the cases:

* `const url = '/page?allowed=value&disallowed=value&allowed=value2';
* `const url = '/page?allowed=value&disallowed=value&allowed=value';

sihu: btw @maxisam thx for the great work! 🙏

maxisam: ummm, the current behavior is simply just remove the disallowed part and keep all allowed part

so they will be 

- `const url = '/page?allowed=value&allowed=value2';
- `const url = '/page?allowed=value&allowed=value';

I guess we can override the duplicated one ?

so we should make it like

- `const url = '/page?allowed=value2';
- `const url = '/page?allowed=value';

another thought is I actually want to make the getCacheKey like a pluggable function. 

so everyone can have their own cacheKey. 

@eneajaho how do you think about pluggable cacheKey function?

eneajaho: @maxisam Yeah we can make the cacheKey function pluggable, good idea.

eneajaho: Let's add it in another PR. Just merged this one.

Issue #1742, opened 2024-07-03
ISR: Invalidating cache does not allow to modify


sihu:
# Description

I was two sided if this is a issue or a feature request. Feel free to change. I noticed a diverging behaviour when invalidating. Before suggesting changes, I wanted to ensure that I understood it correctly and my assumptions are correct:

* when adding a key first time to the cache it calls `renderUrl` and if given modifies the content (e.g. https://github.com/rx-angular/rx-angular/blob/main/libs/isr/server/src/isr-handler.ts#L257-L263) and add this modified html to the cache: https://github.com/rx-angular/rx-angular/blob/main/libs/isr/server/src/isr-handler.ts#L284
* when invalidating the cache, it will also call the `renderUrl` but without modifiying since this cannot be specified in the config (e.g. https://github.com/rx-angular/rx-angular/blob/main/libs/isr/server/src/isr-handler.ts#L284) and add the generated html (without the same modifications) to the cache (e.g. https://github.com/rx-angular/rx-angular/blob/main/libs/isr/server/src/isr-handler.ts#L119)
* when serving from cache, we can also pass a `modifyCachedHtml` which would be applied to the response to the (maybe already modified) cache value. This might be wanted to modifiy it further .

I would expect that the cache content is generated the same way as the first call. Therefore I'd suggest these changes:

* `InvalidateConfig` should also have a `modifyGeneratedHtml` and apply it before adding it to the cache
* applying the `modifyGeneratedHtml` should maybe be part of `renderUrl` so they cannot diverge. 
* using `invalidate` as used on https://www.rx-angular.io/docs/isr/on-demand-revalidation would not behave the same i

The easiest approach for me would be to add the `modifyCachedHtml` and do the same change as i do on `modifyGeneratedHtml` but i think this is not so intuitive. If this would be a Handler-Config and not a config on the function, the usage would be simpler.

# Steps to Reproduce the Issue

* specify a `modifyGeneratedHtml` and make a notable change (a big `h1` at the beginning)
* call a site and put it to cache
* call `invalidate` to the site you just cached
* expect the notable change not to be there anymore

# Tasks to Resolve This

To be discussed. Either add the `modifyGeneratedHtml` to `InvalidateConfig` or change the interface.

maxisam:
[...]

Issue #1758, opened 2024-08-23
ISR: Render behaves differently between render / regenerate / invalidate


maxisam:
# Description

only render has feature for `modifyGeneratedHtml`, it is missing from the other two

I think the issue here is we should consolidate the render function a bit more.

also with modifyGeneratedHtml, I think there is no need for `cacheMsg` in InMemoryCacheHandler

We can have a default `modifyGeneratedHtml` function just like cacheMsg 

# Notes

I will try to fix this in my fork, if you like the proposed solution, I will make a PR.

Sorry making a bunch of changes in a short time, I am trying to port our existing feature to this library. My goal is trying to match the upstream as much as possible.

eneajaho:
Yes, that would be great! Thanks for doing all this! I will make sure to review and merge everything!

maxisam:
Thanks! Your library is really awesome. 

So my current plan is rename regeneration to generation and add another param for different mode, since they actually behave a little different. But I might change my mind after working on this a bit more. 

I plan to do this one first and then #1755 



Welcome to give me any suggestion.

maxisam:
I think we shouldn't have 2 different configs to modify the html for serveFromCache and render

https://github.com/rx-angular/rx-angular/blob/3f919167f68a1bc969e95876d81a4dbace306fc1/libs/isr/models/src/isr-handler-config.ts#L127

https://github.com/rx-angular/rx-angular/blob/3f919167f68a1bc969e95876d81a4dbace306fc1/libs/isr/models/src/isr-handler-config.ts#L110

I understand the need to have a way to serve different variant of a page, but  variants to do that already.

with 2 different configurations, 

we have to provide render config here

https://github.com/rx-angular/rx-angular/blob/3f919167f68a1bc969e95876d81a4dbace306fc1/libs/isr/server/src/isr-handler.ts#L209-L217

If we only allow one config for that, we can just add providers and modifyCachedHtml to ISRConig and use the ISRConfig for everything.

Any thought?

maxisam:
Actually, I think I can still preserve modifyCachedHtml in ServeFromCacheConfig to avoid breaking change.

But I think we have to add modifyGeneratedHtml in ISRHandlerConfig

maxisam:
I created a commit to show the idea. I will test it more tomorrow but I think it should work. It does change a lot of code so feel free to give me some feedback.

maxisam:
this is done

PR #1760, opened 2024-08-29
fix(isr): use modifyGeneratedHtml in all cache generation process

maxisam:
fix: #1742 #1758 

- move modifyGeneratedHtml to ISRHandlerConfig
- consolidate cache generation related code
- serve existing cache first instead of waiting regeneration
- create example to show modifyGeneratedHtml in demo

Feel free to make any suggestion. Thanks!


maxisam:
Trying my best to remove unrelated changes but the lint just keep adding "," back. I hope it is not too hard to review with those changes.

eneajaho (approved):
This cleanup was really needed! Thank you!! Amazing job!

eneajaho:
@maxisam 
A rebase is needed. Then we can merge 🎉

maxisam:
> @maxisam A rebase is needed. Then we can merge 🎉

Do you mean you prefer to rewrite some commits and force push it? I usually don't do that since I use squash merge in general.

However, if you prefer to keep some meaningful commits, I can rebase it and force push

hoebbelsB:
we are generating the changelog from the commit messages. In best case they are meaningful and we don't have to squash. However, it's also possible. If we can fit your efforts into one meaningful commit message, it should also be fine. Wdyt @eneajaho ?

eneajaho:
This PR was breaking, so I will merge this other one too (https://github.com/rx-angular/rx-angular/pull/1766) so we can release these new features in a minor instead of having to release a major version. 

Basically the devs can configure the modify fn in two places and in the next major we can remove the one in the render config.

Issue #1755, opened 2024-08-22
ISR: Compression


maxisam:
# Problem Solved By The Feature

The size of the page could be really big without compression. 

# Solution

Add an option to store cache html as compressed and return with compressed html and compress header directly

# Additional Context

Since we have lots of pages and size are pretty big as well. With compress, it makes the whole process more efficiently like the traffic between redis and return the compressed html directly.

eneajaho:
Should we make the compression pluggable? So the users can provide their own compression? I'd like to have this as a config that they can provide a compress function, this way they can use whatever they need to do with it.

maxisam:
That is a great idea. I will see if I can make it.

maxisam:
Just realized I need to add another change to this PR. Currently cache handler take string as value. But with compress function, it should take Buffer as value. so the value could be either Buffer or String, it will only be Buffer if compress feature is enabled.

Buffer to Base64 string could grow as much as 40% in size.

maxisam:
on the other hands, Angular 19 is coming, maybe we just catch this chance and push to v19 😆

PR #1763, opened 2024-08-30
feat(isr): add compression

maxisam:
#1755 

- update type of express, to work with express compress 
- add two more options
  - compressHtml: compression callback to compress the html before storing it in the cache.
  - htmlCompressionMethod: html compression method, it will use gzip by default if not provided.
- with compressHtml handler, it will automatically disable modifyCachedHtml, since cached html is btye after compressed.

This needs to wait until other PR #1760 #1757 get merged


eneajaho:
Hello @maxisam 
Can you create another PR to only include compression commits? All the previous PRs are already merged so you can just cherry pick the commits I guess. 

Also, I was thinking to compress the result after we modify the HTML, and then store in cache. 

-> Render
-> Modify Html
-> Compress 
-> Store in cache

This way we don't break anything.

maxisam:
yes, I can do that. I thought I render HTML first and compress after currently. I need to double check it when I get back

maxisam:
#1768 use this one instead

PR #1764, opened 2024-08-31
feat(template): add DataSource support for rx-virtual-for directive

friendlyAce:
Proposal
Link to feature request: https://github.com/rx-angular/rx-angular/issues/1765

Currently, the rx virtual for directive supports rendering from signal, observable and static iterables. This proposal aims to extends the functionality by adding support for the DataSource interface that is supported in the CdkVirtualForOf of the Angular CDK. The rx virtual for directive could directly connect to and disconnect from the given DataSource when needed.

This PR introduces support for [DataSource](https://github.com/angular/components/blob/main/src/cdk/collections/data-source.ts) in rx-virtual-for, aligning with [Angular CDK's](https://material.angular.io/cdk/scrolling/overview#specifying-data) implementation while leveraging rx-angular's powerful reactive approach.

When users want to use their existing data source implementations they will need to implement the [CollectionViewer](https://github.com/angular/components/blob/main/src/cdk/collections/collection-viewer.ts#L18) interface at the component that uses rx virtual for and link the viewRange: ListRange output of that virtual for directive to their DataSource implementation for connecting and passing the returned Observable to rxVirtualFor and handle disconnecting from that data source when needed.

Key features:
- Implement DataSource interface compatible with rx-virtual-for
- Add support for DataSource input in the rxVirtualForOf directive
- Enhance the rxVirtualForOf to handle the DataSource input case, the connection to and disconnection from the data source

Benefits:
- Improved compatibility with te existing Angular CDK patterns
- Enhanced flexibility for data handling in virtual scrolling scenarios
- Better integration with complex data fetching and manipulation use cases (lazy on demand requesting data, filtering, etc.)

Implementation details:
- Modified rxVirtualForOf input to accept DataSource<T>
- Add use-case to check whether the input is a data source
- Updated internal logic to connect to the DataSource using the current ListRange and disconnect from the DataSource when destroyed

Documentation:
- Needs to be added if the feature proposal is desired (apps\docs\docs\template\api\virtual-scrolling.mdx)

In addition to the implementation details mentioned above, I have declared the necessary types such as DataSource, CollectionViewer, and the isDataSource method within the rx-angular template package itself.

This approach eliminates the need for a (peer) dependency on Angular CDK, ensuring that rx-virtual-for remains self-contained and independent, similar how it was done with the ListRange interface.

Feedback, concerns, improvements are more than welcome

BR,
Vincent

### DIFF
--- a/libs/template/experimental/virtual-scrolling/src/lib/virtual-for.directive.ts
+++ b/libs/template/experimental/virtual-scrolling/src/lib/virtual-for.directive.ts
@@ -217,6 +217,11 @@ export class RxVirtualFor<T, U extends NgIterable<T> = NgIterable<T>>
   });
 
   /** @internal */
+  private connectedDataSource?: DataSource<T>;
+  /** @internal */
+  private collectionViewer?: CollectionViewer;
+
+  /** @internal */
   private _differ?: IterableDiffer<T>;
 
   /** @internal */
@@ -237,11 +242,11 @@ export class RxVirtualFor<T, U extends NgIterable<T> = NgIterable<T>>
    *     [hero]="hero"></app-hero>
    * </rx-virtual-scroll-viewport>
    *
-   * @param potentialSignalOrObservable
+   * @param potentialSignalOrObservableOrDataSource
    */
   @Input()
   set rxVirtualForOf(
-    potentialSignalOrObservable:
+    potentialSignalOrObservableOrDataSource:
       | Observable<(U & NgIterable<T>) | undefined | null>
       | Signal<(U & NgIterable<T>) | undefined | null>
       | DataSource<T>
@@ -249,13 +254,17 @@ export class RxVirtualFor<T, U extends NgIterable<T> = NgIterable<T>>
       | null
       | undefined,
   ) {
-    if (isSignal(potentialSignalOrObservable)) {
+    if (isSignal(potentialSignalOrObservableOrDataSource)) {
       this.staticValue = undefined;
       this.renderStatic = false;
       this.observables$.next(
-        toObservable(potentialSignalOrObservable, { injector: this.injector }),
+        toObservable(potentialSignalOrObservableOrDataSource, {
+          injector: this.injector,
+        }),
       );
-    } else if (this.isDataSource(potentialSignalOrObservable)) {
+    } else if (this.isDataSource(potentialSignalOrObservableOrDataSource)) {
+      this.disconnectDataSource();
+
       this.staticValue = undefined;
       this.renderStatic = false;
 
@@ -263,20 +272,19 @@ export class RxVirtualFor<T, U extends NgIterable<T> = NgIterable<T>>
         viewChange: this.scrollStrategy.renderedRange$,
       };
 
-      this.observables$.next(
-        potentialSignalOrObservable.connect(collectionViewer),
-      );
+      this.collectionViewer = collectionViewer;
+      this.connectedDataSource = potentialSignalOrObservableOrDataSource;
 
-      this._destroy$.pipe(take(1)).subscribe(() => {
-        potentialSignalOrObservable.disconnect(collectionViewer);
-      });
-    } else if (!isObservable(potentialSignalOrObservable)) {
-      this.staticValue = potentialSignalOrObservable;
+      this.observables$.next(
+        potentialSignalOrObservableOrDataSource.connect(collectionViewer),
+      );
+    } else if (!isObservable(potentialSignalOrObservableOrDataSource)) {
+      this.staticValue = potentialSignalOrObservableOrDataSource;
       this.renderStatic = true;
     } else {
       this.staticValue = undefined;
       this.renderStatic = false;
-      this.observables$.next(potentialSignalOrObservable);
+      this.observables$.next(potentialSignalOrObservableOrDataSource);
     }
   }
 
@@ -290,8 +298,7 @@ export class RxVirtualFor<T, U extends NgIterable<T> = NgIterable<T>>
       | undefined,
   ): value is DataSource<T> {
     return (
-      value !== null &&
-      value !== undefined &&
+      value != null &&
       'connect' in value &&
       typeof value.connect === 'function' &&
       !(value instanceof ConnectableObservable)
@@ -678,7 +685,17 @@ export class RxVirtualFor<T, U extends NgIterable<T> = NgIterable<T>>
   }
 
   /** @internal */
+  private disconnectDataSource(): void {
+    if (this.connectedDataSource && this.collectionViewer) {
+      this.connectedDataSource.disconnect(this.collectionViewer);
+      this.connectedDataSource = undefined;
+      this.collectionViewer = undefined;
+    }
+  }
+
+  /** @internal */
   ngOnDestroy() {
+    this.disconnectDataSource();
     this._destroy$.next();
     this.templateManager.detach();
   }


hoebbelsB:
hey @friendlyAce thanks a lot for your contribution! Will make a review over the weekend!

libs/template/experimental/virtual-scrolling/src/lib/virtual-for.directive.ts
```diff
@@ -251,6 +255,21 @@ export class RxVirtualFor<T, U extends NgIterable<T> = NgIterable<T>>
       this.observables$.next(
         toObservable(potentialSignalOrObservable, { injector: this.injector }),
       );
+    } else if (this.isDataSource(potentialSignalOrObservable)) {
```
hoebbelsB, line 258: wouldn't it be `potentialSignalOrObservableOrDataSource` now? :)

libs/template/experimental/virtual-scrolling/src/lib/model.ts
```diff
@@ -34,6 +34,17 @@ export interface ListRange {
   end: number;
 }
 
+export abstract class DataSource<T> {
```
hoebbelsB, line 37: do you know if it's possible to maybe have an `import type { DataSource } from '@angular/cdk'` without having it to mention as `peerDependency`?

That would be the absolute best case

libs/template/experimental/virtual-scrolling/src/lib/virtual-for.directive.ts
```diff
@@ -251,6 +255,21 @@ export class RxVirtualFor<T, U extends NgIterable<T> = NgIterable<T>>
...
+        viewChange: this.scrollStrategy.renderedRange$,
+      };
+
+      this.observables$.next(
+        potentialSignalOrObservable.connect(collectionViewer),
+      );
+
+      this._destroy$.pipe(take(1)).subscribe(() => {
+        potentialSignalOrObservable.disconnect(collectionViewer);
+      });
```
hoebbelsB, line 272: i'm not a big fan of that. We should do the `disconnect` in ngOnDestroy.

libs/template/experimental/virtual-scrolling/src/lib/virtual-for.directive.ts
```diff
@@ -261,6 +280,24 @@ export class RxVirtualFor<T, U extends NgIterable<T> = NgIterable<T>>
...
+    value:
+      | (U & NgIterable<T>)
+      | Observable<U & NgIterable<T>>
+      | DataSource<T>
+      | null
+      | undefined,
+  ): value is DataSource<T> {
+    return (
+      value !== null &&
+      value !== undefined &&
```
hoebbelsB, line 294: can't this be shortened with value != null?

libs/template/experimental/virtual-scrolling/src/lib/virtual-for.directive.ts
friendlyAce: I'll rename it from `potentialSignalOrObservable` to `potentialSignalOrObservableOrDataSource`

libs/template/experimental/virtual-scrolling/src/lib/virtual-for.directive.ts
friendlyAce: I added the `connectedDataSource` and `collectionViewer` properties for being able to reference the objects
I also added `disconnectDataSource()` now, which does the clean up for the data source instead of doing it with the subscription to _destroy$.

`disconnectDataSource` will now be invoked in the `ngOnDestroy` lifecycle hook
I also added that on new input changes to `rxVirtualForOf`, the previous data source gets disconnected, in case the user does not destroy the component, but re-uses it and just updates the input with a new data source implementation instead)

libs/template/experimental/virtual-scrolling/src/lib/virtual-for.directive.ts
friendlyAce: Yes it can, ill fix it

libs/template/experimental/virtual-scrolling/src/lib/model.ts
friendlyAce: @hoebbelsB Hmm im not fully sure on that, i tried to have a bit of testing on that but i kept running into a bunch of possibly unrelated issues and stopped checking.

My idea was to use Typescripts "type only imports" e. g. `import type { DataSource } from '@angular/cdk/collections` ([doc](https://www.typescriptlang.org/docs/handbook/modules/reference.html#type-only-imports-and-exports))
As when compiling they should just be dropped out of the dist output files
For that i reverted the DataSource and CollectionViewer declarations which i added in

```diff
diff --git a/libs/template/experimental/virtual-scrolling/src/index.ts b/libs/template/experimental/virtual-scrolling/src/index.ts
index 86b34086..8c50f1ed 100644
--- a/libs/template/experimental/virtual-scrolling/src/index.ts
+++ b/libs/template/experimental/virtual-scrolling/src/index.ts
@@ -1,6 +1,4 @@
 export {
-  CollectionViewer,
-  DataSource,
   ListRange,
   RxVirtualForViewContext,
   RxVirtualScrollElement,
diff --git a/libs/template/experimental/virtual-scrolling/src/lib/model.ts b/libs/template/experimental/virtual-scrolling/src/lib/model.ts
index c6ed5c35..aacb1d3e 100644
--- a/libs/template/experimental/virtual-scrolling/src/lib/model.ts
+++ b/libs/template/experimental/virtual-scrolling/src/lib/model.ts
@@ -34,17 +34,6 @@ export interface ListRange {
   end: number;
 }
 
-export abstract class DataSource<T> {
-  abstract connect(
-    collectionViewer: CollectionViewer,
-  ): Observable<NgIterable<T>>;
-  abstract disconnect(collectionViewer: CollectionViewer): void;
-}
-
-export interface CollectionViewer {
-  viewChange: Observable<ListRange>;
-}
-
 /**
  * @Directive RxVirtualScrollStrategy
  *
diff --git a/libs/template/experimental/virtual-scrolling/src/lib/virtual-for.directive.ts b/libs/template/experimental/virtual-scrolling/src/lib/virtual-for.directive.ts
index 174faf3e..14ee2512 100644
--- a/libs/template/experimental/virtual-scrolling/src/lib/virtual-for.directive.ts
+++ b/libs/template/experimental/virtual-scrolling/src/lib/virtual-for.directive.ts
@@ -56,8 +56,6 @@ import {
   tap,
 } from 'rxjs/operators';
 import {
-  CollectionViewer,
-  DataSource,
   ListRange,
   RxVirtualForViewContext,
   RxVirtualScrollStrategy,
@@ -67,6 +65,10 @@ import {
   createVirtualListTemplateManager,
   RxVirtualListTemplateManager,
 } from './virtual-list-template-manager';
+import type {
+  DataSource,
+  CollectionViewer
+} from '@angular/cdk/collections';
 import {
   DEFAULT_TEMPLATE_CACHE_SIZE,
   RX_VIRTUAL_SCROLL_DEFAULT_OPTIONS,
```

And instead i used the type only imports directly in virtual-for.directive.ts, e. g. 
```
import type {
  DataSource,
  CollectionViewer
} from '@angular/cdk/collections';
```

Therefore, at least for the library build, the @angular/cdk package needs to be resolvable, which it is, as its already part of your package.json dependencies [here](https://github.com/rx-angular/rx-angular/blob/main/package.json#L43).

After these changes i ran the build for the template library via `npx nx run template:build:production`, which gave me the according dist output `dist\libs\template`.

Taking a short look at the compiled .mjs file, indeed the import from @angular/cdk is NOT mentioned (as its a type import)
`dist\libs\template\fesm2022\template-experimental-virtual-scrolling.mjs`

But when having a look at the compiled type declaration file:
`dist\libs\template\experimental\virtual-scrolling\lib\virtual-for.directive.d.ts`
It still contains the @angular/cdk import...
import type { DataSource } from '@angular/cdk/collections';
Which could cause an issue for library consumers.

So i wanted to check this and therefore created a new nx angular workspace via
`npx create-nx-workspace demo-workspace --preset=angular`
And installed the local dist template library via 
`npm i ../rx-angular/dist/libs/template`
package.json => "@rx-angular/template": "file:../rx-angular/dist/libs/template"
I also ensured that the @angular/cdk package is not installed in this new demo-workspace project

Then i creating a small dummy component (similar to [this](https://www.rx-angular.io/docs/template/virtual-scrolling#usage) example), that uses the virtual for directive, the viewport etc. to see if the serve and build of that demo application would work or if i would run into some errors related to the missing angular/cdk peer dependency.

But unfortunately my serve had a bunch of unrelated errors in regards of type conflicts of rxjs

```
Type 'import("DEMO-WORKSPACE/node_modules/rxjs/dist/types/internal/Observable").Observable<any> | undefined' is not assignable to type 'import("rx-angular/node_modules/rxjs/dist/types/internal/Observable").Observable<any> | undefined'.
Types of property 'operator' are incompatible.
```
Which is a classic issue when using local packages, as the workspaces have their own node_modules directories with their own rxjs installations, which TS sees as incompatible
So instead i used npm links
```
npm uninstall @rx-angular/cdk @rx-angular/template
cd rx-angular/dist/libs/template
npm link
cd demo-workspace
npm link @rx-angular/template
```
$ npm ls -g --link=true
C:\Users\...\AppData\Roaming\npm
└── @rx-angular/template@19.2.2 -> rx-angular\dist\libs\template

alternatively, removing the rx-angular/node_modules directory also works. 

But nevertheless when serving i keep getting this error.
image

So either something is broken with the local build of the template lib, or something else. After trying a bunch of stuff, i couldn't get this fixed and therefore couldn't properly check if this approach would work for library consumers.

But i would probably still expect, something like "Cannot find module '@angular/cdk/collections' or its corresponding type declarations" when somebody uses the rx-angular/template library, without them having the angular cdk package also installed

PR #1768, opened 2024-09-03
feat(isr): add compression (#1755)

maxisam:
- update type of express, to work with express compress 
- add two more options
  - compressHtml: compression callback to compress the html before storing it in the cache.
  - htmlCompressionMethod: html compression method, it will use gzip by default if not provided.
- with compressHtml handler, it will automatically disable modifyCachedHtml, since cached html is btye after compressed.
- with compressHtml, it will send Buffer to cache handler instead of string, since stringify buffer could add 50% in size. 
 For redis cache handler, user can use hgetBuffer / hgetallBuffer to handle buffer
- update doc to show how to use redis with buffer, also improve the redis example by using hash, so it doesn't need to convert to json string first.


maxisam:
@eneajaho I just cleaned it up. It should be good for review now

moddi3:
would be nice to ship this, much needed feature

PR #1772, opened 2024-09-06
refactor(isr): use CacheKey instead of route / url  for naming

maxisam:
This is just a refactor for using the same term cacheKey across the code base. (Sometime I got confused.)

Feel free to make different suggestion. I don't really have strong opinion on changing it.


eneajaho:
Thanks for the cleanup! LGTM.

PR #1796, opened 2024-10-06
feat(state): use signals internally as state accumulator

mikelgo:
> [!NOTE]
> Just to start a discussion

I made a small PoC of using signals inside RxState as state accumulator to manage the state  internally. Signals are out of the box due to their nature already optimized to manage state and state changes - so why not use them? :) 

This PoC is not complete (did not look at the failing tests so far, etc), but it is a starting point for discussion. There might also be some changes in behavior which needs to be discussed.

Please note that I just put this together very quickly. If it is considered to move into this direction I will put more efforts in.

Looking forward to feedback :)


hoebbelsB:
well, wouldn't it make the whole state machine async then? Async in terms of scheduled to microtask.

```ts
state.set({ foo: 'bar' });

state.get('foo'); // should still work

```

but this scenario would be a breaking change, right?

```ts
// executed second
state.select('foo').subscribe(console.log);

state.set({ foo: 'bar' });
state.get('foo') // executed first

```

mikelgo:
> well, wouldn't it make the whole state machine async then? Async in terms of scheduled to microtask.
> 
> ```ts
> state.set({ foo: 'bar' });
> 
> state.get('foo'); // should still work
> ```
> 
> but this scenario would be a breaking change, right?
> 
> ```ts
> // executed second
> state.select('foo').subscribe(console.log);
> 
> state.set({ foo: 'bar' });
> state.get('foo') // executed first
> ```

That's right - I did not consider that signals are scheduled differently.

So, yes this scenario would be a breaking change. 

Despite the fact that breaking changes are not nice - do you see any problems with the behavior-change? 🤔

hoebbelsB:
it would also make the custom scheduler option obsolete ;)

PR #1814, opened 2024-12-10
fix: replace toObservableMicrotask private API with proper solution

eneajaho:
Instead of relying on a private API, I copied the impl and replace the private API with public & supported alternative.


hoebbelsB (approved):
LGTM

PR #1825, opened 2025-01-02
feat(template): new reconciliation for RxFor

hoebbelsB:
# Description

This PR introduces the new reconciliation algorithm as part of the `@rx-angular/cdk/template` package. It's implemented in `rxFor` right now.

Originally it was shipped by the angular team for the new `@for` control flow to reduce the amount of operations needed for swap operations.

The original implementation is [here](https://github.com/angular/angular/blob/main/packages/core/src/render3/list_reconciliation.ts) & [here](https://github.com/angular/angular/blob/f8d22a9ba4e426f14f9c7fd608e1ad752cd44eb5/packages/core/src/render3/instructions/control_flow.ts#L281)

# Impact

In general, the new reconciliation algorithm should result in less operations being scheduled to achieve the same goal as with the `IterableDiffer` approach from before. However, this only applies for move / swap operations.

It's also more memory efficient than the iterable differ. It is possible that our custom asynchronous implementation again leads to less memory efficiency, which is something we want to work on in the future.

There are also **behavioral impacts**. Instead of actually moving around DOM, the new reconciliation works by `detaching` & `attaching` views. As rxFor by default uses the concurrent mode, it splits each individual task (attach, detach, update, remove) and works them off in a queue. As we are operating on the DOM, we have to run tasks in the given order. The biggest impact is that you'll visually see views disappearing from the screen when the whole data set is being shuffled around.

This leads to visual instability on the one hand, but also makes sure no view is ever in the wrong position as in the legacy approach.

Let's see it in action with some nice visualizations.

## Swap

Swapping the first item with the last item.

### `experimental`

https://github.com/user-attachments/assets/526a337a-a44f-44fa-af2e-5bca03ebfafa

### `legacy`

https://github.com/user-attachments/assets/c869b1a2-ed2e-44cf-a8d9-5d0c672e8142

## Random Shuffle

Randomly shuffle elements in the array

### `experimental`

https://github.com/user-attachments/assets/7165a2aa-c225-4efb-9a55-d4b08c548216

### `legacy`

https://github.com/user-attachments/assets/5518e2b6-76ca-480b-bb7a-07706173d19a

## Filter

Filter items and remove the filter again

### `experimental`

https://github.com/user-attachments/assets/8a7df564-f0fb-40d6-83e1-5f7a96d8e5d3

### `legacy`

https://github.com/user-attachments/assets/19cd31b2-9ac3-48f3-8083-0751cf285cf0

## TODOs:

* [x] adjust / extend docs
* [x] fix / adjust tests -> additional error handling needed ?
* [x] make new reconciliation algorithm opt-in
  * [x] introduce InjectionToken / extend existing configuration


libs/cdk/template/src/lib/rx-live-collection.ts
```diff
@@ -0,0 +1,332 @@
...
+    },
+  ) {
+    if (this.queue.has(view)) {
+      const entries = this.queue.get(view);
+      const lastEntry = entries[entries.length - 1];
+      /*console.log(
+        'patch I has a work in queue',
+        data.type,
+        this.queue.get(view).map((w) => w.type),
+      );*/
```
edbzn, line 51: There are some comments to cleanup.

libs/template/for/src/lib/for.directive.ts
```diff
@@ -215,7 +210,8 @@ export class RxFor<T, U extends NgIterable<T> = NgIterable<T>>
    * - `@ContentChildren`
    *
    * Read more about this in the
-   * [official docs](https://www.rx-angular.io/docs/template/rx-for-directive#local-strategies-and-view-content-queries-parent).
+   * [official
+   *   docs](https://www.rx-angular.io/docs/template/rx-for-directive#local-strategies-and-view-content-queries-parent).
```
edbzn, line 214: It seems like your editor wrongly formatted the comments.

edbzn (commented):
Nice work @hoebbelsB, amazing PR! I'm wondering something about the terminology and migration path. The new algorithm is marked as experimental but is shipped as the new default behavior of RxFor, it feels strange to ship experimental features by default, wdyt?

hoebbelsB:
@edbzn 

> The new algorithm is marked as experimental but is shipped as the new default behavior of RxFor, it feels strange to ship experimental features by default, wdyt?

that's not true. The legacy approach is still default, you can opt in by using the `provideExperimentalRxForReconciliation()` function. It's also stated in the docs like that

edbzn:
Oh right! I missed the default factory fn that provides the legacy reconciliation token. :slightly_smiling_face: Everything looks good.

PR #1826, opened 2025-01-03
feat(docs): change demos executor, and delete base_href 

AdrianRomanski:


hoebbelsB:
can u also make sure that `production` is the default build target? I've noticed that `yarn nx build demos` is not using the production config

PR #1829, opened 2025-01-03
refactor(state): use DestroyRef instead of OnDestroy life cycle hook

michaelbe812:
Use `DestroyRef` within `RxActionFactory` instead of `OnDestroy` life cycle hook


libs/state/actions/src/lib/actions.factory.ts
```diff
@@ -96,12 +105,4 @@ export class RxActionFactory<T extends Partial<Actions>> implements OnDestroy {
       Object.values(s).forEach((subject: any) => subject.complete());
     });
   }
-
-  /**
-   * @internal
-   * Internally used to clean up potential subscriptions to the subjects. (For Actions it is most probably a rare case but still important to care about)
-   */
-  ngOnDestroy() {
```
michaelbe812, line 104: Is it ok to remove this method without providing a migration? It could be a breaking change if somebody uses this method. However it is marked as internal so nobody should use it

edbzn: > However it is marked as internal so nobody should use it

+1 I'm for removing it completely

hoebbelsB: go for it

libs/state/actions/src/lib/actions.factory.ts
```diff
@@ -22,10 +28,13 @@ type SubjectMap<T> = { [K in keyof T]: Subject<T[K]> };
  * }
  */
 @Injectable()
-export class RxActionFactory<T extends Partial<Actions>> implements OnDestroy {
-  private subjects: SubjectMap<T>[] = [] as SubjectMap<T>[];
+export class RxActionFactory<T extends Partial<Actions>> {
+  private readonly subjects: SubjectMap<T>[] = [] as SubjectMap<T>[];
+  private readonly destroyRef = inject(DestroyRef);
```
JeanMeche, line 33: nit: drop the property and inject only inside the constructor

hoebbelsB (approved):

michaelbe812:
Same comment here regarding CI
CI fails bc of a 429 to url https://medium.com/itnext/incremental-static-regeneration-for-angular-42b0a8440e53 
Probably it is enough to just re run it (locally everything is fine)
However it seems like I do not have the rights to restart the Pipeline

edbzn (approved):

PR #1831, opened 2025-01-05
refactor(state): use DestroyRef within EffectsService

michaelbe812:
Use modern `DestroyRef` also within `RxEffects`

### DIFF
--- a/libs/state/effects/src/lib/effects.service.ts
+++ b/libs/state/effects/src/lib/effects.service.ts
@@ -1,4 +1,10 @@
-import { ErrorHandler, Injectable, OnDestroy, Optional } from '@angular/core';
+import {
+  DestroyRef,
+  ErrorHandler,
+  inject,
+  Injectable,
+  Optional,
+} from '@angular/core';
 import {
   EMPTY,
   from,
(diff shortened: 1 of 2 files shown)


libs/state/effects/src/lib/effects.service.ts
```diff
@@ -223,15 +235,7 @@ export class RxEffects implements OnDestroy, OnDestroy$ {
     return <V>(source: Observable<V>) =>
       source.pipe(
         untilDestroyed(this),
```
JeanMeche, line 237: nit: if you rely on `Destroy` ref, you could as well replace `untilDestroyed(this)` with the framework provided operator. (Maybe in a subsequent PR)

michaelbe812: I also thought about this already. Will open another PR if this one get's merged!
Thanks for pointing this out.

michaelbe812:
CI fails bc of a 429 to url https://medium.com/itnext/incremental-static-regeneration-for-angular-42b0a8440e53 
Probably it is enough to just re run it (locally everything is fine)
However it seems like I do not have the rights to restart the Pipeline

edbzn (approved):

hoebbelsB:
> CI fails bc of a 429 to url https://medium.com/itnext/incremental-static-regeneration-for-angular-42b0a8440e53 Probably it is enough to just re run it (locally everything is fine) However it seems like I do not have the rights to restart the Pipeline

yeah that step is flaky af...

Issue #1830, opened 2025-01-05
demos application serve command defaults to `production`


hoebbelsB:
# Description

the `serve` command of the demos application runs with the production configuration by default now. It should use `development` for serve & `production` for build.

PR #1835, opened 2025-01-06
chore(demos): add development config, set as default for serve

AdrianRomanski:
I have added development configuration for build and serve,
set that configuration as default for serve

Resolves: https://github.com/rx-angular/rx-angular/issues/1830


apps/demos/project.json
```diff
@@ -102,6 +111,12 @@
         },
         "zoneless": {
           "buildTarget": "demos:build:zoneless"
+        },
+        "development": {
+          "buildTarget": "demos:build:development",
+          "open": true,
+          "liveReload": true,
+          "watch": true
```
hoebbelsB, line 119: what are those for? I understand `open`, but not sure about the others :)

hoebbelsB, line 119:
```suggestion
          "open": true
```

liveReload and watch are both true by default, we can omit those. If you like to keep open, it's ok for me

AdrianRomanski: @hoebbelsB ok, so im deleting liveReload, and watch

edbzn (approved):

PR #1838, opened 2025-01-08
refactor(state): drop untilDestroyed operator

michaelbe812:
Drops `untilDestroyed` operator in favor of `takeUntilDestroyed`

### DIFF
--- a/libs/state/effects/src/lib/effects.service.ts
+++ b/libs/state/effects/src/lib/effects.service.ts
@@ -1,10 +1,4 @@
-import {
-  DestroyRef,
-  ErrorHandler,
-  inject,
-  Injectable,
-  Optional,
-} from '@angular/core';
+import { DestroyRef, ErrorHandler, inject, Injectable } from '@angular/core';
 import { takeUntilDestroyed } from '@angular/core/rxjs-interop';
 import {
   EMPTY,
@@ -75,6 +69,7 @@ import { toHook } from './utils';
 export class RxEffects implements OnDestroy$ {
   private static nextId = 0;
   private readonly destroyRef = inject(DestroyRef);
+  private readonly errorHandler = inject(ErrorHandler, { optional: true });
   readonly _hooks$ = new Subject<DestroyProp>();
   private readonly observables$ = new Subject<Observable<unknown>>();
   // we have to use publish here to make it hot (composition happens without subscriber)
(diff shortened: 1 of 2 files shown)


libs/state/effects/src/lib/effects.service.ts
```diff
@@ -91,6 +83,16 @@ export class RxEffects implements OnDestroy$ {
   onDestroy$: Observable<boolean> = this._hooks$.pipe(toHook('destroy'));
   private readonly destroyers: Record<number, Subject<void>> = {};
 
+  constructor(
+    @Optional()
+    private readonly errorHandler: ErrorHandler | null,
```
edbzn, line 88: What about using `inject` here as well?

michaelbe812: Done!

I somehow remembered that somewhere `RxEffects` is instantiated in a place using `new` - but i was wrong :)

edbzn (approved):

libs/state/effects/src/lib/effects.service.ts
```diff
@@ -72,17 +67,9 @@ import { toHook, untilDestroyed } from './utils';
...
-  ) {
-    inject(DestroyRef).onDestroy(() => {
-      this._hooks$.next({ destroy: true });
-      this.subscription.unsubscribe();
-    });
-  }
-
   private static nextId = 0;
+  private readonly destroyRef = inject(DestroyRef);
+  private readonly errorHandler = inject(ErrorHandler, { optional: true });
```
hoebbelsB, line 72: isn't that a breaking change? if someone extended from `RxEffects` it can lead to issues

edbzn: Technically yes, but the recommended usage is by injecting RxEffects, not extending, do we want to consider inheritance as a use case we want to support?

michaelbe812: It is only a breaking change imho when `RxEffects` is instantiated with `new`.
Inheritance with instantiating via DI still works the same if I do not miss something.

hoebbelsB: if someone extended and used `super(errorHandler)` it will cause an error after we merged your PR. That's the only scenario I can think of. 

I guess even no single person in the universe is doing that ^^

hoebbelsB: Your commits are the hints, so be careful what you write :)

michaelbe812:
@hoebbelsB @edbzn 
is there anything which speaks against merging this PR? :)

edbzn:
Not from me :) looks good

michaelbe812:
> Not from me :) looks good

Just asking bc. I can not merge it :)

PR #1840, opened 2025-01-08
feat(state): allow to pass injector to rxState and rxEffects

michaelbe812:
This PR will add the possibility to pass a `Injector` to `rxState` and `rxEffects` to make it easier to instantiate them outside of an injection context.

Right now users need to do: 
```
const injector = inject(Injector)

// outside injection context
runInInjectionContext(injector, () => {
 const state = rxState()
} )

```

### DIFF
--- a/apps/docs/docs/state/effects/effects.mdx
+++ b/apps/docs/docs/state/effects/effects.mdx
@@ -186,9 +186,7 @@ export class EffectsService {
   readonly effects = rxEffects();
 
   // share only some APIs if you like
-  register: typeof this.effects.register = this.effects.register.bind(
-    this.effects
-  );
+  register: typeof this.effects.register = this.effects.register.bind(this.effects);
 }
 ```
 
@@ -347,32 +345,28 @@ class MyComponent {
   readonly effects = rxEffects(({ register }) => {
     register(
       this.login$.pipe(
-        exhaustMap(({ user, pass }) =>
-          this.http.post('/auth/', { user, pass })
-        ),
+        exhaustMap(({ user, pass }) => this.http.post('/auth/', { user, pass })),
         // retry when an error occurs
-        retry()
+        retry(),
       ),
       (data) => {
         alert(`welcome ${data.user}`);
-      }
+      },
     );
 
     register(
       this.login$.pipe(
-        exhaustMap(({ user, pass }) =>
-          this.http.post('/auth/', { user, pass })
-        ),
+        exhaustMap(({ user, pass }) => this.http.post('/auth/', { user, pass })),
         // catch the error and return a custom value
         catchError((err) => {
           return of(null);
-        })
+        }),
       ),
       (data) => {
         if (data) {
           alert(`welcome ${data.user}`);
         }
-      }
+      },
     );
   });
 }
@@ -391,14 +385,10 @@ export class ChartComponent {
   chartVisible$ = new Subject<boolean>();
   chartData$ = this.ngRxStore.select(getListData());
 
-  pollingTrigger$ = this.chartVisible$.pipe(
-    switchMap((isPolling) => (isPolling ? interval(2000) : EMPTY))
-  );
+  pollingTrigger$ = this.chartVisible$.pipe(switchMap((isPolling) => (isPolling ? interval(2000) : EMPTY)));
 
   readonly effects = rxEffects(({ register }) => {
-    register(this.pollingTrigger$, () =>
-      this.ngRxStore.dispatch(refreshAction())
-    );
+    register(this.pollingTrigger$, () => this.ngRxStore.dispatch(refreshAction()));
   });
 }
 ```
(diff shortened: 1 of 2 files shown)


libs/state/effects/src/lib/rx-effects.ts
```diff
@@ -3,24 +3,43 @@ import {
...
 export type RxEffectsSetupFn = (
-  cfg: Pick<RxEffects, 'register' | 'onDestroy'>
+  cfg: Pick<RxEffects, 'register' | 'onDestroy'>,
 ) => void;
 
+export type RxEffectsOptions = {
+  injector?: Injector;
+};
+
+function getInjectorFromOptions<
```
michaelbe812, line 30: Hint to reviewer: 
I duplicated this function otherwise I would have e.g. exported it from `@rx-angular/state` in order to use it in both places which results in a public API-surface. 

Probably no user will ever try to import this function but I did not like it.

Wdyt?

michaelbe812:
I did not add any test cases so far. 
Will work on it

libs/state/src/lib/rx-state.ts
```diff
@@ -22,6 +28,23 @@ export type RxStateSetupFn<State extends object> = (
...
 
+export type RxStateOptions = {
+  injector?: Injector;
+};
+
+function getInjectorFromOptions<
+  SetupFn extends Function,
+  Options extends { injector?: Injector },
+>(setupFnOrOptions?: SetupFn | Options, options?: Options) {
+  if (options) {
```
edbzn, line 39: Would it be safer to check `options?.injector` if we add new options in the future?

libs/state/effects/src/lib/rx-effects.ts
edbzn: Since state depends on cdk we could move it in `cdk/internals/core`  secondary entry point. However I'm also not sure about the need of de-duplicating this fn, it is quite specific and it feels overkill to factor this.

libs/state/src/lib/rx-state.ts
```diff
@@ -46,30 +69,51 @@ export type RxStateSetupFn<State extends object> = (
...
+      computedFrom: legacyState.computedFrom.bind(legacyState),
+      $: legacyState.$,
+      setAccumulator: legacyState.setAccumulator.bind(legacyState),
+      asReadOnly: legacyState.asReadOnly.bind(legacyState),
+    };
 
-  setupFn?.(state);
+    if (setupFnOrOptions && typeof setupFnOrOptions === 'function') {
+      setupFnOrOptions?.(state);
+    }
```
edbzn, line 115: Could be simplified:

```suggestion
    if (typeof setupFnOrOptions === 'function') {
      setupFnOrOptions(state);
    }
```

michaelbe812: Done

michaelbe812:
@edbzn 
I added a few test cased for `rxState`, see commit [7d38500](https://github.com/rx-angular/rx-angular/pull/1840/commits/7d3850060f302afb6929637ec3906d4678408111)

Let me know what you think. A few words here from my side:
- not sure if these tests bring some real value
- pretty sure it can be done in a better way, would appreciate some hints here
- I created a separate `.spec` file because afaik it is not possible to scope `jest.mock`. Adding these tests e.g. in `rx-state.spec.ts` leaded to the result that all other tests where broken due to the mocking of `runInInjectionContext`

edbzn:
@michaelbe812 I think we need to test only two things here:
- we can pass a custom injector to `rxState` and it still works
- it should throw when running without an injection context

Instead of creating a new file and mocking `runInInjectionContext`, I'd make a new utility fn to create `rxState` without an injection context, then test the two test cases I mentioned. Here's a [piece of code where we did that](https://github.com/jscutlery/devkit/blob/7a0ddda8226da64c35b555ea11187cf93c594ada/packages/rx-computed/src/lib/rx-computed.spec.ts#L86), I think you can take it as an inspiration. Wdyt?

michaelbe812:
> @michaelbe812 I think we need to test only two things here:
> 
> * we can pass a custom injector to `rxState` and it still works
> * it should throw when running without an injection context
> 
> Instead of creating a new file and mocking `runInInjectionContext`, I'd make a new utility fn to create `rxState` without an injection context, then test the two test cases I mentioned. Here's a [piece of code where we did that](https://github.com/jscutlery/devkit/blob/7a0ddda8226da64c35b555ea11187cf93c594ada/packages/rx-computed/src/lib/rx-computed.spec.ts#L86), I think you can take it as an inspiration. Wdyt?

Sounds good! Thank you for giving me another perspective. Will look into the snippet you provided and update the test cases. 

Will let you know once I'm done

michaelbe812:
@edbzn just updated the test cases following your recommendation.

Let me know if it meets your expectations or some polishing is needed.

michaelbe812:
@edbzn 
just to quickly ask for feedback: anything missing so that we can merge this PR?

edbzn (approved):

hoebbelsB:
@michaelbe812 I would love to see some docs about this? Just adding a section to the rx state docs would be fair enough

michaelbe812:
> @michaelbe812 I would love to see some docs about this? Just adding a section to the rx state docs would be fair enough

Oh how could i miss that! 😅 

Done! :)

apps/docs/docs/state/effects/effects.mdx
michaelbe812, line 1: `nx format:write` 
changed it like this.

Probably we should run it once for the complete project

michaelbe812:
@edbzn 
I saw that this PR is approved - you might wanna merge it once u have a minute :)

PR #1851, opened 2025-01-28
chore: update semver plugin

hoebbelsB:
Updating the jscutlery/semver plugin.

@edbzn I've experienced a lot of issues recently when publishing our libs. Looks like it wouldn't run `build` after versioning. So it always tries to publish the last version again and fails. Is the update fixing it?

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,7 +77,7 @@
     "@angular/language-service": "19.0.0",
     "@commitlint/cli": "^19.2.1",
     "@commitlint/config-angular": "^19.1.0",
-    "@jscutlery/semver": "^4.1.0",
+    "@jscutlery/semver": "^5.5.1",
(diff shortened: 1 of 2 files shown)


edbzn:
The new update won't fix anything related to the issue you're mentioning. Are you running build with `--skipNxCache` ? it seems to be the reason behind your issue.

edbzn (approved):

hoebbelsB:
I don't run build at all. I just do `nx publish state --skipNxCache` and it fails because it doesn't execute the proper build

edbzn:
image

Ok weird, publish correctly depends on build so I'm not sure why it doesn't execute it in your case. It would be better to have a release pipeline so we are sure it always run in the same way.

PR #1855, opened 2025-01-28
chore: migrate Nx to v20.4.0

edbzn:


edbzn:
Seems like the CI is out of memory:

> One or more agents restarted due to out of memory errors
If the issue persists, consider trying a [larger resource class](https://nx.dev/pricing#resource-classes).

hoebbelsB:
that didn't happen before. but you can use a higher class if needed

hoebbelsB:
@edbzn you wanna continue on the nx upgrade?

Issue #1857, opened 2025-02-04
Virtual Scroll: scroll position jumps when using keepScrolledIndexOnPrepend


m-gericke:
# Description

![Image](https://github.com/user-attachments/assets/b16d61ab-c53c-4daa-9656-dd4c3a426f51)

This happens when inserting items during `(viewRange)` events while scrolling upwards. The scroll position jumps to the start of the previous item and scrolls upwards from there. This example is using the `FixedSizeVirtualScrollStrategy` but the bug should be present in all strategies.

This happens because the view index has already (correctly) changed when the [keepScrolledIndexOnPrepend handler](https://github.com/rx-angular/rx-angular/blob/96750c4ef957bd82e5b4d2fb4488535b6020190e/libs/template/experimental/virtual-scrolling/src/lib/scroll-strategies/fixed-size-virtual-scroll-strategy.ts#L268) happens. That handler then jumps to the *top* of that index instead of the *current offset* of that index.

# Steps to Reproduce the Issue

1. add virtual scroll viewport with FixedSizeVirtualScrollStrategy
2. enable [keepScrolledIndexOnPrepend] input
3. add (viewRange) event handler that prepends items to the array

This is more noticable the larger the item height is.

# Environment

Angular CLI: 19.1.3
Node: 18.19.1
Package Manager: pnpm 10.2.0
OS: win32 x64

Angular: 19.1.2
... animations, common, compiler, compiler-cli, core, forms
... platform-browser, platform-browser-dynamic, platform-server
... router

Package                         Version
---------------------------------------------------------
@angular-devkit/architect       0.1901.3
@angular-devkit/build-angular   19.1.3
@angular-devkit/core            19.1.3
@angular-devkit/schematics      19.1.3
@angular/cli                    19.1.3
@angular/ssr                    19.1.3
@schematics/angular             19.1.3
rxjs                            7.8.1
typescript                      5.7.3

# Notes

I've worked around this by including an offset between the current items top and the current scroll position like this:

```ts
// add offset parameter to scrollToIndex
scrollToIndex(index: number, behavior?: ScrollBehavior, offset: number = 0): void {
  const scrollTop = this.itemSize * index;
  this.viewport!.scrollTo(this.viewportOffset + scrollTop + offset, behavior);
}

// inside keepScrolledIndexOnPrepend handler
this.scrollToIndex(scrollTo, undefined,  this.scrollTop % this._itemSize);
```

[...]

PR #1859, opened 2025-02-10
fix(template): keep scroll offset when using keepScrolledIndexOnPrepend (#1857)

m-gericke:
This fix prevents items from jumping from the anchor items current offset to it's top edge when prepending items during an upwards scroll.


hoebbelsB:
hey @m-gericke thank you for the contribution! Sorry for the late reply :pray: 

Can u give me en example where and when this happens? Any kind of demo would be appreciated. In best case we have a test case for it

PR #1870, opened 2025-04-27
fix(cdk): mark scheduler tasks as pending tasks

### LINKED ISSUE #1867: RxFor destroys/recreates DOM on long initial rendering, despite Angular non-destructive hydration was enabled
# Description
When the Server-Side-Rendered HTML is initially displayed in the browser, everything is visible. 
And then CSR kicks in, and RxFor cuts long rendering task into smaller tasks (as intended).
But unfortunately Angular doesn't wait until all of DOM elements are rendered by RxFor and Angular already considers the hydration as done and destroys the Server-Side-Rendered DOM elements which were not yet rendered in CSR by RxFor. Therefore RxFor indirectly causes the DOM elements to be destroyed and recreated only later incrementally in separate tasks.
This might affect negatively CLS because the incrementally arriving components can cause layout shifts.

See the minimal app reproduction repo:
https://github.com/Platonn/test-rx-angular-non-destructive-hydration

For more, see screenshots

Initial SSR HTML displayed:

Image

... and then CSR kicks in, Angular destroys the DOM and RxFor renders DOM elements incrementally in separate tasks:

Image

Image

Image

Image

Image

Image

For reference, I'm attaching also an exported Performance trace from Chrome Dev Tools:
[Trace-20250402T144018.json](https://github.com/user-attachments/files/19568068/Trace-20250402T144018.json)

# Environment
```
@rx-angular/template@19.2.1
```

```
Angular CLI: 19.2.5
Node: 22.13.1
Package Manager: npm 10.9.2
OS: darwin arm64

Angular: 19.2.4
... platform-browser, platform-browser-dynamic, platform-server
... router

@angular-devkit/architect       0.1902.5
@angular-devkit/build-angular   19.2.5
@angular-devkit/core            19.2.5
@angular-devkit/schematics      19.2.5
@angular/cli                    19.2.5
@angular/ssr                    19.2.5
@schematics/angular             19.2.5
rxjs                            7.8.2
typescript                      5.7.3
zone.js                         0.15.0
```

| Google Chrome | 134.0.6998.166 (Official Build) (arm64) |
| -- | -- |
| Revision | 0b26d3a1ee1e44572492002c2e52ffcd13ac701b-refs/branch-heads/6998@{#2123} |
| OS | macOS Version 14.6.1 (Build 23G93) |
| JavaScript | V8 13.4.114.21 |
| User agent | Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/134.0.0.0 Safari/537.36 |

hoebbelsB:
# Description

fixes #1867 

Not the most beautiful of all solutions, but we are lacking DI context in the scheduler. 

@Platonn please have a look at the solution. I've tested it on the repository you've shared with us in your issue. For me it solved the issue

### DIFF
--- a/libs/cdk/render-strategies/src/lib/concurrent-strategies.ts
+++ b/libs/cdk/render-strategies/src/lib/concurrent-strategies.ts
@@ -1,4 +1,4 @@
-import { NgZone } from '@angular/core';
+import { NgZone, PendingTasks } from '@angular/core';
 import { coalescingManager, coalescingObj } from '@rx-angular/cdk/coalescing';
 import {
   cancelCallback,
@@ -17,6 +17,14 @@ import {
 // set default to 60fps
 forceFrameRate(60);
 
+let pendingTasks: PendingTasks | undefined;
+
+export function setPendingTasks(p: PendingTasks) {
+  if (!pendingTasks) {
+    pendingTasks = p;
+  }
+}
+
 const immediateStrategy: RxStrategyCredentials = {
   name: 'immediate',
   work: (cdRef) => cdRef.detectChanges(),
@@ -27,7 +35,7 @@ const immediateStrategy: RxStrategyCredentials = {
           ngZone,
           priority: PriorityLevel.ImmediatePriority,
           scope,
-        })
+        }),
       );
   },
 };
@@ -42,7 +50,7 @@ const userBlockingStrategy: RxStrategyCredentials = {
           ngZone,
           priority: PriorityLevel.UserBlockingPriority,
           scope,
-        })
+        }),
       );
   },
 };
@@ -57,7 +65,7 @@ const normalStrategy: RxStrategyCredentials = {
           ngZone,
           priority: PriorityLevel.NormalPriority,
           scope,
-        })
+        }),
       );
   },
 };
@@ -72,7 +80,7 @@ const lowStrategy: RxStrategyCredentials = {
           ngZone,
           priority: PriorityLevel.LowPriority,
           scope,
-        })
+        }),
       );
   },
 };
@@ -87,7 +95,7 @@ const idleStrategy: RxStrategyCredentials = {
           ngZone,
           priority: PriorityLevel.IdlePriority,
           scope,
-        })
+        }),
       );
   },
 };
@@ -99,7 +107,7 @@ function scheduleOnQueue<T>(
     scope: coalescingObj;
     delay?: number;
     ngZone: NgZone;
-  }
+  },
 ): MonoTypeOperatorFunction<T> {
   const scope = (options.scope as Record<string, unknown>) || {};
   return (o$: Observable<T>): Observable<T> =>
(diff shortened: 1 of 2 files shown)


eneajaho (approved):

libs/cdk/render-strategies/src/lib/concurrent-strategies.ts
```diff
@@ -17,6 +17,14 @@ import {
...
 forceFrameRate(60);
 
+let pendingTasks: PendingTasks | undefined;
+
+export function setPendingTasks(p: PendingTasks) {
+  if (!pendingTasks) {
+    pendingTasks = p;
+  }
+}
+
```
Platonn, line 27: Many thanks @hoebbelsB for taking care of this issue. 🙌 

Do we really have to store an application-specific instance of `PendingTasks` class in a global context?
I'm afraid it's prone to bugs in SSR where multiple apps rendered in parallel can and overwrite this single global context property, interfering with other renderings (which should be independent).

More details:
- I can see in Angular Source code, [that `PendingTasks` is provided in the root injector](https://github.com/angular/angular/blob/104ad0a3512f6db893083e4336448dbdf0a06a03/packages/core/src/pending_tasks.ts#L155-L159), so it's scoped to a single instance of an application.
- In SSR there can exist multiple instances of Angular applications in parallel (each being rendered for a different parallel incoming request) - all executed in the same global NodeJS context.
- Therefore, I'm afraid in SSR parallel renderings may interfere with each other, overwriting the NodeJS global shared property `let pendingTasks`, which can potentially cause bugs.

Again, thank you for taking a look into this issue :)

hoebbelsB: thanks for pointing that out. It might require a bigger refactoring from our end in that case. Let me think about it

Platonn: Yea, I can see in the source code of RxAngular that we're now drilling down the `NgZone` argument deep down through the chain of functions (from RxLet, RxFor, ... directives down to `scheduleOnQueue()`).

I believe the usecase for the bugfix is valid.
But I can understand your concern: "is the refactoring worth it?"

Platonn: Hi @hoebbelsB ,
I'm not rushing for implementation, I'm aware you have plenty of responsibilities. 
I could possibly go ahead and contribute a PR proposal, if only having green light from you. But I'm also not pushing for it, just asking.

So let me just ask for your architectural opinion:
Do you think it would be OK to drill down the `PendingTasks` object of Angular, same like we drill down the `NgZone` object - from the `RxFor` instance down through all nested function invocations down to the `scheduleOnQueue()` function?

I could take the human effort implement it. The long term business and architectural pros&cons:
- pros: it unblocks us to fix the bug of `RxFor` (and other directives) causing the destroy of the DOM (and sometimes degrading CLS metric) despite Angular native "non-destructive-hydration" being enabled
- cons: changing many places in the codebase just for one usecase - "pollutting" many the functions with the new (optional) param `PendingTasks`

Breaking changes in the TS API can be avoided thanks to:
- some functions already being private API, so the new param can be required in TS types
- public API functions can get the new param as an optional one

I have problem to find alternative approaches, due to my limited knowledge about the RxAngular/Template architecture. Feel free to share what alternative approaches and their pros&cons can you see. Or maybe you're fine with the approach mentioned above.

Many thanks!

arturovt: I'm wondering is that possible to pass the pending tasks instance in the same way as ngzone is passed?

eneajaho: At this point, I'd drill down the injector, so at any point we can inject anything we need, and replace anything we need in these helper functions.

PR #1875, opened 2025-06-17
chore(deps): update nx to v21.2.0 & angular to v20.0.3

gestj:
We also can't update to Angular 20 because of `rx-angular` staying behind.

So, I started this branch & rather big update. The 20 update is painful. Its big with many changes for big / complex projects. We also suffer in our project.

I hoped I could make it work for `rx-angular` which is less complex than ours.
But sadly, it isn't that easy in this project as well.

I of course struggle with the tiny details since I am not used to behind-the-scenes coding tightly coupled to deep nasty details of Angular framework.

This failing test in stopped my approach for now.

`libs/template/unpatch/src/lib/tests/unpatch.directive.spec.ts`

BUT; Maybe you can make use of it if you consider an Angular 20 update.

PLEASE; Take over. I won't be able to successfully make this ready for merge.

---

Maybe, what did I do?

```
yarn nx migrate latest
# After this I changed package.json > engines > node; Cause 18.13 isn't supported anymore by Angular 20; https://angular.dev/reference/versions
yarn
yarn nx migrate --run-migrations # during this you'll run into this: https://github.com/nrwl/nx/issues/20172#issuecomment-1825783434; But I could verify (with temporary local additional changes that the failed migrations aren't needed for this codebase)
```

The migrations of NX & Angular will change a LOT of files. E.g. the template changes where `ngIf`, `ngFor`, `ngSwitch` get replaced by their control flow counter parts. Others are config changes by NX migrations.

After that:
-  I disabled `@angular-eslint/prefer-inject` where needed (and in one place I fixed the issue since it was only one spec file).
- I fixed the broken `test-setup.ts` files (the way `jest-preset-angular` works changed with 14.6.0)
- I fixed broken effect setup in `toObservableMicrotask` since API change due to https://github.com/angular/angular/pull/60535 (part of Angular 20)
- Very nasty change => in `zone-config.spec.ts` you patch `Zone` in window; This broke. I disabled (commented) the related `beforeAll` & `afterAll`. I had this idea after I saw how tests are written. I asked myself why the hell is this needed in first place, I bet it works without it. And yes, the tests run with out it. But, I can't judge it since I don't know the code by heart....


package.json
```diff
@@ -65,69 +65,69 @@
...
+    "jest-preset-angular": "14.6.0",
     "jsonc-eslint-parser": "^2.1.0",
     "klaw-sync": "^6.0.0",
     "lint-staged": "^12.0.3",
     "lodash": "^4.17.21",
     "markdown-link-check": "^3.11.2",
     "ng-morph": "^4.8.4",
-    "ng-packagr": "19.0.1",
-    "nx": "20.1.0",
+    "ng-packagr": "20.0.0",
```
hoebbelsB, line 121: can u update to 20.0.1?

package.json
```diff
@@ -39,20 +39,20 @@
...
+    "@angular/common": "20.0.3",
+    "@angular/compiler": "20.0.3",
+    "@angular/core": "20.0.3",
+    "@angular/forms": "20.0.3",
+    "@angular/material": "20.0.3",
+    "@angular/platform-browser": "20.0.3",
+    "@angular/platform-browser-dynamic": "20.0.3",
+    "@angular/platform-server": "20.0.3",
+    "@angular/router": "20.0.3",
+    "@angular/ssr": "20.0.2",
```
hoebbelsB, line 54: /ssr also was released with 20.0.3. can u update this?

PR #1896, opened 2026-01-02
fix(template): keep scroll offset when using keepScrolledIndexOnPrepend

### LINKED ISSUE #1857: Virtual Scroll: scroll position jumps when using keepScrolledIndexOnPrepend

hoebbelsB:
Extension of #1859, as I couldn't push to the fork.

Fixes #1857.

## Problem

`keepScrolledIndexOnPrepend` compensated the scroll position by counting how many items were inserted ahead of the anchored item. That only holds for a pure prepend.

A real reverse infinite scroller — a chat client loading older messages — also renders a transient *"loading older messages"* row while the request is in flight, so the update that resolves it is an insert **and** a remove. Counting insertions mis-compensates by the height of that row and the list jumps.

## Approach

Locate the anchored item again via `trackBy` instead of counting insertions. That nets out inserts, removals and moves ahead of the anchor in a single pass.

The compensation then shifts by the resulting delta rather than scrolling to the *top* of the new anchor index — the latter drops the anchor's sub-item offset, which is the jump reported in #1857.

## Also fixed

Validating the above against a realistic demo surfaced several further defects in the same area. Each has its own regression spec.

| Symptom | Cause |
| --- | --- |
| Empty list, nothing rendered at all | The strategies only compute a range once `containerRect$` emits, and it was fed exclusively by a `ResizeObserver` — which never delivers its first entry without a rendering frame (hidden or fully occluded tab). Now seeded with a synchronous measurement on init. |
| Views stacked at the end of the content, gaps in the viewport | A prepend shifts the range while the rendered slice stays identical, so the differ reports no changes and rendering bailed out via `NEVER`. Now re-renders in place — and announces the batch correctly, since an empty batch made the strategies' position pass fast-forward its cursor past views that still emit afterwards. |
| Overlapping items and gaps that never recover while scrolling | Autosize booked measurements through `view.context.index`, which goes stale when a second emission rebuilds the size ledger while a render or `ResizeObserver` pass for the previous dataset is still in flight. Because sizes are cached, that corruption was permanent. Now resolved through the `trackBy` cache. |
| Stale loading row frozen on screen | `waitForScroll` blocks rendering until a scroll event arrives, but a target outside the scrollable bounds is clamped by the browser and emits no event — latching `isStable` to `false` forever. Now decided against the clamped target. |
| Viewport pinned to the top, reloading forever | When the anchored item itself disappears — the anchor sits *on* the transient row, which happens whenever the user waits at the very top — the compensation bailed out entirely. Now falls back to the nearest surviving item below it. |

Additionally, autosize applies the range shrink independently of the relocation (a transient row that was both the anchor and the last item of the rendered range used to wedge a pending `scrollToIndex`), and render/measure passes tolerate indices belonging to a superseded dataset instead of dereferencing a removed ledger entry.

## Testing

`nx run template:component-test` — **82 specs, 23 of them new**, across the fixed, dynamic and autosize strategies. Each fix listed above has a spec that was verified to fail against the unfixed code.

## Demo

`apps/demos` → *Virtual Scrolling* → *Reverse Infinite Scroll* was reworked to behave like an actual chat client, which is what surfaced most of the defects above:

- a transient loading row inside the list (toggleable) and a slower simulated backend, so the in-flight state is actually observable
- history requests driven by user input — wheel upwards, or arriving at the very top — instead of range emissions. Passive scroll changes (initial scroll, prepend compensations, measurement corrections) all report a range starting at `0` and would otherwise chain-load without any interaction. Wheel events are debounced to one trigger per gesture, since trackpad momentum keeps emitting for seconds after the flick.
- history runs dry at the beginning of the conversation instead of refetching the first page forever

## Known limitation

Scrolling upward into items that have never been measured can still produce a visible correction: the anchor walk uses `tombstoneSize` estimates, and `maybeAdjustScrollPosition` applies the real geometry a frame later, mid-gesture. With high per-item size variance this reads as rubber-banding near the top.

Setting `tombstoneSize` close to the real mean item height reduces it substantially, and that is worth documenting for reverse infinite scrollers. Damping corrections while a gesture is in progress would address the remainder, but that is a behavioural change to the scroll strategies and is left for a separate PR.


G-iele:
@hoebbelsB when do you plan to merge this PR? ))

PR #1904, opened 2026-01-30
feat(template): introduce server-aware virtual views

eneajaho:
Check individual commits.

### DIFF
--- a/libs/template/virtual-view/src/lib/virtual-view.directive.ts
+++ b/libs/template/virtual-view/src/lib/virtual-view.directive.ts
@@ -413,12 +413,23 @@ export class RxVirtualView
    * @private
    */
   #showContentWhenDisabled() {
-    const content = this.#content();
-    if (content) {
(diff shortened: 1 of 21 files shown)


libs/template/virtual-view/src/lib/virtual-view.directive.ts
```diff
@@ -83,9 +81,17 @@ export class RxVirtualView
...
   readonly #destroyRef = inject(DestroyRef);
   readonly #config = inject(VIRTUAL_VIEW_CONFIG_TOKEN);
+  readonly #platform = inject(PLATFORM);
+  readonly #injector = inject(INJECTOR);
+
+  readonly #enabled = computed(() =>
+    isSignal(this.#config.enabled)
+      ? this.#config.enabled()
+      : this.#config.enabled,
+  );
```
hoebbelsB, line 91: this never changes in case it's not a signal, right? Is that intended?

PR #1912, opened 2026-07-02
feat(rebundle): rebundle build optimizer

ChristopherPHolder:
## TL;DR

Introduce `rebundle`, an esbuild plugin for optimizing the bundle output of Angular applications.

**Motivation**

## Motivation

Angular CLI migrated from [Webpack](https://webpack.js.org/) to [esbuild](https://esbuild.github.io/). This has brought major build-time performance improvements, but it can also introduce [runtime performance regressions in large projects](https://github.com/angular/angular-cli/issues/27321).

The root cause is esbuild’s [code splitting algorithm](https://github.com/evanw/esbuild/blob/main/docs/architecture.md#code-splitting), which focuses on minimizing the amount of code required by each application entry point. That is a useful default, but it does not account for additional factors such as entry point hierarchy, route structure, application configuration, or project-specific loading behavior. esbuild also does not currently expose an API for configuring code splitting behavior.

To address part of this problem, the Angular team introduced an [experimental chunk optimizer](https://github.com/angular/angular-cli/pull/27953), which rebundles the application with Rolldown to reduce the number of generated bundles. This can significantly reduce chunk count, especially for initial bundles.

However, that solution is intentionally limited: it does not expose extension points or configuration hooks for application-specific optimization strategies. `rebundle` builds on the same general idea of post-processing esbuild output with Rolldown, while adding configurable merge strategies so applications can tune how chunks are grouped based on their own dependency graph and loading characteristics.

## Solution

TODO add explanation

## Impact

TODO add experiment details

## Usage

TODO


ChristopherPHolder:
Moved this PR to an upstream repository branch now that I have contributor access: #1916

PR #1925, opened 2026-07-13
fix(rebundle): improve generator and build output

ChristopherPHolder:
## Summary

- migrate Angular application and dev-server targets to the Nx executors required by the configure generator
- include the rebundle README and package metadata in the published package
- enable Rolldown minification while removing legal comments and debug metadata
- log before/after file counts, raw sizes, transfer sizes, and rebundle duration

## Why

The configure generator accepted only Nx executors even though Angular CLI projects use the Angular build executors, which prevented rebundle from being configured directly. The published package also omitted its README and did not expose its package metadata. Finally, rebundled output needed explicit production minification settings and visible build statistics.

## Impact

Angular CLI projects can now run the configure generator and be migrated to compatible Nx executors. Published packages include their README and exported package metadata, while production builds produce smaller output and a comparison table showing rebundle results.

### DIFF
--- a/.nxignore
+++ b/.nxignore
@@ -1,1 +1,2 @@
 libs/**/*.md
+!libs/rebundle/README.md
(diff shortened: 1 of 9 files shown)


ChristopherPHolder:
Example stats log from rebundle demo

image

